PR #16, opened 2012-12-19
Created and modififed Markdown version of README.TXT

sroberts:
Love the project and just wanted to help out a bit. I created a Markdown version of the README, which tends to display a little better, and rearranged some of the content to flow a little better, I believe without changing any meaning.

Hope this helps, you guys rock! :+1: 

SR


hsluoyz:
@brenosilva @sroberts , I'm new to this project and I want to learn it. But the first thing I encountered is that the format of the README hurts my eyes :(

I had a read about this thread. It has been 5 years and we are still at GitHub. Moreover, Markdown is a neutral and defacto dominating standard for open-source intro pages. It is not only useful in GitHub but also other places. Even if one day we leave GitHub, we still can utilize the Markdown documents. So I think we should delete TXTs and use Markdown instead. What do you think?

victorhora:
@hsluoyz we are already using a Markdown with the README in the newest version of ModSecurity, libModSecurity (aka v3.0):

https://github.com/SpiderLabs/ModSecurity/blob/v3/master/README.md

hsluoyz:
Hi @victorhora , my team still believes that ModSecurity 3.0 is not mature enough, there are still many bugs and the performance cannot reach 2.x for now. Moreover, ModSecurity 3.0 is not available on Windows but I'm a Windows user :(

So I think it's still meaningful to port 2.x's README to Markdown. What do you think?

Issue #705, opened 2014-04-18
ProcessPartial mode does not function as expected


justintime32:
I'm currently rolling out ModSecurity, and I've found that ProcessPartial mode does not function as one would expect.

As I understand it, the expected behavior is that ModSec will buffer as much of the request body as it is configured to (based on SecRequestBodyLimit) and use that as a basis for rule processing while allowing the request to pass through (in accordance with the rules).

What actually seems to happen is that the entire request body will be buffered, regardless of the limits set (prior to my earlier patch, #597, it would reject the request based on the Content-Length header in phase 1). This is not ideal for me, as I do not want data to be buffered to disk at all, and buffering entire requests (which may be as large as 5 GB) in memory has a number of disadvantages.

I had attempted to make some small changes to the input filter, but ultimately these ended poorly (requests being truncated under certain circumstances, not unlike #406). I've had to employ a really unfortunate workaround so that ModSec is usable (basically just disabling buffering for larger requests).

Does anyone else see this behavior? Perhaps I'm missing something?

zimmerle:
The fix is now available on our mainline. The pull request #709 was just merged.

PR #709, opened 2014-04-28
Stop buffering when the request is larger than SecRequestBodyLimit in ProcessPartial mode

justintime32:
This change should address #705. It will cause ModSec to stop buffering the incoming request body if it is larger than the configured request limit and in ProcessPartial mode. It will not send an EOS bucket in the forwarding phase of the input filter if it did not read an EOS bucket.

A side effect of handling buffering like this is that request body parsers will sometimes throw errors if they cannot see all of the request data.


zimmerle:
Hi @justintime32, thank you for the patch. Finally merged. It will be released as part of v2.9.1

PR #745, opened 2014-06-20
Added possibility to specify a data file with a name relative to httpd root (as include files) + conditional logging + allow duplicate rule ids and no id

marcstern:
1. Logic to look for data file:
  - try given filename (absolute or relative to current dir)
  - if not absolute, try
      - from the rule directory
      - from apache root directory

2. Pre-compiler flags for conditional logging in audit log (disable some entries like producer, file name, etc.)

3. Added rule id & msg in "Rule processing failed" log entry

4. Pre-compiler flags to allow duplicate rule ids and no id

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -786,6 +786,7 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
         return my_error_msg;
     }
 
+#ifndef ALLOW_ID_NOT_UNIQUE
     /* Rules must have uniq ID */
     if (
 #if defined(WITH_LUA)
@@ -818,6 +819,7 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
             //    return "ModSecurity: Found another rule with the same id";
         }
     }
+#endif
 
     /* Create default actionset if one does not already exist. */
     if (dcfg->tmp_default_actionset == NULL) {
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -1507,6 +1507,8 @@ static const char *cmd_default_action(cmd_parms *cmd, void *_dcfg,
         return apr_psprintf(cmd->pool, "ModSecurity: SecDefaultAction must not "
             "contain any metadata actions (id, rev, msg, tag, severity, ver, accuracy, maturity, logdata).");
     }
+
+#ifndef LOG_NO_DEFAULT_DEPRECATED
     /* These are just a warning for now. */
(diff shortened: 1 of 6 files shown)


zimmerle:
Implementing the include logic in libModSecurity, using:

1. Try to open the resource relative to the binary directory (current dir or absolute path).
2. Try to open the resource relative to the rule directory

https://github.com/SpiderLabs/ModSecurity/tree/libmodsecurity

zimmerle:
Hi @marcstern, thanks again for the patch!

Extracting the patches from this pull request, we have the following gits:

https://gist.github.com/zimmerle/f6321b081ed949cf4022

Splinting the code into functionalities, we have:

* Update the order to look for "resource files" on different patches.
  - 0001-Added-possibility-to-specify-a-data-file-with-a-name.patch
  - 0002-UTF-8-typos.patch

**Comments:** That was "merged" already into ModSecurity version 3, as commented above.

* Compile flag NO_CON_DROP_WARN to suppress message "Error: Connection drop not implemented on this platform" and various compilation flags to enable/disable specific logging message.
 - 0003-Compile-flag-NO_CON_DROP_WARN-to-suppress-message-Er.patch
 - 0004-Conditional-compile-flags-for-logging-entries-in-aud.patch
 - 0005-Forgot-rc-in-error-message.patch

**Comments:** We need to understand why it is necessary. If it is a benefit to the entire community, we have to have a documentation and configuration flags enable/disable those compilation flags.

* Disable status message warning.
 - 0006-Compile-flag-LOG_NO_WARN_STATUS-to-disable-Status-en.patch

**Comments:**  I can't see the benefit of having this patch in the mainline. The usage status is something important to the project. The "ModSec status" help us to prioritize the issues and grab general attention to the project. More details here: https://www.trustwave.com/Resources/SpiderLabs-Blog/Introducing-ModSecurity-Status-Reporting/ 

* Disable the checking for rules with the same ID. 
 - 0007-Added-ALLOW_ID_NOT_UNIQUE-compile-flag-to-allow-dupl.patch

**Comments:** As discussed in the last community meeting, this may not be a good idea "usability wide". It is very common to have users complain about rules loaded twice, imagine without those validations? This may be very hard for the average user. May not be a good idea for the mainline.


The general status of this pull request is that it was partially merged. Some of items still open for discussion. I would like to ask @marcstern to open a new pull request, one for each of the features/benefits that he wants to discuss with the community. Therefore, I will close this one.

marcstern:
OK, I'll create new issues to discuss the details.
I assume "0005-Forgot-rc-in-error-message.patch" will be merged, right?

PR #749, opened 2014-06-30
Passthrough the saved Response headers in the response

rmongia:
Tickets: https://github.com/SpiderLabs/ModSecurity/issues/735


zimmerle:
Hi @rmongia,

Thanks for the patch.  It is now merged into nginx_refactoring.

PR #767, opened 2014-08-08
Allow non-zero Content-Length for HEAD requests

rmongia:
--- Quote from RFC-2616 ---
" The Content-Length entity-header field indicates the size of the entity-body,
in decimal number of OCTETs, sent to the recipient or, in the case of the HEAD
method, the size of the entity-body that would have been sent had the request
been a GET."


zimmerle:
Hi @rmongia, 

Thank you for the patch. Just got it merged on nginx_refactoring.

PR #775, opened 2014-09-26
mlogc-batch-load.pl.in: fix searching SecAuditLogStorageDir files with A...

yersinia:
...pache 2.4

Setting SecAuditLogType concurrent the files created by modsecurity in the directory SecAuditLogStorageDir
are of the type

%Y%m%d-%H%M%S-UNIQUE_ID

where UNIQUE_ID is produced by the module mod_unique_id. In apache 2.2
UNIQUE_ID was 24 characters long while in apache 2.4 is 27 characters long,
as a result of the new version of mod_unique_id. This patch extends
the regular expression for searching these  log files for apache 2.4 and apache 2.2,
and also with future releases, at least with regard to the length of UNIQUE_ID

Signed-off-by: Elia Pinto <elia.pinto@gmail.com>


zimmerle:
Hi @yersinia,

Thank for the patch! The patch was merged and it will be released as part of ModSecurity version 2.9.1

Issue #781, opened 2014-10-03
mod_security bug with random numbers generation


gilperon:
I finally could create a way to reproduce the bug that I am having. When 2 or more users call a page at the same second modsecurity generates the same random number (using rand() function from php) to all users.

Here is a demonstration of the bug:

http://quemfazsite.com.br/em_criacao/modelo9/teste.php

Opening this page 2 iframes will load and each one should be generating random numbers independetly of each other but both frames are generating the same random numbers! The very simple source code can be seen at:

http://pastebin.com/8ELAnmdp

If this is not the right place to post this kind of bug please tell me the right place. This is my first contribution and I dont know too much about this community.

zimmerle:
Hi @gilperon the current mainline code contains the fix for that issue. The fix was provided by @micha137. Thank you for the report and thanks @micha137 for the fix [#838]  ;)

It will be release as of v2.9.1. The release date is not defined yet.

PR #796, opened 2014-10-29
Remove misguided call to srand()

### LINKED ISSUE #778: Bug in mod_security
Hi,

I've been using mod_security for a long time and it's great but recently I discovered a bug that many customers of mine always complained but I didnt care. Today I decided trying to solve it and I discovered it's a bug with mod_security.

I am running Apache 2.4 in Centos 7 with the most updated version of mod_security installed using "yum install mod_security". The bug is this:

when a user tries to upload 2 or more files using any upload plugin (like valums or jquery.onprogress event) that send each file using a php POST request and my php file generates a random number with rand(0,999999) function from PHP library for some reason the number generated in all file upload is the same. When I disable the mod_security (adding # in front of LOADMODULE) the rand number is generated fine.

For example: if you try to upload 3 files using valums, for each file the plugin will request the script /server/php.php in the server trying to send the 3 files (I can see that with the Chrome developer tools F12). When the upload is completed my script under server/php.php generates a random number to save each file with a random number and this number happens to be exactly the same for the 3 files when mod_security is enabled. Is 3 files have about the same size the problems happens at about 90% of the time, but if the 3 files have very different sizes the problem does not happen and the random generated number is different.

I can only make this bug go away if I disable mod_security (which is not a good idea cause I recently get attacked in my server) or to generate the file names using microtime(true) cause doing so the file name will not be random but will have an incremental number according to the microtime it was processed.

I am sorry if I could not explain this bug clearer but I did not find any easier way to do so. If you need additional details I will be glad to provide. I hope mod_security community can help solve this bug or at least point me in the right direction explaining why the rand(0,99999) is generating the same number accross different requests done in a close interval of time

micha137:
A random number generator needs to be initialized once per process after a fork, but not after each request, more so with an argument that changes only once per second.

This fixes #778 and its duplicate #781.

### DIFF
--- a/apache2/Makefile.am
+++ b/apache2/Makefile.am
@@ -41,7 +41,8 @@ mod_security2_la_CFLAGS = @APR_CFLAGS@ \
     @LUA_CFLAGS@ \
     @MODSEC_EXTRA_CFLAGS@ \
     @PCRE_CFLAGS@ \
-    @YAJL_CFLAGS@
+    @YAJL_CFLAGS@ \
+    @SSDEEP_CFLAGS@
 
 
 mod_security2_la_CPPFLAGS = @APR_CPPFLAGS@ \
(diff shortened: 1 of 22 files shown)


micha137:
I noticed the call to srand() is still in 2e09b8e542bb877f4849881c101a9e57f8595323 (in file apache2/modsecurity.c line 240).

Those commits from 5 Dec 2013 are confusing.  I will close this pull request and create a now one, only including the change from deec149.

PR #810, opened 2014-12-04
Fix compiler warnings about pointer integer conversion

jpereira:
that is it.


zimmerle:
Most recent version we have changed the ap_find_command to msc_remote_find_command (that is our own implementation of ap_find_command). That made easy the compilation of ModSecurity in different platforms. I am closing this Pull Request.

PR #817, opened 2014-12-18
custom x-* headers are just skipped when using nginx with proxy_pass to origin

ton31337:
Custom x-* headers are skipped when using nginx as proxy_pass to origin. They are displayed only in audit_log, but the client doesn't see them. For example X-Flash-Message header.


zimmerle:
Hi @ton31337, 

I am closing this pull request in favor of #826. It seems that you've got more comments there.

Just to keep it registered, the changes on #817 and #826 are the same:
https://github.com/ton31337/ModSecurity/commit/9abda55714d46523576bfb057c9c304d545ea852

PR #824, opened 2014-12-30
fixing compliation for systems where APR doesn't have crypto support

nevins-b:


micha137:
I think this can be closed, now that #834 was fixed by f9464b46bf3cddcf857c453cc2524b2cacf2b73b.

Issue #814, opened 2014-12-15
Lua versions 5.2+ are not working well inside ModSecurity core


zimmerle:
The support for Lua is a very useful mechanism inside ModSecurity core. However, the core is not ready for most recent versions of Lua, leading to some instability. As reported by @lifeforms.

https://gist.github.com/lifeforms/3ecc60c67012a053d060

(Lua 5.3 is currently on -RC0 -- http://www.lua.org)

zimmerle:
Support to Lua 5.3 was added.

Issue #762, opened 2014-07-31
build: Check to the version of the dependencies during the build.


zimmerle:
Currently the build system is failing to identify the version of Yajl and Lua, which may leads the build to fail.

Version check should be supported inside the find_lua and find_yajl scripts:
 - https://github.com/SpiderLabs/ModSecurity/blob/master/build/find_lua.m4
 - https://github.com/SpiderLabs/ModSecurity/blob/master/build/find_yajl.m4

zimmerle:
Lua version is now fixed.

PR #837, opened 2015-02-13
Fix build issue with Lua >= 5.3

athmane:
Hi,

Please find this PR to build mod_security with Lua 5.3 (the default in Fedora rawhide).

### DIFF
--- a/apache2/msc_lua.c
+++ b/apache2/msc_lua.c
@@ -111,8 +111,11 @@ char *lua_compile(msc_script **script, const char *filename, apr_pool_t *pool) {
     dump.pool = pool;
     dump.parts = apr_array_make(pool, 128, sizeof(msc_script_part *));
 
+#if LUA_VERSION_NUM >= 503
+    lua_dump(L, dump_writer, &dump, 1);
+#else
     lua_dump(L, dump_writer, &dump);
-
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @athmane,

Thank you for the patch. This is related to the issues: #814 and #762

Looking at your patch it seems that this very change could also work with Lua 5.2 right? 

Not placing on buildbots right now because I have to upgrade the buildslaves to use Lua 5.3 :)

athmane:
Hi @zimmerle 

Yes, the change does not affect older Lua versions, I tested the patch with Lua 5.2.2  (default in Fedora 21) and Lua 5.3.0  (default in Fedora Rawhide / devel).

zimmerle:
The build seems fine, however there are some crashes on runtime. I am investigating. 

We have two new build slaves for Lua 5.2 and 5.3:
- http://www.modsecurity.org/developers/buildbot/buildslaves/Linux64-lua52
- http://www.modsecurity.org/developers/buildbot/buildslaves/Linux64-lua53

zimmerle:
Hi @athmane,

Thanks for the patch. Partially merged and fixed into our mainline. It will be released as part of ModSecurity v2.9.1

PR #852, opened 2015-03-26
Change auditlog file permission from CREATEMODE to dcfg->auditlog_fileperms

littlecho:
Change third parameter(which is the apr file permission flag) from CREATEMODE to dcfg->auditlog_fileperms. Due to the user can specify the desired file permission setting for the audit log files with setting the value of SecAuditLogFileMode, we should follow the file permission setting from the config file. Therefore, as the dcfg->auditlog_fileperms will be modified in cmd_audit_log_dirmode function, we can use the value while calling apr_file_open to meet the file permission that specified in modsecurity.conf.


zimmerle:
Hi @littlecho, 

Sorry for the huge delay. Your patch is now merge. Made some modification to replicate your work on the secondary audit log file. It will be released as part of v2.9.1

PR #872, opened 2015-04-21
Compiles against ssdeep in non-standard location

cpanelkurt:
Issue 871: (*nix) Tacks on the include and lib directory when
a the user specifies a path for ssdeep in an alternate location.


cpanelkurt:
This is related to [Issue 871](https://github.com/SpiderLabs/ModSecurity/issues/871).

zimmerle:
Merged into our mainline. Thanks.

Issue #662, opened 2014-02-12
DURATION fails to accurately work for requests that take more than one second


aweits:
apr_time_usec is apparently defined as follows:
# define apr_time_usec (   time)    ((time) % APR_USEC_PER_SEC)

Which leads re_variables.c:1740 to not behave as expected when duration exceeds one second.

victorhora:
Fixed as of: https://github.com/SpiderLabs/ModSecurity/commit/3044ad012b7fe0b417374bd032dc183b50ab4fbc

PR #879, opened 2015-04-30
Fix for #662.

aweits:
apr_time_usec is apparently defined as follows:

#define apr_time_usec ( time) ((time) % APR_USEC_PER_SEC)

Which leads DURATION to not behave as expected when duration exceeds one second.


zimmerle:
Hi @aweits,

Your code was merge already. It will be release as part of v2.9.1

Thank you!

PR #881, opened 2015-05-03
Allow user to choose between TLS versions(TLSProtocol option introduced)...

badbug0101:
A new option `TLSProtocol` allows use to select the TLS protocol version to use:
`TLSProtocol 2`

If the detected libcurl version at runtime is less than 7.34.0, then it reverts back to the old method (i.e., use TLS v1.x) since version below 7.34.0 do not support TLS v1.2.


zimmerle:
Hi @badbug0101,

Thank you for the patch! Sorry for the long time to review it. It is now on our Buildbots.

zimmerle:
Hi @badbug0101,

Your code was merged and it will be released as part of ModSecurity version 2.9.1

Thank you!

PR #883, opened 2015-05-05
Fix WITH_APU_CRYPTO check on 32-bit Linux platform

cpanelkurt:
All of the apr flags needed to compile APU_HAVE_CRYPTO check
in configure.ac aren't passed.  While this works fine for 64-bit
machines (because _LARGEFILE64_SOURCE is already defined),
this does not work on 32-bit.  This in-turn breaks the apr_off_t
definition in apr.h.

By passing along the apr --cflags and --cppflags to compile,
32-bit machines will allow WITH_APU_CRYPTO to be set if there's
support for it.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,9 +1,11 @@
 DD mmm YYYY - X.Y.Z (To be released)
 ------------------------------------
 
+ * Fix apr_crypto.h check on 32-bit Linux platform
+   [Issue #882 - Kurt Newman]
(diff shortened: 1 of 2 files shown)


cpanelkurt:
This pull request has a logic error.  I will be re-submitting the request in a moment.

zimmerle:
Hi @cpanelkurt,

Thank you for the patch! It was merged into our mainline. Your code will be released as part of v2.9.1.

Issue #830, opened 2015-01-20
Modsecurity Ajax Error: Internal server Error


rz300:
After installing modsec from official instructions my site doesn't work correctly. Admin side from cms is build with ajax and something else and it doesn't work.

```
        location / {
        ModSecurityEnabled on;
        ModSecurityConfig /etc/modsecurity/modsecurity.conf;
        testcookie off;
        root /srv/rewired.ru/www;
        proxy_set_header X-Real-IP $remote_addr;
        proxy_set_header X-Forwarded-for $remote_addr;
        proxy_set_header Host $host;
        proxy_pass http://rewired/;
        proxy_connect_timeout 180s;
        proxy_send_timeout 90;
        proxy_read_timeout 180s;
        proxy_set_header Connection close;
        proxy_pass_header Content-Type;
        proxy_pass_header Content-Disposition;
        proxy_pass_header Content-Length;
        proxy_cache_valid any 10m;
        proxy_buffer_size 8k;
        }
```

tail from nginx error.log

```
2015/01/20 14:21:14 [error] 13250#0: *323475 upstream prematurely closed connection while reading response header from upstream, client: 178.140.252.82, server: id-st.ru, request: "POST /admin/update/index.php HTTP/1.1", upstream: "http://127.0.0.1:8080/admin/update/index.php", host: "rewired.ru", referrer: "http://rewired.ru/admin/"
2015/01/20 14:26:52 [error] 29780#0: *4 upstream prematurely closed connection while reading response header from upstream, client: 178.140.252.82, server: id-st.ru, request: "GET /admin/ HTTP/1.1", upstream: "http://127.0.0.1:8080/admin/", host: "rewired.ru"
2015/01/20 14:26:52 [error] 29781#0: *14 recv() failed (104: Connection reset by peer) while reading response header from upstream, client: 178.140.252.82, server: id-st.ru, request: "GET / HTTP/1.1", upstream: "http://127.0.0.1:8080/", host: "plitka-zel.ru"
2015/01/20 14:26:53 [error] 29841#0: *16 connect() failed (111: Connection refused) while connecting to upstream, client: 178.140.252.82, server: id-st.ru, request: "GET /favicon.ico HTTP/1.1", upstream: "http://127.0.0.1:8080/favicon.ico", host: "rewired.ru"
[...]

Issue #582, opened 2013-10-23
Nginx with Modsecurity: POST request gives 500 error


pijnack:
Nginx with Modsecurity gives 500 errors after a POST request.

In error.log:

2013/10/23 13:21:07 [alert] 29432#0: *4 no upstream configuration, client: 88.159.16.136, server: test.huygenscollege.nl, request: "POST /proefwerk/prog/index.php?day=28&month=10&year=2013&klas=2 HTTP/1.1", host: "test.huygenscollege.nl", referrer: "http://test.huygenscollege.nl/proefwerk/prog/index.php?day=4&month=11&year=2013&klas=2"

In nginx.conf:

upstream smoothrac {
        server 88.159.13.153;
}
server {
        listen       80;
        server_name  test.huygenscollege.nl;
        location / {
         ModSecurityEnabled on;
         ModSecurityConfig modsecurity.conf;
        proxy_pass         http://smoothrac/;
        proxy_set_header Host test.huygenscollege.nl;
        proxy_read_timeout 150s;
        }
}

After changing ModSecurityEnabled on to off, the problem is gone.
I guess something is missing in my upstream configuration? 

Piet Pijnacker,
Eindhoven, Netherlands

ton31337:
I can't sudo to root from donatas user. Can we chat in skype or gtalk,
because mail isn't a good way to go..

On Tue, Dec 30, 2014 at 9:33 PM, smtalk <notifications@github.com> wrote:

> ton31337, any news on the problem? :)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpiderLabs/ModSecurity/issues/582#issuecomment-68388922>
> .
>



-- 
Donatas

rz300:
let's use gtalk

rz300:
any news? i wait for gtalk rewired.ag@gmail.com

sasah:
Hello, any news?

rz300:
i 've compiled the original source and and made changes in the /etc/modsecurity/modsecurity.conf 
``SecRequestBodyAccess On`` to ``SecRequestBodyAccess Off`` it helps me.

rm3nchaca:
Please use the nginx_refactoring branch, it appears to be solved. .

jowrjowr:
When will the nginx refactoring branch get merged into the main branch that
OS maintainers use?

zimmerle:
[...]

Issue #142, opened 2013-09-12
file uploads over 8k fail when using ModSecurity 2.7.5 and Nginx 1.4.2


wellumies:
Linux debian 3.2.0-4-amd64 #1 SMP Debian 3.2.46-1+deb7u1 x86_64 GNU/Linux

ModSec 2.7.5 and Nginx 1.4.2

I have an Apache backend and it receives my file uploads and requests if the file is below 8k. Only got the basic modsecurity.conf loaded without any rules. If I set the SecRequestBodyAccess = Off even those pass through. Succesful upload:

---

[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Initialising transaction (txid AcAcAGl3AcAcAcAcAcAcAcAc).
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Transaction context created (dcfg 7f35a9f41980).
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Starting phase REQUEST_HEADERS.
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Second phase starting (dcfg 7f35a9f41980).
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Input filter: Reading request body.
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Multipart: Created temporary file 1 (mode 0600): /var/log/modsecurity_workdir/20130912-151049-AcAcAGl3AcAcAcAcAcAcAcAc-file-vIn5DC
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][5] Adding request argument (BODY): name "submit", value "Submit"
[12/Sep/2013:15:10:49 +0300] [/sid#7f35a9f410a0][rid#7f35a9d950a0][/upload_file.php][4] Request body no files length: 150
[...]

PR #904, opened 2015-06-29
Fix issues with uploading files over 8K : #142, #582 and #830

zimmerle:


zimmerle:
Merged into nginx_refactoring.

PR #906, opened 2015-07-06
Build system fixes (libmodsecurity branch)

LinuxJedi:
1. Fix the build order so "make -j4" works
2. Remove binaries from git tree
3. Add .gitignore file


zimmerle:
Hi @LinuxJedi, thanks for you patch! merged.

PR #950, opened 2015-10-28
Use nginx's parsed hostname

TrungACZNE:
Within the module the request structure's hostname is assigned by ngx_cycle->hostname, which in turn is assigned gethostname(). I don't understand why this is the case, it's not the same as the `Host` header which tells which virtual host is being targeted by the request.

This pull request uses nginx's parsed `Host` header for req->hostname instead.


zimmerle:
Hi @TrungACZNE,

Thank you for your patch! The code is now on the hands of our Buildbots.

zimmerle:
Thanks again @TrungACZNE!

Your patch is now merge into nginx_refactoring branch.

PR #980, opened 2015-11-20
Update issue-960.json with 3rd test.

ajrpayne:
Added third test that checks !@within.

Removed GET from allowed_methods list.

Should trigger rule.


zimmerle:
Merged! Thank you!

PR #982, opened 2015-11-23
Support for correct implimentation of REQUEST_URI

csanders-git:
"This variable holds the full request URL including the query string data (e.g., /index.php? p=X). However, it will never contain a domain name, even if it was provided on the request line."

### DIFF
--- a/src/assay.cc
+++ b/src/assay.cc
@@ -278,7 +278,32 @@ int Assay::processURI(const char *uri, const char *protocol,
     m_collections.store("REQUEST_METHOD", protocol);
     m_collections.store("REQUEST_PROTOCOL",
         "HTTP/" + std::string(http_version));
-    m_collections.store("REQUEST_URI", uri);
+
+std::string parsedURI = uri;
+    // The more popular case is without domain
+    if (m_uri_decoded.at(1) != '/'){
+        bool fullDomain = true;
+        size_t scheme = m_uri_decoded.find(":")+1;
+        if(scheme==std::string::npos){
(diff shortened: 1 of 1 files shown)


src/assay.cc
```diff
@@ -278,7 +278,32 @@ int Assay::processURI(const char *uri, const char *protocol,
     m_collections.store("REQUEST_METHOD", protocol);
     m_collections.store("REQUEST_PROTOCOL",
         "HTTP/" + std::string(http_version));
-    m_collections.store("REQUEST_URI", uri);
+
+std::string parsedURI = uri;
+    // The more popular case is without domain
+    if (m_uri_decoded.at(1) != '/'){
```
zimmerle: It will be very good to have a "sanity" validation like:

```
!m_uri_decoded.empty()
```

We may not have an element on the position "1". The empty string will lead c++ to throw an instance of 'std::out_of_range'

src/assay.cc
```diff
@@ -278,7 +278,32 @@ int Assay::processURI(const char *uri, const char *protocol,
...
+        bool fullDomain = true;
+        size_t scheme = m_uri_decoded.find(":")+1;
+        if(scheme==std::string::npos){
```
zimmerle: Spaces between "if"  and "("  are welcomed :)

zimmerle:
Hi @csanders-git,

Thanks for the patch! :+1: I made some comments inline.

zimmerle:
Hi @csanders-git, thanks for the patch! merged!

PR #1031, opened 2015-12-16
Optionally nocanon

mariodsantana:
[Resubmission of pull #961 after rebasing for more alternatives.]

These two small patches fix issues I'm having when trying to do some tricky proxy actions. For example if I do:

SecDefaultAction "phase:2,log,setvar:SESSION.badness=1,proxy:'http://192.168.190.12%{REQUEST_URI_RAW}"

I've been applying them to new releases without issue for my use case. I posted them a while back as "issues" but they never got answered there? Submitting as a pull request now in case that's the preferred method. The fixes are:
1. when executing disruptive actions, it's necessary to execute the intercept action as well. I don't think this will break anything, but it certainly fixes this scenario where mod_proxy doesn't expand the REQUEST_URI_RAW macro. Maybe we should only do this for proxy actions?
2. give the user the option not to "canonicalize" the proxy URL - if we want to send the URI unchanged to the backend server to whom we are proxying, specify "[nocanon]" as the first characters in the target URL.

### DIFF
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -2513,6 +2513,11 @@ static void msre_perform_disruptive_actions(modsec_rec *msr, msre_rule *rule,
             }
         }
     }
+    if (actionset->intercept_action_rec->metadata->type == ACTION_DISRUPTIVE) {
+        if (actionset->intercept_action_rec->metadata->execute != NULL) {
+            actionset->intercept_action_rec->metadata->execute(msr, mptmp, rule, actionset->intercept_action_rec);
+        }
+    }
 
     /* If "noauditlog" was used do not mark the transaction relevant. */
     if (actionset->auditlog != 0) {
--- a/tests/regression/action/00-disruptive-actions.t
+++ b/tests/regression/action/00-disruptive-actions.t
@@ -476,6 +476,37 @@
 },
 {
 	type => "action",
+	comment => "nocanon proxy in phase:1 (get)",
+	conf => qq(
+		SecRuleEngine On
+		SecRequestBodyAccess On
+		SecResponseBodyAccess On
+		SecResponseBodyMimeType null
+		SecRule REQUEST_URI "\@streq /test2.txt" "phase:1,proxy:'[nocanon]http://$ENV{SERVER_NAME}:$ENV{SERVER_PORT}/test.txt',id:500005"
+	),
+	match_log => {
+		error => {
+			apache => [qr/ModSecurity: Access denied using proxy to \(phase 1\)/, 1],
+			nginx => [qr/ModSecurity: Access denied with code 500 \(phase 1\) \(Configuration Error: Proxy action to .* requested but proxy is only available in Apache version\)./, 1],
+		},
+	},
+	match_response => {
+		status => {
+			apache => qr/^200$/,
+			nginx => qr/^500$/,
+		},
+		content => {
+			apache => qr/^TEST$/,
+			nginx => qr/^*$/,
+		},
+	},
+
+	request => new HTTP::Request(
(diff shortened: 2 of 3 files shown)


apache2/re.c
```diff
@@ -2513,6 +2513,11 @@ static void msre_perform_disruptive_actions(modsec_rec *msr, msre_rule *rule,
             }
         }
     }
+    if (actionset->intercept_action_rec->metadata->type == ACTION_DISRUPTIVE) {
```
zimmerle: "proxy-nocanon" should be checked here, right?

tests/regression/action/00-disruptive-actions.t
```diff
@@ -476,6 +476,37 @@
 },
 {
 	type => "action",
+	comment => "nocanon proxy in phase:1 (get)",
```
zimmerle: Great! Those test cases are very helpful. ;)

zimmerle:
Hi @mariodsantana,

Thank you for the patch. I made some comments inline in the code.

uxbod:
Hello and a Happy New Year!

Am running NGINX 1.9.9 with the nginx_refactored branch in a proxy configuration.  For some reason, most probably configuration, the following rule never appears to get applied:

SecRule REMOTE_ADDR "@pmFromFile /etc/asl/whitelist" phase:1,pass,t:none,id:318743,nolog,skipAfter:END_RBL

Even though I have added the IP to the whitelist file it is still hitting this rule:

SecRule REQUEST_HEADERS:Host "^[\d.:]+$" \
"chain,phase:2,rev:2,t:none,pass,msg:'Atomicorp.com WAF Rules: Suspicious activity detected - Host header is a numeric IP address', severity:'2',id:'331032',severity:'5',tag:'no_ar'"
SecRule REMOTE_ADDR "!@ipMatch 127.0.0.1,::1"  "t:none"

Any thoughts as to why its being missed in phase 1 ?

Thanks, Phil

apache2/re.c
mariodsantana: I don't think so...?  Here's my understanding:

The loop before this line looks through all the actions in `actionset->actions` and run the execute callback for each.  For some reason, though, it doesn't seem like the `actionset->intercept_action_rec` is part of the list in `actionset->actions` -- so I added a few lines to check whether `actionset->intercept_action_rec` is an intercept action, and if so to run the execute callback for it.

tests/regression/action/00-disruptive-actions.t
mariodsantana: Just trying to be a good citizen.  I appreciate all the work you guys do!  :)

mariodsantana:
Thanks for the feedback, @zimmerle - I've responded to your comments.

apache2/re.c
zimmerle: I've got your point. You are right.

tests/regression/action/00-disruptive-actions.t
zimmerle: Thanks ;)

zimmerle:
Hi @mariodsantana,

Thank you again for you path. 

Your patch is now on our buildbots. Made a small modification, as you can see here:
https://github.com/SpiderLabs/ModSecurity/commit/c711808ef73a16df830cebf3e381846510942ed3

zimmerle:
Documentation was changed already: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#proxy

zimmerle:
Merged! It will be released as part of ModSecurity v2.9.1

mariodsantana:
Thanks!

PR #1038, opened 2015-12-29
Update readme to use shell syntax

pwnbus:
Looks like in addition to updating the readme shell syntax, my editor killed some trailing whitespace :)


csanders-git:
Thanks your merge has been pulled.

PR #1040, opened 2016-01-07
Updated geoip m4 file to support --with-geoip properly

csanders-git:


zimmerle:
Hi @csanders-git,

merged! Thanks for the patch ;)

PR #1041, opened 2016-01-07
Checked return value of hashing function so we can see if injection failed.

csanders-git:
inject_hashed_response_body() never checks its return value and as a result will always report that it was successful at injection the hash into the response body, even if it wasn't. This pull fixes that issue but allows the request to continue with just a warning put into the logs (level 1).


zimmerle:
Hi @csanders-git,

Merged! thanks.

PR #1051, opened 2016-01-12
Fix error on older systems

csanders-git:
Add support for older versions as noted in https://www.gnu.org/software/libtool/manual/html_node/Invoking-libtoolize.html


zimmerle:
Hi @csanders-git,

Merged ;) thanks.

PR #1052, opened 2016-01-12
Adds a spec file for building an RPM package

antonyh:
I needed to create an RPM file to install ModSecurity in environments where it's not possible to build from source and where there's no available RPM that is suitable. I figured the best place to share this is here so that others can benefit, but also so it can be reviewed and corrected, and I couldn't find any guidelines on what you consider acceptable for pull requests. 

It's a single RPM spec file, installs/uninstalls cleanly, tested on Centos 6.6. I hope this is a useful addition to ModSecurity, and if there's alterations needed I'm happy to help.

### DIFF
--- a/rpmbuild/mod_security.spec
+++ b/rpmbuild/mod_security.spec
@@ -1,0 +1,45 @@
+Summary: ModSecurity module for Apache HTTP Server
+Name: mod_security
+Version: 2.9.0
+Release: 1%{?dist}
+License: ASL 2.0
+URL: http://www.modsecurity.org/
+Group: System Environment/Daemons
+Source: http://www.modsecurity.org/download/modsecurity-%{version}.tar.gz
+BuildRoot: %{_tmppath}/%{name}-%{version}-%{release}-root-%(%{__id_u} -n)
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @antonyh,

Thank you for the patch. We don't usually host the specs file into our repository. But that is may be a good idea. Like you said it will be very helpful for the community in general. 

I am going to test the package generation.

zimmerle:
Hi @antonyh,

Thank you again for the pull request. We had our community meeting this week, where we decide to have a wiki page to host everything related to the package creation, including the recipes. 

The wiki page is available here:
https://github.com/SpiderLabs/ModSecurity/wiki/Package-Generation

I already placed you rpm recipe there! Thanks!

If you can, please fill the information about how to actually generate the packages, it will be very helpful to the community.

antonyh:
I've updated the wiki based on memory (and shell history) - it should be accurate. I'll double check it later to confirm and update if needed. I've also added how to install, as it wasn't obvious to me how to get rpm to pull in dependencies.

Thank you for taking the time to review this and to share it with the community.

PR #1060, opened 2016-01-29
This is fix for reborn of https://github.com/SpiderLabs/ModSecurity/i…

eik00d:
https://github.com/SpiderLabs/ModSecurity/issues/334 - has been reborn in new Apache 

This bug has been reborn, because Apache (at least in RedHat/CentOS) since version 2.2.15-47
returns in same case (connection closed before all data has been sent, so received data size is is less then Content-Length) APR_INCOMPLETE  (not APR_EOF, like before).

Based on same patch I have added handler for APR_INCOMPLETE.

### DIFF
--- a/apache2/apache2_io.c
+++ b/apache2/apache2_io.c
@@ -208,6 +208,9 @@ apr_status_t read_request_body(modsec_rec *msr, char **error_msg) {
              *      too large and APR_EGENERAL when the client disconnects.
              */
             switch(rc) {
+                case APR_INCOMPLETE :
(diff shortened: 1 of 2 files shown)


zimmerle:
Hi @eik00d,

Thanks for the patch. Delivered to the buildbots. Waiting for the results.

zimmerle:
Hi @eik00d, your patch was merged. It will be released as part of version 2.9.1

Thanks.

eik00d:
It will not be released in 2.9.1?

zimmerle:
Hi @eik00d, it is part of our mainline now. Marked to be released as part of version 2.9.2.

eik00d:
Thank you for the response!

PR #1071, opened 2016-02-05
Make url path absolute for SecHashEngine only when it is relative in the first place. Fix #752

### LINKED ISSUE #752: SecHashing href rewrites to invalid href
I am serving this html in the directory meta on an apache behind nginx with modsecurity 2.8:

``` html
<html>
<body>
<a href="subedit.pml?testparam=123456">Edit</a>
<a href="/admin/index.pml?testparam=123456">Admin</a>
</body>
</html>
```

With these settings:
SecContentInjection On
SecStreamOutBodyInspection On
SecDisableBackendCompression On

SecHashEngine On
SecHashParam "hmac"
SecHashKey "rand" "KeyOnly"
SecHashMethodRx "HashHref" "testparam"

The html presented to the client looks like:

``` html
<html><body>
<a href="/meta/subedit.pml?testparam=123456&hmac=2e90fe01904715639c65b8c8c424e24d159bd79e">Edit</a>
<a href="/meta/admin/index.pml?testparam=123456&hmac=91318d8dd856e50fca4e86e51b840bee3b9dd5c4">Admin</a>
</body></html>
```

As you can see, both hrefs are now prefixed with "/meta/" which the 2nd href does not exist on my server and therefore breaks my application. I have sniffed the traffic directly on the apache-server, and it serves nginx/modsec with html exactly as it looks like in the original html file. So that concludes that it is modsecurity which rewrites it wrongly.

hideaki:


zimmerle:
Hi @hideaki,

Sorry for the long delay. That seems to be an interesting feature, however we also understand the value of having a relative path. What do you think about having the fixed path as a configuration option?

hideaki:
Hi @zimmerle,
This is a bug fix for #752, rather than a feature.
As per my observation and understanding, SecHashEngine today always turns paths into absolute. It even tries to absolutize already absolute path, and that is resulting in broken paths, which is #752.
This fix is about avoiding the double-absolutization by absolutizing the path only when it is still relative.

Having said that, config for keeping relative path relative sounds like a good idea to me.

zimmerle:
Merged thanks!

PR #1087, opened 2016-03-02
Avoid redundant operator invokation in multimatch rules

### LINKED ISSUE #1086: multimatch applies operator too often
Consider this rule:

```
SecRule ARGS:c ".*" "id:4,multimatch,t:none,t:lowercase,t:lowercase,t:lowercase,msg:%{matched_var}"
```

Intuitively, this rule should never invoke the operator more than twice, right?

My test cases look like this:

``` json
{
    "request": "/?c=test",
    "comment": "Only initial apply",
    "matches": [
        {"id": 4, "msg": "test"}
    ]
},
{
    "request": "/?c=Test",
    "comment": "Initial apply and after first lowercase transform",
    "matches": [
        {"id": 4, "msg": "Test"},
        {"id": 4, "msg": "test"}
    ]
}
```

The first test is successful, but the second request yields these matches ("(id, msg)"):

```
('4', 'Test'), ('4', 'test'), ('4', 'test')
```

**The operator is applied once too often if there are more multiple transforms and any of them modify the input.** This problem only occurs if multiple transforms are present. The operator is always applied with the final transform value.

Another case:

Rule:

```
SecRule ARGS:b ".*" "id:3,multimatch,t:none,t:lowercase,t:hexEncode,t:length,t:removeNulls,msg:%{matched_var}"
```

Test:

``` json
{
    "request": "/?b=Test",
    "matches": [
        {"id": 3, "msg": "Test"},
        {"id": 3, "msg": "test"},
        {"id": 3, "msg": "74657374"},
        {"id": 3, "msg": "8"}
    ]
}
```

Reported matches:

```
('3', 'Test'), ('3', 'test'), ('3', '74657374'), ('3', '8'), ('3', '8')
```

Removing the `removeNulls` transform makes the duplicate match disappear, but it shouldn't appear at all if `removeNulls` doesn't modify the input.

Am I misinterpreting the multimatch specification or is there a bug?

dstelter:
Fixes #1086

### DIFF
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -2851,7 +2851,10 @@ static apr_status_t msre_rule_process_normal(msre_rule *rule, modsec_rec *msr) {
     arr = apr_table_elts(tartab);
     te = (apr_table_entry_t *)arr->elts;
     for (i = 0; i < arr->nelts; i++) {
+        /* Variable was modified by *any* transformation */
         int changed;
+        /* Variable was modified by *last applied* transformation (needed by multimatch) */
+        int tfnchanged;
         int usecache = 0;
         apr_table_t *cachetab = NULL;
         apr_time_t time_before_trans = 0;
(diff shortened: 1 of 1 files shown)


zimmerle:
The original discussion on that matter is being held at #1086

zimmerle:
As explained at #1086, this is now closed.

Issue #971, opened 2015-11-13
Implement the support to t:removeCommentsChar into libmodsecurity


zimmerle:
Transformation **removeCommentsChar** needs to be implemented inside libmodsecurity.

The documentation of the transformation can be found here:
https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#removeCommentsChar

The file already exist in the repository:
https://github.com/SpiderLabs/ModSecurity/blob/libmodsecurity/src/actions/transformations/remove_comments_char.cc

zimmerle:
Covered by pull request #1098

PR #1098, opened 2016-03-17
added remove_comments_char transformation to address issue #971

bjh7242:
Implemented remove_comments_char transformation for libmodsecurity.

Removes the following characters

- /*
- */ 
- 
- # 

These are the characters that the function msre_fn_removeCommentsChar_execute in https://github.com/SpiderLabs/ModSecurity/blob/master/apache2/re_tfns.c removes.

### DIFF
--- a/src/actions/transformations/remove_comments_char.cc
+++ b/src/actions/transformations/remove_comments_char.cc
@@ -37,16 +37,29 @@ RemoveCommentsChar::RemoveCommentsChar(std::string action)
 
 std::string RemoveCommentsChar::evaluate(std::string value,
     Transaction *transaction) {
-    /**
-     * @todo Implement the transformation RemoveCommentsChar
-     */
-    if (transaction) {
-#ifndef NO_LOGS
-        transaction->debug(4, "Transformation RemoveCommentsChar " \
-            "is not implemented yet.");
-#endif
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @bjh7242,

Thanks for the patch. I will try to have a look on it today.

zimmerle:
Hi @bjh7242,

Your pull request is now merged. Thanks again for patch.

Few comments:

Usually, before add a functionality to version 3 we expect to have regression tests that covers its basic usage, plus a few corner cases. Those unit tests are very important, among of other things, those tests are executed by our buildbots, so we can make sure that the functionality is really working in all supported platforms.

Other detail that you may want to consider is the fact that we try to not use lines bigger than 80 characters. 

Changes that I have made:
 - Unit tests:
  https://github.com/SpiderLabs/secrules-language-tests/commit/ee477d117430f695fd556afa7132de923ccd5d76
 - Coding style: 
https://github.com/SpiderLabs/ModSecurity/commit/0a60924c8f731f2a08484215388b920cd6ac89bd
 - Fix in the parser to make the functionality really available to the end user:
https://github.com/SpiderLabs/ModSecurity/commit/6f93563fc288739abf65c6db743f71deb11e6c25

Issue #1107, opened 2016-03-30
PATCH not supported on IIS


culyerr:
PATCH verb is not support on IIS.
I have forked and will be looking to fix.

victorhora:
Closing due to https://github.com/SpiderLabs/ModSecurity/pull/1108.

PR #1108, opened 2016-03-30
Added PATCH support

culyerr:
Adds PATCH support for IIS - relates to #1107

### DIFF
--- a/iis/mymodule.cpp
+++ b/iis/mymodule.cpp
@@ -1007,6 +1007,16 @@ CMyHttpModule::OnBeginRequest(
 		break;
 	}
 
+	if (r->method_number == M_INVALID)
+	{
+		if ((req->pUnknownVerb != NULL) && (strcmp(req->pUnknownVerb, "PATCH")==0))
+		{
+			// this is a PATCH command and should be supported
+			r->method = "PATCH";
+			r->method_number = M_PATCH;
+		}
+	}
+	
 	if(HTTP_EQUAL_VERSION(req->Version, 0, 9))
 		r->protocol = "HTTP/0.9";
 	else if(HTTP_EQUAL_VERSION(req->Version, 1, 0))


csanders-git:
This is very interesting - for my own edification - where is the PATCH method used?

culyerr:
From RFC5789
"The existing HTTP PUT method only allows a complete replacement of a document.This proposal adds a new HTTP method, PATCH, to modify an existing HTTP resource."

csanders-git:
This is interesting - i've seen the RFC - does anything use this? Just so we can test it :-P

culyerr:
We have tested this with part of our payments solution but the PATCH methods are not publicly accessible for 3rd party testing.

I guess you could test against any resource on an IIS server - without the fix if you have "detection mode = on" it will record "Guid_0.00" errors if it does not work.

zimmerle:
Relevant documentation: 
 - HTTP_REQUEST - https://msdn.microsoft.com/en-us/library/windows/desktop/aa364545(v=vs.85).aspx
 - HTTP_REQUEST_V1 - https://technet.microsoft.com/pt-br/aa364612(v=vs.71)
 - HTTP_REQUEST_V2 - https://msdn.microsoft.com/en-us/library/windows/desktop/aa364613(v=vs.85).aspx

Almamu:
@csanders-git The standard methods (GET, POST, PUT, PATCH, DELETE) are widely used in REST API interfaces, so having support to inspect those would be really helpful.

csanders-git:
@Almamu - any idea of some examples that might be around?

Almamu:
@csanders-git I don't have any publicly accesible example as I've always worked on REST API for private systems, but searching about API REST in google gives a handful of results of how it should be implemented and documentation that specifies how the standar HTTP verbs are used in it and what they mean, although there is no RFC for REST API (as long as I recall).

victorhora:
So it looks like there's still not that many apps using PATCH these days. I could only find a few APIs to test (which were easy to deploy on Apache/Nginx, but a pain to get it working under IIS). Maybe someone could give us a hand in that direction... :)

That said, by doing some trickery with IIS WebDAV module I could run a few tests with CRS and some custom rules. As PATCH is not explicitly added to the main code base of ModSecurity looks like you can't use [REQUEST_METHOD](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#REQUEST_METHOD) or [REQUEST_LINE](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#REQUEST_METHOD) variables straightforward to match with PATCH requests. But It looks like you can still match stuff with [ARGS](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#ARGS), [REQUEST_LINE](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#REQUEST_LINE) or [FULL_REQUEST](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#FULL_REQUEST) variables.

On all my testes I couldn't spot any error related to "Guid_0.00" as mentioned but I'm really not familiar with this error message. It looks like that this issue could be related with [PCRE limits](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SecPcreMatchLimit) parsing large bodies and / or XML but this is only a guess at this point. 

@culyerr @Almamu maybe if you could share the rules you have loaded, some debug logs and ideally a HTTP PATCH request which triggers this issue would be great.

What I'm trying to understand is if this pull request is actually to fix a bug specific for ModSecurity on IIS or to add a functionality (PATCH verb support) =)

zimmerle:
Closing this pull request, I will merge it once I find myself the manners to test it.

Issue #1109, opened 2016-03-30
IPv4 addresses not parsed correctly in IIS


culyerr:
Port number is included when passed to tree_contains_ip(..) and code attempts to parse as an IPv6 address.

Error is raised to Windows Event Log when doing IPmatch rules
```
[client 213.78.82.1:34984] ModSecurity: IPmatch: bad IPv6 specification "213.78.82.1:34984". [hostname "ACTUSAA1"] [uri "/api/ping"] [unique_id "16717361818946771401"]
```

zimmerle:
This issue was resolved by #1110.

zimmerle:
Related to: #734

zimmerle:
Fixed by: #1220.

PR #1110, opened 2016-03-30
Fixed IPv4 address parsing when port is present at end of address

culyerr:
Relates to #1109

### DIFF
--- a/apache2/msc_util.c
+++ b/apache2/msc_util.c
@@ -2695,7 +2695,11 @@ int ip_tree_from_uri(TreeRoot **rtree, char *uri,
 int tree_contains_ip(apr_pool_t *mp, TreeRoot *rtree,
     const char *value, modsec_rec *msr, char **error_msg)
 {
+	const char *format = "%15[0-9.]:%5[0-9]";
+	char ip[16] = { 0 };  // ip4 addresses have max len 15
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @cylyerr,

Thanks for you patch. It is currently being tested by our buildbots.

zimmerle:
Hi @culyerr,

This function "tree_contains_ip" is not only used by IIS. It is used by our different ports: for Apache, NGINX, and others. We may also expect to receive an IPv6 as input of that function.

I believe that this bug that you want to address, may related to the fact that the IP address should never ever come with the port at that stage. Most likely the update needs to be made in the IIS specific code. So that we can get rid of the port in the earlier stages of ModSecurity processing.

Testing the patch I got 2582 unit tests failing. Please, have a look at:
http://www.modsecurity.org/developers/buildbot/builders/Linux32%20-%20Nginx%20%28Stable%29/builds/255/steps/unit%20test_1/logs/stdio

zimmerle:
@culyerr if you want to discuss it, please reach me (zimmerle) or csanders on #modsecurity at FreeNode.

culyerr:
Thanks. I am picking this up again now and will look to implement the check in the IIS specific code.

zimmerle:
@culyerr Nice ;) if you need something we are at: https://modsecurity.slack.com/

PR #1123, opened 2016-04-18
Added mbed functionality

csanders-git:
added mbed open source framework
supports md5, sha1 and base64
also added hexify funcitons


zimmerle:
Merged with little modifications. The core feature is on the "libmodsecurity"  already.

PR #1151, opened 2016-05-17
Replacing include subdirectory name, transaction --> collection

manishmalik:
Hi!
From what I understood from this commit https://github.com/SpiderLabs/ModSecurity/commit/5643d2fa287ce51cf3041b0afe3b9e41daa49011, its seems like it require some minor change in `/src/Makefile.am` namely replacing `transaction` with `collection`.


zimmerle:
Hi @manishmalik,

Thank you for the patch! merged!

PR #1164, opened 2016-06-16
Use internal PCRE based implementation of regular expressions instead of std C++ regex library.

phantom-az:
With that change https://github.com/SpiderLabs/ModSecurity/commit/d83987c2fc4ca5b2dab4e3ac0d3014704116d336 we move from std c++ implementation of regex to custom (pcre driven) version of regex.  Unbreaking unit_test runtime for gcc 4.8 based OSes

### DIFF
--- a/examples/multithread_c/Makefile.am
+++ b/examples/multithread_c/Makefile.am
@@ -8,6 +8,7 @@ multi_SOURCES = \
 multi_LDADD = \
 	-L$(top_builddir)/src/.libs/ \
 	-lmodsecurity \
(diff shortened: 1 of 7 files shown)


phantom-az:
Here's a compilation fix missed on previous merge.  Required to compile multithread_c example on FreeBSD 10.  
https://github.com/SpiderLabs/ModSecurity/pull/1164/commits/aa88fd67b62fb81070a969788821f473ceb0397d

zimmerle:
Merged! Thanks!

PR #1175, opened 2016-06-30
Unbreak build with custom location of libyajl.so

phantom-az:
... with that patch https://github.com/SpiderLabs/ModSecurity/commit/c9293993270bd177aed69e8001bd2964d93481fc


zimmerle:
merged ;) thanks.

PR #1180, opened 2016-07-06
preliminary support for "make dist" target

defanator:
The commits in this request make able to use "make dist" target in order to create archive with libmodsecurity sources that can be used later without "./build.sh" step.

There are still some unresolved things like the "BUILT_SOURCES" defined in the src/Makefile.am which doesn't work: the seclang-parser.cc and seclang-scanner.cc files are coming to the final distdir regardless of the "BUILT_SOURCES" value (I have also tried things like "nodist_*" for those files, but had no luck). I.e., in the source tree extracted from the archive generated by "make dist" command, you have to run "make clean" before "make" in order to get all the things done.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -274,13 +274,16 @@ AC_CONFIG_FILES([\
     doc/Makefile \
     src/Makefile \
     others/Makefile \
-    test/Makefile \
-    test/benchmark/Makefile \
-    test/fuzzer/Makefile \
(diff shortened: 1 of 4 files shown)


zimmerle:
Hi @defanator,

Thank you for the patch! :)

We may want to run flex and bison before create the .tar.gz but that is a good starting point!

Thank you. It is merged already.

Issue #1173, opened 2016-06-28
Possible representation bug in uploads part of JSON audit log format


dune73:
I have the following entry in the JSON audit log format:

```
...
    "uploads": {
        "info": [
            "file_size",
            48458,
            "file_name",
            "example.png",
            "content_type",
            "<Unknown Content-Type>"
        ],
        "total": 48458
    }
...
```
Within the info tag, there is a list of items and not a list of key-value pairs as would be expected. It might have to do with the way the data is represented within ModSecurity (-> corresponding native audit log format part "J" is a CSV), but this seems to be a bug to me.

zimmerle:
Linking this issue with #1174, as they seems to be on the same piece of code.

p0pr0ck5:
Not the same issue as #1174, but looks like an easy fix. `msc_logging.c:1363` opens this with `yajl_gen_array_open` instead of `yajl_gen_map_open`. Not sure why I did this :p I'll have a PR for this within a day or two.

zimmerle:
Thanks @p0pr0ck5 ;)

dune73:
Thank you guys.

p0pr0ck5:
@dune73 can you try the fix in https://github.com/p0pr0ck5/ModSecurity/commit/e54b1504fc0105248b874b4f50b3f9614c347679? I won't have time to test for a bit but this should resolve the issue, small oversight on my part. We don't log file upload data so I wasn't as thorough as I should have been.

p0pr0ck5:
^ Once this is tested I'll submit the PR

dune73:
Looks good here:

```
    "uploads": {
        "info": [
            {
                "content_type": "<Unknown Content-Type>",
                "file_name": "tmp1",
                "file_size": 457
            }
        ],
        "total": 457
    }

```

Thank you for the quick fix.

zimmerle:
Pull request #1181 was merged! thanks!

PR #1181, opened 2016-07-06
Fix file upload JSON audit log entry

p0pr0ck5:
Each uploaded file is a separate yajl array, but we forgot to open
the a map for the proper k/v pairs.

This fixes issue #1173.


zimmerle:
Thanks @p0pr0ck5  and @dune73. Merged!

Issue #1174, opened 2016-06-28
Msg/Logdata block misrepresented in JSON audit log matched_rules block


dune73:
The following rule:

```
SecRule ARGS:test "."	"id:1006,phase:2,pass,log,msg:'HIT: %{MATCHED_VAR}',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}'"
```
Results in the following audit log block:
```
...
        {
            "chain": false,
            "rules": [
                {
                    "actionset": {
                        "id": "1006",
                        "is_chained": false,
                        "logdata": "Matched Data: /etc/passwd found within REMOTE_ADDR: 127.0.0.1",
                        "msg": "HIT: 127.0.0.1",
                        "phase": 2
                    },
                    "config": {
                        "filename": "/apache/conf/httpd.conf_pod_2016-06-28_06:13",
                        "line_num": 233
                    },
                    "is_matched": true,
                    "operator": {
                        "negated": false,
                        "operator": "rx",
                        "operator_param": ".",
                        "target": "ARGS:test"
                    },
                    "unparsed": "SecRule \"ARGS:test\" \"@rx .\" \"phase:2,id:1006,pass,log,msg:'HIT: %{MATCHED_VAR}',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}'\""
                }
            ]
        },
...
```
msg and logdata are obviously wrong. 

If you look at the native alert, it is perfectly ok:
```
[2016-06-28 06:19:00.628478] [-:error] 127.0.0.1:59413 V3H6tH8AAQEAADNUhysAAAAA [client 127.0.0.1] ModSecurity: Warning. Pattern match "." at ARGS:test. [file "/apache/conf/httpd.conf_pod_2016-06-28_06:13"] [line "233"] [id "1006"] [msg "HIT: /etc/passwd"] [data "Matched Data:  found within ARGS:test: /etc/passwd"] [hostname "localhost"] [uri "/index.html"] [unique_id "V3H6tH8AAQEAADNUhysAAAAA"]

```

zimmerle:
Hi @dune73,

Thanks for the report. Let me add @p0pr0ck5 to the loop, he may know by heart what is going on.

p0pr0ck5:
[...]

PR #1190, opened 2016-07-19
Remove logdata and msg fields from JSON audit log rule elements

p0pr0ck5:
Writing macro-expanded strings to JSON elements during the post-logging
phase can be misleading, because it's possible that variable contents
(such as MATCHED_VAR) could have changed after the rule match, altering
their expected contents. Writing macro-epanded audit data really only
makes sense when the macros are expanded immediately following the
rule match. See issue #1174 for more details.


zimmerle:
Merged.

PR #1200, opened 2016-08-08
Include pre-generated parser sources into the distribution

defanator:
This patch addresses an issue that has not been resolved here:
https://github.com/SpiderLabs/ModSecurity/pull/1180

It allows to produce fully prepared source archive using the "make dist" target.


zimmerle:
Hi @defanator 

thank you for the patch. Merged.

PR #1231, opened 2016-10-18
SecLang uses RESPONSE_STATUS as variable, not STATUS

abhijoglekar:
Seclang uses RESPONSE_STATUS as variable to encode the status code for the
request.
https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#RESPONSE_STATUS

The CRS v3.0.0-dev rules, for instance, uses the RESPONSE_STATUS variable.
https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.0.0-dev/rules/RESPONSE-50-DATA-LEAKAGES-IIS.conf

When processing response headers, the variable was named STATUS when creating/storing
it in the collection. Fix it, and update regression testcases.

### DIFF
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -876,7 +876,7 @@ int Transaction::processResponseHeaders(int code, const std::string& proto) {
 #endif
 
     this->m_httpCodeReturned = code;
-    this->m_collections.store("STATUS", std::to_string(code));
+    this->m_collections.store("RESPONSE_STATUS", std::to_string(code));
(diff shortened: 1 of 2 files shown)


abhijoglekar:
@zimmerle , created a fork of the repo and sent a PR for a fix in libmodsec. Please take a look and let me know if this method works for contributions. Thanks.

zimmerle:
Hi @abhijoglekar,

Thank you for your patch! It is already merged.

That is a perfect way to send us contribution!

PR #1249, opened 2016-11-05
~Rule will delete chainedRule

dkamen:
Came across this memory leak when reloading nginx with hundreds of rule chains


zimmerle:
Hi @dkamen,

Thank you for the patch.

zimmerle:
Merged.

PR #1256, opened 2016-11-09
Fixed install and dist targets to include actions/action.h

defanator:
This is needed due to movement of actions/action.h in 3ee7b24.


zimmerle:
Thanks @defanator. Merged.

PR #1257, opened 2016-11-10
 Fix return value of msc_rules_merge()  and fix typos

soonum:
Hi all,

Regarding to the documentation msc_rules_merge() should
return the number of merged rules instead of 0 in all cases.

Fix typos in documentation.


zimmerle:
Hi @soonum,

Thank you for the patch ;)

zimmerle:
Merged!

PR #1258, opened 2016-11-11
Czechia replaces Czech Republic

theforcedk:
As of 17 May 2016, Czech Republic is now named Czechia.


zimmerle:
Hi @theforcedk thanks for the patch! merged.

PR #1270, opened 2016-11-22
Add (void) argument to be ANSI C compliant

soonum:
Add void argument has to avoid warning messages when compiling Python bindings with CFFI since it uses -Wstrict-prototypes option by default.

Modify`msc_create_rules_set` and `msc_init` internals, now it returns directly an instance like `msc_new_transaction` in transaction.cc.


zimmerle:
Hi @soonum,

Thanks for the patch, merged!

PR #1271, opened 2016-11-23
Use correct debugging CFLAGS in readme

defanator:


zimmerle:
Hi @defanator, thanks for the patch! merged.

Issue #1273, opened 2016-11-24
SESSION.counter is inaccurate


csutherl:
Using the [SESSION collection configuration](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SESSION) does not yield expected results for the SESSION counter, however the global counter works as expected. This issue affects both prefork and worker MPMs. Additionally, the tests that I am doing are all after applying the fix from PR #1224 because that PR corrects a lot of the concurrency issues that existed prior to the fix. One more piece of important information is that I can only seem to reliably reproduce this behavior on a physical machine; I guess because it's a concurrency issue that slower VMs I'm using just don't hit the race condition.

To reproduce, use the following mod_security configuration, start httpd with mod_security enabled, and make some requests. 

~~~
# SESSION 
SecRule REQUEST_COOKIES:SESSIONID !^$  phase:5,id:118,nolog,pass,setsid:%{REQUEST_COOKIES.SESSIONID}
SecAction phase:5,id:119,nolog,pass,setvar:SESSION.my_counter=+1
~~~

I'm using the following loop to make 500 requests using five different sessions.

~~~
for i in `seq 1 500`;do
    curl -b "SESSIONID=testing$(shuf -i 1-5 -n 1)" localhost &>/dev/null &
done
~~~

After the requests are done, I'm checking the collection with modsec-sdbm-util to see if the counter values add up to the total number of requests, and they don't (I usually get ~490). Example:

~~~
# modsec-sdbm-util /var/lib/mod_security/default_SESSION -du | egrep '(^Key|my_counter)'
Key: "testing4", Value len: 212
                    my_counter: 99
Key: "testing2", Value len: 214
                    my_counter: 103
Key: "testing1", Value len: 214
                    my_counter: 104
Key: "testing5", Value len: 214
                    my_counter: 100
Key: "testing3", Value len: 212
                    my_counter: 90
~~~

The total number of all the my_counter variables should be equal to the number of requests.

csutherl:
Tagging @p0pr0ck5 here since we talked a bit about it on IRC.

zimmerle:
Hi @csutherl,

That is kind the expected behavior. To avoid the utilization of locks, each process (or threads) has it own version of the collection that is synced every once in a while. The idea is to reduce the idle wait caused by the utilization of a lock. With that strategy we got a closer value without have to hold the request for too long.

[...]

PR #1274, opened 2016-11-24
Hack to show that there is an issue with the session collection's logic

csutherl:
This PR is a hack to show that there is an issue somewhere with the collection storage/retrieval logic for session collections (I also think this affects user and resource collections, but haven't verified this yet) which is causing #1273. After applying this change the test in #1273 always adds up to the total number of requests. I arrived at this conclusion because I noticed some inconsistencies in the logging, like:
~~~
Added collection "default_SESSION" to the list as "SESSION".
....
collection_retrieve_ex: collection_retrieve_ex: Retrieving collection (name "default_SESSION", filename "/var/lib/mod_security/default_SESSION")
~~~
and that the SESSION collection never receives delta updates like the global one does:
~~~
collection_store: Delta applied for global.counter 1->2 (1): 1 + (1) = 2 [2,1]
~~~
because orig_col in re_actions.c always returns NULL for the session collection:
~~~
458     orig_col = (const apr_table_t *)apr_table_get(msr->collections_original, var_name->value);
~~~

The only difference that I see in the way that the global and session collections are handled is in the real_col_name and col_name passed into init_collection.

Obviously I don't want this to be accepted, but I wanted to get a conversation started with someone that understands mod_security better than I do and post the results of my testing thus far. I was told that a PR is the best way to do that :)

### DIFF
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -223,7 +223,13 @@ static void modsecurity_persist_data(modsec_rec *msr) {
 
         /* Only store those collections that changed. */
         if (apr_table_get(msr->collections_dirty, te[i].key)) {
-            collection_store(msr, col);
+            if (msr->txcfg->debuglog_level >= 9) {
+                msc_string *name = (msc_string *)apr_table_get(col, "__name");
+                if (name != NULL) {
+                    msr_log(msr, 9, "Storing dirty collection %s at key %s", name->value, te[i].key);
+                }
+            }
+            collection_store(msr, col, te[i].key);
         }
     }
 
--- a/apache2/persist_dbm.c
+++ b/apache2/persist_dbm.c
@@ -321,7 +321,7 @@ apr_table_t *collection_retrieve(modsec_rec *msr, const char *col_name,
 /**
  *
  */
-int collection_store(modsec_rec *msr, apr_table_t *col) {
+int collection_store(modsec_rec *msr, apr_table_t *col, const char *col_name) {
     char *dbm_filename = NULL;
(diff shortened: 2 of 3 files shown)


csutherl:
@bostrt

csutherl:
Oh yeah, after this change is applied you have to update the configuration to use 'default_SESSION' instead of 'SESSION' (which is why I said this is a hack) because it tries to lookup the exact name you give it there.

~~~
SecAction phase:5,id:119,nolog,pass,setvar:default_SESSION.my_counter=+1
~~~

csutherl:
I left the original hacky commit in the PR (reverted) to show exactly what logic I think is flawed, but the newer commit looks like a good fix to me. Feedback appreciated!

zimmerle:
Just to add more details...

Not sure if you have noticed but the **deafult_** that you are referring to, is in fact the name of the _WEBAPPID_. The idea behind the WebAppID is to create a namespace for the collection. There are further information about the **SecWebAppId** here:

- https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SecWebAppId
- https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#WEBAPPID

As I've commented before, regardless of the computation of the delta, it is expected to have a _conter_  on the top of a collection with values that are not the exactly same amount of requests. That may differ upon to the limitations of the server vs the amount of requests.

zimmerle:
As described at #1273, this issue has being fixed therefore we no longer need this hack to expose the problem. Thank you for the report @csutherl  :)

Issue #1278, opened 2016-11-30
Questionable block of code in collection persistence


bostrt:
Is this duplicate code or was this done on purpose? Looks like this can be compacted down to just `blob_size += len * 2 + 4`. 

https://github.com/SpiderLabs/ModSecurity/blob/master/apache2/persist_dbm.c#L482

`````
        len = var->name_len + 1;
        if (len >= 65536) len = 65536;
        blob_size += len + 2;

        len = var->value_len + 1;
        if (len >= 65536) len = 65536;
        blob_size += len + 2;
`````

zimmerle:
Indeed. I don't see a reason for that. It seems to be like that for a quite some time. 

Check out the git blame:
https://github.com/SpiderLabs/ModSecurity/blame/master/apache2/persist_dbm.c#L482

Want to send a patch for it?

bostrt:
Tagging @csutherl who has done some testing with a patch

bostrt:
Heh, so those blocks are different. One is for `var->name_len` and other is for `var->value_len`. I think @csutherl is going to submit a code comment addition.

bostrt:
Closing out since this is not a real issue. Would appreciate if #1279 still got accepted just to clear up anyone else who might stumble over same code.

PR #1279, opened 2016-12-02
Adding comments around odd looking code to prevent future scrutiny

csutherl:
This comment is a result of #1278 :) Adding a comment will make it a bit easier to identify the difference because the duplicated block looks like an error.


zimmerle:
Hi @csutherl,

Thank you for the patch. Sorry for the delay.

PR #1283, opened 2016-12-09
Fix building with -Wl,--as-needed linker option with older ld versions

defanator:
In particular, this allows to build libmodsecurity on Debian 7 "wheezy" (it has ld version 2.22 out of the box).

Before this change, the following errors appeared during the linking:

    /usr/bin/ld: ../../src/.libs/libmodsecurity.a(libmodsecurity_la-system.o): undefined reference to symbol 'clock_gettime@@GLIBC_2.2.5'
    /usr/bin/ld: note: 'clock_gettime@@GLIBC_2.2.5' is defined in DSO /usr/lib/x86_64-linux-gnu/librt.so so try adding it to the linker command line
    /usr/lib/x86_64-linux-gnu/librt.so: could not read symbols: Invalid operation
    collect2: error: ld returned 1 exit status

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -38,14 +38,14 @@ noinst_HEADERS = \
        regression/*.h
 
 unit_tests_LDADD = \
-	$(GLOBAL_LDADD) \
 	$(top_builddir)/src/.libs/libmodsecurity.a \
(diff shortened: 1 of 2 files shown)


zimmerle:
Merged! thanks!

PR #1301, opened 2017-01-09
 Add -K- parts support.

phantom-az:
Here's last year's latest part of audit_log updates.  -K- parts support.


zimmerle:
Thank @phantom-az!

Saving a request on auditlog does not depends only in the auditlog action. Currently this pull request contains conflicts, i am closing this and it is going to be used as a base for the implementation on the -K- entry in the auditlogs.

Issue #992, opened 2015-12-08
The action 'drop' does not work with mod_http2


sidney-j-r-m:
The action 'drop' does not work with mod_http2.

zimmerle:
Hi @sidney-j-r-m,

Can you give us more details?

What the debug logs says?

It is only about the "drop" action?

sidney-j-r-m:
Error: Connection drop requested but failed to close the  socket
ModSecurity: Access denied with code 500 (phase 1)

Modules: mod_http2 and mod_security(both compiled from source)
Apache version: 2.4.18-dev(compiled from source)
System: CentOS Linux

Log:
/var/log/httpd/error.****.log:[Sat Nov 28 21:29:32 2015] [error] [client
*.*.*.*] ModSecurity: Access denied with code 500 (phase 1) (Error:
Connection drop requested but failed to close the  socket). Operator EQ
matched 1 at IP:dos_block. [file
"****/modsecurity_crs_11_dos_protection.conf"] [line "*****"] [id "981045"]
[hostname "*****"] [uri "/*****"] [unique_id "VlpG7H8AAAEAAE9fewUAAAAQ"]

On Wed, Dec 9, 2015 at 10:57 AM, Felipe Zimmerle <notifications@github.com>
wrote:

> Hi @sidney-j-r-m <https://github.com/sidney-j-r-m>,
>
> Can you give us more details?
>
> What the debug logs says?
>
> It is only about the "drop" action?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpiderLabs/ModSecurity/issues/992#issuecomment-163244247>
> .
>



-- 
Sidney

sidney-j-r-m:
Only "drop" action is affected. The "deny" action works fine.

zimmerle:
Hi @sidney-j-r-m,

The "drop" happens to work without the mod_http2 ? Same scenario and rules, but without the mod_http2?

sidney-j-r-m:
"drop" works fine without mod_http2.

bazzadp:
This was discussed on the [mod_http2 project issue list too](https://github.com/icing/mod_h2/issues/127), and seems the fix is as simple as [changing line 269 of mod_security2.c](https://github.com/SpiderLabs/ModSecurity/blob/master/apache2/mod_security2.c#L269) from this:
```
                apr_socket_t *csd = ap_get_module_config(msr->r->connection->conn_config,
                    &core_module);
```
to this:
[...]

PR #1308, opened 2017-01-21
This is a fix for #992 to allow drop to work with mod_http2

bazzadp:
As discussed here: https://github.com/SpiderLabs/ModSecurity/issues/992 and here: https://github.com/icing/mod_h2/issues/127.


zimmerle:
Hi @bazzadp,

Thank you for the patch. It was merged into v2/master.

Notice that I've made this fix here:  9b3c32bb54935b232b0a3bbf14625b8409b926d7
It was necessary as ```master``` was added on version 2.4.18

PR #1332, opened 2017-02-22
Don't leak an fd on fuzzy hash initialization

p0pr0ck5:
Since we're re-opening this file with every invocation, let's
close our sanity check fd.


zimmerle:
Closing this issue as I am reviewing it together with #1339.

Issue #1288, opened 2016-12-20
ssdeep + SecUploadKeepFiles On = Memory leaks?


meatlayer:
Hello,

After enabling SecUploadKeepFiles in the apache configuration to use [fuzzy hash (ssdeep)](https://www.trustwave.com/Resources/SpiderLabs-Blog/ModSecurity-Advanced-Topic-of-the-Week--Detecting-Malware-with-Fuzzy-Hashing/) verification, i found crazy memory leaks. 

```
Apache/2.4.10 (Debian) mpm-itk/2.4.7-02 OpenSSL/1.0.1t
ModSecurity for Apache/2.9.1
```

I tried to change the settings for these options in the config. I increased values, then do reduced values. But it had no effect. The leak has not disappeared.

    SecRequestBodyInMemoryLimit 512000
    SecRequestBodyNoFilesLimit 512000
    SecResponseBodyLimit 32000000

Below you can see how the increased memory consumption of an apache process at load time CSV file with a maximum size of 15MB:

```
============================================
LA=94
PIDS=2
CPU=7.2%
MEM=79.9%
SQL=0%

PID USER PR NI VIRT RES SHR S %CPU %MEM TIME+ UID COMMAND (TOP)
-----------------------------------------------------------------------
9773 to*y 30 10 28.344g 0.025t 3520 D 7.2 79.9 0:53.49 1652 /usr/sbin/apache253 -D php53 -k restart
4332 to*y 20 0 110728 2052 0 S 0.0 0.0 0:00.12 1652 proftpd: to*y - 82.193.139.*: IDLE


Srv PID Acc M CPU SS Req Conn Child Slot Client VHost Request (APACHE)
-----------------------------------------------------------------------
13-0 16769 0/952/2951 W 0.04 142 0 0.0 17.50 55.07 82.193.139.* lamp***b.ru:8083 POST /admin/index.php?route=catalog/suppler/start&token=19c5a91

============================================

LA=7
PIDS=2
CPU=100%
MEM=49.6%
SQL=0%

PID USER PR NI VIRT RES SHR S %CPU %MEM TIME+ UID COMMAND (TOP)
-----------------------------------------------------------------------
32550 u23**5 30 10 16.503g 0.015t 4648 R 100.0 49.5 0:26.41 1870 /usr/sbin/apache2 -k restart
[...]

PR #1339, opened 2017-03-03
Read fuzzy hash db on init

p0pr0ck5:
Instead of reading the fuzzy db on every invocation, read and store
the db contents during initialization and store the contents in memory.
The only significant behavior change here is that a change in db contents
now (obviously) requires a daemon restart, as no API is provided to
flush the list of ssdeep chunks.

### DIFF
--- a/apache2/re.h
+++ b/apache2/re.h
@@ -409,8 +409,14 @@ struct msre_cache_rec {
     apr_size_t               val_len;
 };
 
+struct fuzzy_hash_chunk {
+    const char *data;
+    struct fuzzy_hash_chunk *next;
+};
+
 struct fuzzy_hash_param_data {
     const char *file;
+    struct fuzzy_hash_chunk *head;
     int threshold;
 };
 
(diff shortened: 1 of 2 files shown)


p0pr0ck5:
This also obviates #1332, so we can nix that one if this is worthwhile.

p0pr0ck5:
(re-did this to insert elements to the tail of the list upon initialization, so that the file contents are processed top-down during execution, matching the existing behavior).

Rendername:
@p0pr0ck5 Hello.
I also encountered memory leaks when using fuzzy hash.
[Issue #1288 ](https://github.com/SpiderLabs/ModSecurity/issues/1288)
Interested. You think your changes really can help to get rid of the leaks?

p0pr0ck5:
No, this PR is not about a memory leak. This just removes the need to re-read the fuzzy hash db from disk on every invocation. Your noted issue is unrelated to this PR.

zimmerle:
Pull request is currently under our buildbots - http://www.modsecurity.org/developers/buildbot/waterfall

zimmerle:
Thanks @p0pr0ck5  -- merged ;)

PR #1343, opened 2017-03-07
Fix "make dist" after recent changes to parser

defanator:
In particular, it is now possible to either build ModSecurity with pre-generated parser, or use `--enable-parser-generation` configure option to rebuild parser from sources.


zimmerle:
Merged! thanks ;)

Just to notice: the Travis failure has nothing to do with the commit, it was an internal error.

PR #1371, opened 2017-03-31
Allow mod_session to populate environment before rule execution

nickmarden:


zimmerle:
Hi @nickmarden,

Thank you for the patch!

The normal way to change this execution order is by having the modules loaded into a different order at the Apache configuration file. However, this is not the case for this particular hook as ModSecurity is marked to be an **earlier** hook.

The negative side of this patch is its benefit. Let me explain: We may not be able to block a request before it hit the mod_session. Tracking session inside ModSecurity is something doable via [SESSION](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SESSION) schema. According to ```mod_session.c``` documentation , I am assuming that you are willing to read the session values via ENV vars, is that correct? Can you described a little bit more about your use case scenario? 

Just for the record: this patch affect phase 2 in ModSecurity, the normal order is to have ModSecurity inspecting the content prior to ```mod_session.c``` this patch revert this order. Other phases are not affected. Phase 3 has access to the ENV vars [populated with mod_session.c] regardless of this patch.

csanders-git:
how does this affect processing at phase 2, i'm unclear.

zimmerle:
Hi @csanders-git,

Before the patch:

```
ModSecurity -> mod_session.cc
```

After the patch:

```
mod_session.cc -> ModSecurity
```

It means that ModSecurity will inspect a request after it be processed by ```mod_session.c```. That would be during the phase 2.

@csanders-git is that clear?

csanders-git:
Interesting, I think this generally makes sense, we want to receive a processed request -- However, this can make rules incompatible with different servers.

nickmarden:
Hi @zimmerle @csanders-git thanks for the feedback. I'm going to review the information on [SESSION](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SESSION) and see if it can be used to implement our use case. If so, I'll simply withdraw this PR. If not, I'll provide a clearer explanation of where the shortcoming is.

zimmerle:
Thank you @nickmarden!

nickmarden:
@zimmerle @csanders-git The reason we wanted this change was that we were trying to implement the following:
- A user exceeds some leaky-bucket request limit implemented in ModSecurity+lua and is redirected to a CAPTCHA challenge page.
- The backend server indicates that a user has successfully completed a CAPTCHA challenge by setting a response header.
- We use mod_session's [SessionHeader(https://httpd.apache.org/docs/2.4/mod/mod_session.html#sessionheader) directive in order to intercept the backend server's response header, storing the fact that the user is not a bot in the user's (encrypted) mod_session-managed session.
- During subsequent requests, this change allows us to examine the already-decrypted mod_session session contents to see if the user should be exempted from any more rate-limiting/CAPTCHA challenges.

Can you think of a way to accomplish something like this without the change we propose?

csanders-git:
I can think of ways but not that are as efficient as this. It'd make a good writeup once it's working. Currently ModSec's persistence is very expensive to use so isn't well suited for this.

zimmerle:
Hi @nickmarden,

Let me close this pull request and lets have our discussion over #1372.

victorhora:
@zimmerle I'm wondering if this is deemed important for some users we might think about adding a compile time option disabled by default? Might be simple enough (ifdef?) assuming it doesn't break other stuff I'm not predicting at this time :)

PR #1373, opened 2017-04-03
Fix building with nginx >= 1.11.11

defanator:
Closes SpiderLabs/ModSecurity#1359

See also:
http://hg.nginx.org/nginx/rev/e662cbf1b932


defanator:
This one may be incomplete.

See the corresponding PR here:
https://github.com/openresty/lua-nginx-module/pull/1017

PR #1374, opened 2017-04-03
add support for soap+xml

csanders-git:
As was talked about by @emphazer in https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/721, RFC 3902 adds support for the application/soap+xml header used by SOAP 1.2.


zimmerle:
It was also merge into v3/master.

Thank you.

Issue #1069, opened 2016-02-02
Conditional logging of "log producer"


marcstern:
Each time a rule triggers and is logged, it generates a line in the audit log with the "log producer".
This wastes a lot of disk space and is not interesting if you have a good configuration management.
We should have a directive (or a compile flag) to disable that logging.
Ex. of compile flag: #marcstern@9e68805

zimmerle:
closed. a5c3548

PR #1382, opened 2017-04-10
Option to disable logging of log producer in audit log

marcstern:
when log level < 9.
[Issue #1069 - Marc Stern]

### DIFF
--- a/apache2/msc_logging.c
+++ b/apache2/msc_logging.c
@@ -1174,6 +1174,9 @@ void sec_audit_logger_json(modsec_rec *msr) {
             yajl_kv_bool(g, "response_body_dechunked", 1);
         }
 
+#ifdef LOG_NO_PRODUCER
+        if (msr->txcfg->debuglog_level < 9) return;
+#endif
         sec_auditlog_write_producer_header_json(msr, g);
 
         /* Server */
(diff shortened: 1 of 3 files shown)


zimmerle:
Please make sure that the JSON audit log is also covered.

marcstern:
done

apache2/msc_logging.c
```diff
@@ -1174,6 +1174,9 @@ void sec_audit_logger_json(modsec_rec *msr) {
             yajl_kv_bool(g, "response_body_dechunked", 1);
         }
 
+#ifdef LOG_NO_PRODUCER
+        if (msr->txcfg->debuglog_level < 9) return;
```
zimmerle: Please check this condition here. It does not seems to be doing what you aiming to do in the patch/commit description.

zimmerle:
Hi @marcstern,

Made a comment in your patch, please verify. Before send me any updates make sure to do it in a clean tree. Please also avoid to long commit to replaced the \r with \r\n. Thanks.

apache2/msc_logging.c
marcstern: Indeed, seems to be part of another patch.
Do I recommit? It seems I cannot amend (actually, I can but it detects no difference if I try)

marcstern:
I can't see, in "Files changed", in which files \r were replaced by \r\n.
I can't find this on my local copy either.

apache2/msc_logging.c
zimmerle: Yes, please. Try to amend all the commits and make this patch a single shot. addressing those comments.

Issue #1392, opened 2017-04-20
JSON requests with no body fail as "premature EOF"


metalspawn:
I have set up the `v3/master` version with the nginx connector. Our team is using it with Centos 7.3 SE and nginx 1.10.3. Further details can be provided on request.

While running our API test suite I noticed basic GET requests were failing with a `400`. Upon inspection of Modsecurity's audit logs, it was apparent there was an issue parsing the request body - the error returned was: `[msg "Failed to parse request body."]`. This was followed with: `[data "XML parsing error: parse error: premature EOF\x0a"]`

The error message was initially confusing as the request contained the header `Content-Type: application/json`, but we confirmed that was just a minor bug with the error message (PR to follow soon).

Assuming there was some unknown content contained within the body of the request, we spent some time evaluating the exact contents of the payload and concluded that no body was being sent at all (as expected).

So it seems ModSecurity validates the body of the request even if there is none to parse. It works correctly with XML but an empty string is considered invalid JSON. I noted the commented out lines above the parsing execution code at: https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/transaction.cc#L656-L663

I do wonder on what basis the decision to parse even when there is no content was made. Not wanting to presume too much, our temporary fix was to move that check into the JSON parsing block only (we could have excluded the GET requests and others explicitly in a chained rule, but if a body is present, we believe it should be parsed).

I am happy to provide another PR but wanted to get the community's thoughts on the above first.

cc: @jayrapson

zimmerle:
Hi @metalspawn,

Regardless of the request body content, there are rules that still needs to be processed during the request body phase. Those may or not may be related to the request body. The key point is this call here:
https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/transaction.cc#L799

Ideally there is no reason to put a rule in the request body phase if it is not directly related to a request body, but, some users thinks otherwise.

zimmerle:
Hi @metalspawn,

One thing that I was testing here was the behavior of ModSecurity version 2 dealing with an empty JSON. The behavior is to treat an empty JSON as an error. Check the logs here:
https://gist.github.com/zimmerle/bef703e5f57c42d5d958fe20a59e8971

I was using:
```
curl -H "Content-Type: application/json" -X POST -d '' http://localhost
```

Ultimately the ModSecurity version 2 and 3 have to behave in the same fashion. At least for the first version of ModSecurity version 3.

[...]

PR #1393, opened 2017-04-20
Fix parsing error message for JSON format

metalspawn:
This is a very small change to bring the request body parsing error message in line with the format being checked.

It was discovered during an investigation into the causes of #1392.


zimmerle:
Merged thanks ;)

PR #1400, opened 2017-04-27
Option to disable logging of log producer (2nd try)

marcstern:
[Issue #1069 - Marc Stern]


marcstern:
Fixed

apache2/msc_logging.c
```diff
@@ -1271,6 +1271,9 @@ void sec_audit_logger_json(modsec_rec *msr) {
         if (been_opened == 1) {
             yajl_gen_map_close(g); // sanitized args map is finished
         }
+#ifdef LOG_NO_PRODUCER
```
zimmerle: This is way more than the simple producer tag.

apache2/msc_logging.c
```diff
@@ -2013,6 +2013,9 @@ void sec_audit_logger_native(modsec_rec *msr) {
             sec_auditlog_write(msr, text, strlen(text));
         }
 
+#ifdef LOG_NO_PRODUCER
```
zimmerle: Same as above - too many things ignored that are not the producer.

marcstern:
The goal was to remove all the "contextual" environment, which is indeed more than the producer. Would it be OK with an explicit name (also in configure) like "LOG_NO_SERVER_CONTEXT"?

zimmerle:
If that is the goal, it should be clear to whom may read the code and to the user who may enabling it during the configure. So we have to change the help message and the if/defs.

marcstern:
Done: {dis|en}able-server-context-logging

PR #1413, opened 2017-05-10
Option to disable logging of notice info at start (versions, ...) in error log.

marcstern:
Also fixed typo in configure.ac in $log_server_context

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD MMM YYYY - 2.9.2 - To be released
 ------------------------------------
 
+ * {dis|en}able-start-notice-logging: Option to disable logging of
+   notice info at start (versions, ...) in error log.
  * Cosmetics: added comments on odd looking code to prevent future
    scrutiny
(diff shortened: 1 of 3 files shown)


zimmerle:
Hi @marcstern,

I am not able to foresee where this specific option can cause a benefit. Can you enlighten me ?

marcstern:
This information is useless when you have a good change control in place and takes a lot of space in the log when you restart the server several times a day

zimmerle:
Hi @marcstern,

Maintain a lot of options in the configuration file is tricker. I would recommend you to pass the log over a pipeline where you can filter the log to save what is interesting to you. It seems to me that this request is particular interesting to your environment. Please explain why a filter via pipe is not interesting to you.

marcstern:
performance on high load

zimmerle:
We do have very high load servers running logs into pipes and they work like a charm. The expensive computation is actually save the data to the disk (or SSD). The filter is negligible. What are you using as a the pipe application?

PR #1414, opened 2017-05-10
Allow $1, $2 ... $9 in @rsub replacement string

marcstern:
PCRE supports \1 and $1 for capture groups in replacement string.
ModSecurity only supports \1.
Although the \1 syntax works most of the time, it is incompatible with mod_macro, as \ are stripped from parameters. To use \1 in a parameter, you need to use \\\\1. It's even worse: if you don't know how many macros are nested (it's the case most of the time), you don't know how many \ to use.
This adds the support for the $1 syntax, that works nicely with mod_macro (or without it).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,7 @@
 DD MMM YYYY - 2.9.2 - To be released
 ------------------------------------
 
+ * Allow $1, $2 ... $9 in @rsub replacement string
  * Cosmetics: added comments on odd looking code to prevent future
    scrutiny
    [Issue #1279 - Coty Sutherland]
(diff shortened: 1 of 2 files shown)


zimmerle:
Hi @marcstern,

I understand the value of this possibility, but maintain those multiple options may be trickier. There are some things that we are willing to test, one of those test includes the option to try other regular expression engines, therefore we want to limit the options to make easy the integration with other tools, that may or may not have those different options. So the confort to not to deal with the multiple slashes may be a problem to integrate other engines, or even to keep the compatibility between version 3 and version 2.x. Therefore I am closing this pull request.

PR #1433, opened 2017-05-26
Add missing hexDecode transformation to seclang parser

victorhora:
This solves solves the `Invalid action: t:hexDecode` error if one tries to use the [hexDecode](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#hexDecode) transformation with libModSecurity. 
The transformation itself was already implemented at [hexdecode.cc](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/actions/transformations/hex_decode.cc#L59-L74) but it was missing some references in the parser.


zimmerle:
Merged ;) thanks!

Issue #1432, opened 2017-05-26
SecConn(Read|Write)StateLimit warning floods errorlog


emphazer:
apache 2.2.15, 2.4.6
rhel 6, 7

i guess it has something to do with the following commit #1340

```
[Fri May 26 12:12:19.354369 2017] [:warn] [pid 13045:tid 140075442624256] [client  x.x.x.251:41630] ModSecurity: going to loop through 2050 servers with 64 threads
[Fri May 26 12:12:19.355458 2017] [:warn] [pid 13045:tid 140075442624256] [client  x.x.x.251:41630] ModSecurity: threads in READ: 1 of 50, WRITE: 0 of 50, IP:  x.x.x.251
[Fri May 26 12:12:19.387705 2017] [:warn] [pid 13129:tid 140075392268032] [client  x.x.x.94:25222] ModSecurity: going to loop through 2050 servers with 64 threads
[Fri May 26 12:12:19.389408 2017] [:warn] [pid 13129:tid 140075392268032] [client  x.x.x.94:25222] ModSecurity: threads in READ: 1 of 50, WRITE: 0 of 50, IP:  x.x.x.94
[Fri May 26 12:12:19.393744 2017] [:warn] [pid 13045:tid 140075392268032] [client  x.x.x.94:25226] ModSecurity: going to loop through 2050 servers with 64 threads
[Fri May 26 12:12:19.394955 2017] [:warn] [pid 13045:tid 140075392268032] [client  x.x.x.94:25226] ModSecurity: threads in READ: 1 of 50, WRITE: 0 of 50, IP:  x.x.x.94
[Fri May 26 12:12:19.400368 2017] [:warn] [pid 13129:tid 140075367089920] [client  x.x.x.251:41636] ModSecurity: going to loop through 2050 servers with 64 threads
[Fri May 26 12:12:19.401456 2017] [:warn] [pid 13129:tid 140075367089920] [client  x.x.x.251:41636] ModSecurity: threads in READ: 1 of 50, WRITE: 0 of 50, IP:  x.x.x.251
[Fri May 26 12:12:19.405759 2017] [:warn] [pid 13045:tid 140075574433536] [client  x.x.x.94:25232] ModSecurity: going to loop through 2050 servers with 64 threads
[...]

PR #1340, opened 2017-03-03
Fix SecConn(Read|Write)StateLimit on Apache 2.4

tjikkun:
Fixes #1337 and #786

PR #1436, opened 2017-05-29
Fixes issue #1432 by not logging normal behaviour to error.log and using APLOG_DEBUG instead

victorhora:
Apparently APLOG_TRACE3 is not supported on older versions of Apache/APR (2.2.x) so AP_LOG_WARNING was used on #1340 but caused logs to flood. This patch suggests using [APLOG_DEBUG](https://ci.apache.org/projects/httpd/trunk/doxygen/group__APACHE__CORE__LOG.html#gadfcef90537539cf2b7d35cfbbbafeb93) which should be safe for all supported versions of Apache/APR.

### DIFF
--- a/apache2/mod_security2.c
+++ b/apache2/mod_security2.c
@@ -1494,7 +1494,7 @@ static int hook_connection_early(conn_rec *conn)
             }
         }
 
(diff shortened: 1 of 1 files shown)


zimmerle:
Merged. Thanks ;)

PR #1446, opened 2017-06-06
added "empy chunk" check

gravagli:
**Environment:**
- Apache 2.2.15 on CentOS 6, 
- Mod_security 2.9.1
- SecRequestBodyAccess On
- ProxyPass

**Scope:**
Apache receives SOAP requests and then proxies them to a target server. In the meantime Mod_security logs the entire request and response sequentially, thanks to SecRequestBodyAccess.

**Issue:**
If the client sends a chunked request big enough to force Mod_security to store the chunks on disk (more than 128 kB by default), the webserver generates two final chunk with zero instead of one.
The destination server behave like it received two requests, but the second one is sintactically uncorrected because it contains only an empty chunk.
This issue happens only when flag _SecRequestBodyAccess_ is on, but I would like keep it enabled.
I tried to upgrade Apache to version 2.4, CentOS to version 7 but nothing changed.

**Possible fix:**
I checked out the code and I found the way to fix the problem:

In **apache2_io.c** at **line 88** one more "if" is needed to prevent Mod_security to add an empty chunk in excess
  ```
  if (chunk->length > 0){
        if (chunk && (!msr->txcfg->stream_inbody_inspection || (msr->txcfg->stream_inbody_inspection && msr->if_stream_changed == 0))) {
            /* Copy the data we received in the chunk */
            .....
        } else if (msr->stream_input_data != NULL) {
           .....
        }
  }
```
I applied this fix in master, v2.9.1 tag and v2/master and it always worked out.

I ran all tests after my fix, and none failed.

### DIFF
--- a/apache2/apache2_io.c
+++ b/apache2/apache2_io.c
@@ -85,10 +85,11 @@ apr_status_t input_filter(ap_filter_t *f, apr_bucket_brigade *bb_out,
         return APR_EGENERAL;
     }
 
-    if (chunk && (!msr->txcfg->stream_inbody_inspection || (msr->txcfg->stream_inbody_inspection && msr->if_stream_changed == 0))) {
-        /* Copy the data we received in the chunk */
-        bucket = apr_bucket_heap_create(chunk->data, chunk->length, NULL,
-                f->r->connection->bucket_alloc);
+    if (chunk->length > 0) {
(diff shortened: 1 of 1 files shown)


zimmerle:
Merged! Thanks ;)

PR #1454, opened 2017-06-12
Few missing headers, in the <arpa/inet.h> inclusions ones

devnexen:
mainly due to the fact APR_HAVE* constants are simply into apr.h
in addition a small memory leak fix.


zimmerle:
Merged! Thanks ;)

PR #1467, opened 2017-06-19
Add Docker Support for ModSecurity

csanders-git:


bostrt:
I'm not sure of your intentions here but an option to copy in a custom mod_security.conf would be helpful for reproducers. Also, an option to install OWASP ModSecurity CRS would be handy too.

csanders-git:
@bostrt CRS will be releasing (and hosting for that matter) this docker image and their own docker image that builds on top of this and adds CRS. This is just a base so that you may use whichever ruleset you want :).  It would be helpful to allow a separate mod_security.conf file, that is true though. This is only a base.

bostrt:
Understood, I think my needs would be better served by extending this base image into my own. Thanks for the work you've put into it!

zimmerle:
I am not sure if that is the right place to have this docker file. For the very same reason that we don't host scripts (or configuration) to build packages. They are complete independent from our release cycle, and every user may want to make custom modifications. Why not to have an independent repo and/or gists to host this docker content? Or even an entry on our wiki to describe it for the docker users?

csanders-git:
Closing this branch as a separate Repository has been added for this.
ModSecurity can now be found on dockerhub https://hub.docker.com/r/owasp/modsecurity/

zimmerle:
Hi @csanders-git,

Great :) I think that is the best place for it to be. I would also recommend it to be renamed to ModSecurity insetead of OWASP/modsecurity.

Issue #1285, opened 2016-12-10
Bad order of files described with a wildard in nginx


thierry-f-78:
Hi,

I probably found a bug with the order of files loaded when the files are described with a wildcard. 
I fixed this. The commit description contains more information.

UnfortunateIy, I do not understand anything with the pull request system, so I put the patche on the git format on my web site here:

[http://www.arpalert.org/0001-Fix-bug-when-load-files.patch](http://www.arpalert.org/0001-Fix-bug-when-load-files.patch)

And a copy/paste here (I have some doubts about this method)

```
From 38d327581e2b7dd75809e19855dc5c110ad0a21f Mon Sep 17 00:00:00 2001
From: Thierry FOURNIER <tfournier@arpalert.org>
Date: Sat, 10 Dec 2016 18:40:01 +0100
Subject: [PATCH] Fix bug when load files

The order of files loaded is very immortant. Typically the "SecRuleRemoveById"
rules must be loaded after the concerned rules. Today, when I try to load
AWASP rules with a wildcard directive, an strace shows the following
order (I remove useles syscall and I short the path):

   strace -e open ....
   [...]
   open("./modsecurity.conf", O_RDONLY|O_CLOEXEC) = 4
   [...]
   open("./rules/modsecurity_crs_90_exceptions.conf", O_RDONLY|O_CLOEXEC) = 5
   open("./rules/modsecurity_crs_47_skip_outbound_checks.conf", O_RDONLY|O_CLOEXEC) = 5
   open("./rules/modsecurity_crs_46_slr_et_wordpress_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   [...]
   open("./rules/modsecurity_crs_46_slr_et_phpbb_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   [...]
   open("./rules/modsecurity_crs_46_slr_et_joomla_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   [...]
   open("./rules/modsecurity_crs_41_xss_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   open("./rules/modsecurity_crs_41_sql_injection_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   open("./rules/modsecurity_crs_40_generic_attacks.conf", O_RDONLY|O_CLOEXEC) = 5
   [...]
[...]

PR #1479, opened 2017-06-29
IIS build refactoring + proposed fix for wildcard op when loading conf files on Nginx / IIS

victorhora:
Proposed fix for #1285 by @thierry-f-78.

Did a quick review of the patch and confirmed working as intended on Nginx and IIS. My version of Apache/APR seems ok with it as well. Let's see what the buildbots and @zimmerle have to say :)


victorhora:
Closing in favor of #1486 and #1487

PR #1486, opened 2017-07-04
Proposed fix for wildcard operator when loading conf files on Nginx / IIS

victorhora:
Proposed fix for #1285 by @thierry-f-78.


zimmerle:
Merged. thanks.

PR #1487, opened 2017-07-04
Refactoring of IIS build scripts

victorhora:
Fixes some hardcoded stuff, updates all dependencies, allows more customization and downloads required dependencies etc. It should make the IIS build a breeze. Hopefully will not break @zimmerle 's buildbots :P


zimmerle:
Merged already as part of v2.9.2 release.

Issue #1482, opened 2017-07-03
@DetectXSS doesn't have a capture option.


csanders-git:
while DetectSQLi has a capture option which puts data into tx.0 DetectXSS doesn't. This behavior is inconsistent and confusing. It should be fixed in order to prevent confusion on behalf of developers. Also @victorhora indicated that he would add elements that support capture to the reference manual.

victorhora:
Proposed fix at: #1488 

Reference manual updated stating the operators which supports "capture" actions: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual/8a14f18461156f31cbb609afb7feac2d63cd4744

PR #1488, opened 2017-07-04
Add capture action to @detectXSS operator

victorhora:
Proposed fix for #1482.

Do notice that the API for @detectXSS in libInjection is still in Alpha stage so there's some missing features when compared with @detectSQLi.

In this case, there's no "[fingerprint](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re_operators.c#L2138)" returned from libInjection API to be fed into a "capture" action to tell us exactly what triggered the match like it's done with [libinjection_sqli](https://github.com/client9/libinjection/blob/master/src/libinjection.h#L42-L50)

So this proposed "capture" action would fill TX.0 with the full analyzed parameter value (var->value) when libinjection_xss returns true.

### DIFF
--- a/apache2/re_operators.c
+++ b/apache2/re_operators.c
@@ -2158,12 +2158,14 @@ static int msre_op_detectSQLi_execute(modsec_rec *msr, msre_rule *rule, msre_var
  */
 static int msre_op_detectXSS_execute(modsec_rec *msr, msre_rule *rule, msre_var *var,
     char **error_msg) {
-
+    int capture;
     int is_xss;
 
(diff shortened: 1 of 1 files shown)


fgsch:
Is this any different to using MATCHED_VAR?

victorhora:
In this context, no @fgsch. But I'm not sure if there's much more we can do as the "API" looks limited in this aspect as of now.

But there's a chance I'm missing something, so let's see what others have to say :)

zimmerle:
Merged. Thanks!

PR #1500, opened 2017-07-13
Ignore built files.

lkarsten:
Clean up git status output so the important parts are visible.


lkarsten:
Hi. Is there anything I can do to move this forward?

zimmerle:
Merged thanks ;)

PR #1502, opened 2017-07-13
Use documentation class IP-addresses in tests.

lkarsten:
This commit changes the client and server IP adresses used in test cases
from 200.249.12.31 (Brazil, unknown origin) to the documentation scope
established in RFC5737.

Related: https://tools.ietf.org/html/rfc5737


lkarsten:
Hi. I've force pushed a new commit to my local branch that removes a test case change that snook into the original PR. Tests all pass on my end.

zimmerle:
Hi @lkarsten,

The propose of those test is to actually work with real (or somewhat similar to real) data. Is that any specific reason why to change the IP address to this one reserved for documentation?

lkarsten:
Hi. Thanks for replying.

Traffic tend always end up leaking to these live adresses in the end.  If not now, then in a couple of years when someone extends the test system to be more complete, and nobody thinks of this.

IPs get reassigned, and suddenly some third party get the unintended traffic. When this is realised 6-12 months after the change (when they start complaining), getting new code out there takes probably a year. Lots of wasted effort by the involved.

Another issue is that it (potentially) opens up to the Brazilian IP tracking who is building/testing libmodsecurity. I'd be surprised if not something already does a reverse DNS lookup of the IP.

Anyway, it is a cosmetic change. It shouldn't matter as I understand the tests, but it looks better and is a safer long term choice. It is a suggestion, feel free to use it if you want. :)

lkarsten:
Hi. We should try to wrap this up.

Who owns the `200.249.12.31` IPv4 address? Do they know it is being used for this?

I can close and forget about this PR if you think it is a bad idea and don't want to proceed. My view is that these IPv4 addresses are set aside for such use, and we should trust the internet elders on this.

zimmerle:
Hi @lkarsten,

The use cases contains real IPs as they meant to mimic a real request. I cannot foresee any problem of using a real IP. There is no content to be leaked as the information on the test cases are all public. With that said, I am going to close this pull request.

PR #1511, opened 2017-07-24
Reverse LUA check order

AlfredoRamos:
`find_lua.m4` should check for newer version of LUA first, this patch only reverses the string that is used to check this library.


AlfredoRamos:
Test failed, however it's not related to this patch.

PR #1514, opened 2017-07-27
Cleaner logging

lkarsten:
Attempt at reducing clutter in the debug log to make it easier to read.

### DIFF
--- a/src/rule.cc
+++ b/src/rule.cc
@@ -186,7 +186,7 @@ bool Rule::evaluateActions(Transaction *trans) {
 
 void Rule::updateMatchedVars(Transaction *trans, std::string key,
     std::string value) {
-    trans->debug(4, "Matched vars updated.");
+    trans->debug(9, "Matched vars updated.");
     trans->m_variableMatchedVar.set(value, trans->m_variableOffset);
     trans->m_variableMatchedVarName.set(key, trans->m_variableOffset);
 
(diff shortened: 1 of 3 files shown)


src/rule.cc
```diff
@@ -644,29 +641,27 @@ void Rule::executeActionsAfterFullMatch(Transaction *trans,
...
-                    "action: " + a->m_name);
-                    a->evaluate(this, trans, ruleMessage);
+                trans->debug(4, "Running (non-disruptive) action: " + a->m_name);
+                a->evaluate(this, trans, ruleMessage);
             }
             continue;
         }
         if (trans->m_rules->m_secRuleEngine == Rules::EnabledRuleEngine) {
-            trans->debug(4, "Running (disruptive) action: " + a->m_name);
+            trans->debug(4, "Running (disruptive)     action: " + a->m_name);
```
zimmerle: Keeping the disruptive vs. non-disruptive in the same col?

zimmerle:
Hi @lkarsten,

Thank you for the patch. As a consequence of this changes, one of our test cases was failing. The **3rd test** of the **action-disruptive.json**. The test was based on debug log match, which was changed by your patch.

I've merged your patch and fixed the test case. Thank you ;)

src/rule.cc
lkarsten: Yes! :)

lkarsten:
Thanks for merging!

PR #1525, opened 2017-08-03
Correctly handle the case when implicit @rx starts with "!" (negation).

asterite3:
When a parser saw a token `NOT` followed by `OPERATOR_RX_CONTENT_ONLY` (implicit `@rx`), it used the following rules (https://github.com/SpiderLabs/ModSecurity/blob/0508395/src/parser/seclang-parser.yy#L848):
```C++
| NOT OPERATOR_RX_CONTENT_ONLY
      {
        OPERATOR_CONTAINER($$, new operators::Rx("!" + utils::string::removeBracketsIfNeeded($2)));
```
This code creates an operator Rx with regular expression starting with `!`, while it should create a negated Rx with regex not containing the initial `!`. The following code can be used to demonstrate it:
```C++

}

const char * req_uri = "/test?param=invalid";

int main() {

    delete modsecTransaction;
    delete rules;
    delete modsec;
    return 0;
}
```
With `test.conf`:
```
SecRule ARGS:param "!^valid$" id:1
```
This code should produce an alert because the rule only allows `param` to have value `valid`. Instead, it will print nothing, because the rule checks `param` against regex `!^valid$`.

This breaks OWASP CRS rules that use negation of implicit `@rx`, for example rule `920170` (which contains `SecRule REQUEST_HEADERS:Content-Length "!^0?$"`), that ensures that value of `Content-Length` header can't be other than zero when request method is `GET` or `POST`. In fact this rule will never fire because no header value can satisfy regex `"!^0?$"` (exclamation mark before beginning of the line).

The proposed fix is to let parser use the Rx constructor that accepts 3 arguments (` Rx(std::string op, std::string param, bool negation)`) and pass `true` as negation value:
```C++
 | NOT OPERATOR_RX_CONTENT_ONLY
      {
        OPERATOR_CONTAINER($$, new operators::Rx("Rx", utils::string::removeBracketsIfNeeded($2), true));
```

### DIFF
--- a/src/parser/seclang-parser.cc
+++ b/src/parser/seclang-parser.cc
@@ -1443,7 +1443,7 @@ namespace yy {
   case 33:
 #line 849 "seclang-parser.yy" // lalr1.cc:859
     {
-        OPERATOR_CONTAINER(yylhs.value.as< std::unique_ptr<Operator> > (), new operators::Rx("!" + utils::string::removeBracketsIfNeeded(yystack_[0].value.as< std::string > ())));
+        OPERATOR_CONTAINER(yylhs.value.as< std::unique_ptr<Operator> > (), new operators::Rx("Rx", utils::string::removeBracketsIfNeeded(yystack_[0].value.as< std::string > ()), true));
         std::string error;
         if (yylhs.value.as< std::unique_ptr<Operator> > ()->init(driver.ref.back(), &error) == false) {
             driver.error(yystack_[2].location, error);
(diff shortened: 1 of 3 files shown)


asterite3:
Closing and reopening to trigger `travis-ci` re-build (previous build failed due to `apt-get` timeout: https://travis-ci.org/SpiderLabs/ModSecurity/builds/260504613?utm_source=github_status&utm_medium=notification).

zimmerle:
Hi @asterite3,

Good catch. Sorry for the delay. Patch merged! thanks.

Issue #1521, opened 2017-08-02
nginx: modsecurity_rules_file on empty file stalls forever


lkarsten:
Hi.

Issue: Trying to add an empty rules file leads to nginx blocking indefinitively during configtest and/or start.
Expected: normal startup of the daemon, file ignored since it contains zero rules.

Reported here since the nginx module is just a thin layer on top of `msc_rules_add()`. I believe (but haven't verified) this issue to be in Driver::parseFile() in libmodsecurity.

Using modsecurity-nginx abbf2c47 (current master) and libmodsecurity on 0508395f8. nginx/1.13.2.

```
root@host:/opt/nginxmodsec/conf# pkill nginx; time /opt/nginxmodsec/sbin/nginx -c /opt/nginxmodsec/conf/null.conf 

real	0m0.004s
user	0m0.004s
sys	0m0.000s
root@host:/opt/nginxmodsec/conf# touch /tmp/empty
root@host:/opt/nginxmodsec/conf# ls -l /tmp/empty
-rw-r--r-- 1 root root 0 Aug  2 17:21 /tmp/empty
root@host:/opt/nginxmodsec/conf# vim null.conf
root@host:/opt/nginxmodsec/conf# cat null.conf
user www-data;
worker_processes auto;
pid /run/nginx.pid;

events {
	worker_connections 768;
}

http {
	modsecurity on;
        # LINE BELOW ADDED SINCE LAST TIME
	modsecurity_rules_file /tmp/empty;

	include /etc/nginx/mime.types;
	default_type application/octet-stream;
}

root@host:/opt/nginxmodsec/conf# pkill nginx; time /opt/nginxmodsec/sbin/nginx -c /opt/nginxmodsec/conf/null.conf 
^C
real	0m39.152s
user	0m0.000s
sys	0m0.000s

root@host:/opt/nginxmodsec/conf# 
```

met3or:
I can confirm this behavior in my installation also, whether you're restarting or reloading the service.

Using both current master versions of libmodsecurity + the connector on CentOS 7 & nginx 1.12.1.

zimmerle:
Fixed.

PR #1530, opened 2017-08-03
Handle zero byte rule files correctly.

lkarsten:
This corrects the stalling behaviour seen when trying to parse an empty
rule file.

Fixes: #1521

### DIFF
--- a/src/parser/driver.cc
+++ b/src/parser/driver.cc
@@ -134,6 +134,10 @@ int Driver::parse(const std::string &f, const std::string &ref) {
         this->ref.push_back(ref);
     }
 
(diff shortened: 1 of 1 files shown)


lkarsten:
The edge case here is 0 bytes. A 1 byte file loads just fine. The 0 byte length means that the look-ahead in the parser blocks indefinitely on getc(). We don't have that kind of time.

I don't see an empty file as an error, just return and move on.

lkarsten:
Hello. Anything I can do to move this along? Different approach needed?

zimmerle:
Sorry for the delay. Merged. thanks!

PR #1536, opened 2017-08-19
Fix #1528: add macro expansion for @rx operator

### LINKED ISSUE #1528: Macro expansion inside regex does not work in ModSecurity V3
Documentation on macro expansion says (https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#macro-expansion): ` You can use macro expansion for operators that are "compiled" such @rx, etc. however you will have some impact in efficiency.`.
In V3, macro expansion in regexes is not performed and operator `"@rx ^%{var}$"` will match data against literal `^%{var}$`.
The following code can be used for demonstration:
```C++

}

const char * req_uri = "/test?param=attack";

int main() {

    delete modsecTransaction;
    delete rules;
    delete modsec;
    return 0;
}
```
With following `test.conf`:
```
SecAction "id:1, nolog, setvar:tx.bad_value=attack"

SecRule ARGS:param "@rx ^%{tx.bad_value}$" id:2
```
It won't produce expected alert because modsec will match `param` against literal `^%{tx.bad_value}$` instead of `^%attack$`.

This breaks OWASP CRS rule `920420` which uses macro expansion of variable `tx.allowed_request_content_type` (this part of the rule will fire for any request because negated operator is used):
```
SecRule TX:0 "!^%{tx.allowed_request_content_type}$"
```

asterite3:
There is a problem that macro expansion in regex of `@rx` operator does not work (see #1528).

This pull request adds a code that tries macro expansion on regex before matching, and, if regex text changed after expansion, create new Regex object from it and use that for matching.

Also, add an optimization: in Rx constructor, check if macro expansion is possible at all for regex (`%{...}` present in it), cache the result of this check (save it in a boolean field of Rx object) and try macro expansion before matching only if check was positive. This should avoid macro expansion attempts + string compares for most regexes (at least, for most regexes in OWASP CRS).

Also, add a regression test.

Fixes #1528

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -128,6 +128,7 @@ TESTS+=test/test-cases/regression/variable-TIME_MIN.json
 TESTS+=test/test-cases/regression/action-setuid.json
 TESTS+=test/test-cases/regression/action-setrsc.json
 TESTS+=test/test-cases/regression/issue-1152.json
+TESTS+=test/test-cases/regression/issue-1528.json
 TESTS+=test/test-cases/regression/config-calling_phases_by_name.json
 TESTS+=test/test-cases/regression/variable-USERID.json
 TESTS+=test/test-cases/regression/request-body-parser-multipart.json
--- a/src/operators/rx.cc
+++ b/src/operators/rx.cc
@@ -39,10 +39,12 @@ bool Rx::evaluate(Transaction *transaction, Rule *rule,
         return true;
     }
 
-    std::string eparam = MacroExpansion::expand(m_param, transaction);
+    if (m_macro_expansion_possible) {
+        std::string eparam = MacroExpansion::expand(m_param, transaction);
 
-    if (eparam != m_param) {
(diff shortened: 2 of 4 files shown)


victorhora:
PR looks great. Will also fix #1552. 

Ready to merge once @zimmerle confirms it's all good with his buildbots :)

defanator:
Hey @zimmerle @victorhora - any reasons this PR is still not merged?

Just wondering due to related issue reported here:
https://github.com/SpiderLabs/ModSecurity/issues/1552

Thanks!

zimmerle:
Hi @defanator,

It is in my queue to review. In particular I am interested to understand if there is memory corruption in 
 a thread environment. It seems to be safe to use in multi-process.

zimmerle:
Hi @asterite3,

Thank you for the patch! It is already merge. Notice however that I have removed the optimization, as it demands further investigation.

Issue #1366, opened 2017-03-26
SecStreamInBodyInspection=On causes slow uploads


tomsommer:
This post is still very very relevant: https://gryzli.info/2015/11/07/waf-modsecurity-slow-upload/
`SecStreamInBodyInspection=On` makes uploads extemely slow.

If this is unavoidable, then at the very least you should put a big fat warning on the usage of `SecStreamInBodyInspection=on`

Since `StreamInBodyInspection` cannot be controlled in `ctl` it's a big problem.

bostrt:
@tomsommer Do you think adding something like below in docs would suffice?

```
Note: This directive may significantly impact file upload times due to added overhead. 
```

https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#secstreaminbodyinspection

tomsommer:
Well, if this is working as intended, then yes.

bostrt:
@tomsommer I tried running some tests myself but wasn't able to notice any performance impact. I do not doubt that there is a performance impact by enabling `SecStreamInBodyInspection`, my test just probably isn't decent.

Do you have a simple test you can share in here that others can see/test with? 

(Would be great if the poster of https://gryzli.info/2015/11/07/waf-modsecurity-slow-upload/ would share their tests showing the 20x slow-down and high CPU usage)

Regarding the docs change, maybe we'll hold off on that until we can validate any slowdown.

tomsommer:
It's been a while so I cannot find a testcase, however I can surely confirm this issue. I see the feature is removed in v3, for good reason I guess.

tomsommer:
Maybe #1316 is related

victorhora:
Yes SecStreamInBodyInspection directive is no longer supported on libModSecurity (aka v3). By default v3 will try to use stream whenever it is possible. I've added a note on the [docs](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual/34cff89be85c79fc52e6dfbd9da1d4404314223e#SecStreamInBodyInspection) to reflect this a couple of weeks ago.

I agree with @bostrt it would be great if the poster of that [blog](https://gryzli.info/2015/11/07/waf-modsecurity-slow-upload/) could share his test cases with us. But anyways I think this might be a won't fix for 2.9.x as it's already fixed by libModSecurity. I'll tag @zimmerle here for him to confirm what's the case for this one.

zimmerle:
[...]

PR #1538, opened 2017-08-23
Preallocate memory when SecStreamInBodyInspection is on

allanbomsft:
This is an improved fix for #1366 .

- Preallocate memory when SecStreamInBodyInspection is on. On my dev box this gave 20x speed improvement for 10mb upload (and even more for larger files).
- Simplified the structure of modsecurity_request_body_to_stream
- Removed null termination for the stream_input_data buffer, as I did not see a reason for having this. The buffer is not a string. A request body can contain binary data.
- Fixed a typo in build_yajl.bat
- Added a check of the return value of modsecurity_request_body_to_stream

Thanks!

### DIFF
--- a/apache2/apache2_io.c
+++ b/apache2/apache2_io.c
@@ -281,7 +281,9 @@ apr_status_t read_request_body(modsec_rec *msr, char **error_msg) {
             }
 
             if (msr->txcfg->stream_inbody_inspection == 1)   {
-                modsecurity_request_body_to_stream(msr, buf, buflen, error_msg);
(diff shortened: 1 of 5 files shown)


zimmerle:
Fixed a typo in build_yajl.bat -> merged! thanks! 

Looking the other stuff.

zimmerle:
Indeed, that seems to be a good optimization. However, it may not be an optimization for certain users where the uploads are very small, therefore making it an optional compilation flag.

Thanks @allanbomsft

PR #1539, opened 2017-08-23
Treat _NAMES variables as collections (#5)

abhijoglekar:
* Treat _NAMES variables as collections

* Fix an issue with the offset of ARGS_NAMES.

* Fix regression tests for the new behaviour.

* Add generated seclang files.


zimmerle:
Merged! :) thanks

PR #1540, opened 2017-08-23
Fixes a bug with an unitialized variable.

abhijoglekar:
new_debug_log was unitialized during an error code path.

Fixed this by explicit initializing it to NULL and fixing the order of
the error labels. They now present the correct (reverse) order of the
goto statements.


zimmerle:
Hi @abhijoglekar,

Merged! thanks ;)

PR #1543, opened 2017-08-23
Remove SecRequestBodyInMemoryLimit from configuration template

defanator:
It is no longer supported since 81879cd, so it would be better to remove any mentions from configuration as well to avoid possible confusions.


victorhora:
Makes sense! Merged. Thanks :)

PR #1571, opened 2017-09-21
Store the connection and url parameters in std::string

majordaw:
Passing stack allocated strings to the Transaction object could lead to memory corruption when the Transaction object outlive the stack. In this case the _m_clientIpAddress_, _m_serverIpAddress_, _m_httpVersion_ and the _m_uri_ pointers will point to invalid addresses.
I bumped into this when I made a transaction cache and one of my thread initialized the transactions (did the request phases) and a other finalized them (response phases, logging, free).


zimmerle:
Merged! thanks ;)

PR #1572, opened 2017-09-21
Eliminate some reorder and sign warnings

majordaw:
During linking with libmodsecurity there are many -Wreorder warnings. This patch try to fix them.


zimmerle:
Merged. thanks ;)

Issue #1576, opened 2017-09-27
Incomplete JSON parsing => security vulnerability


marcstern:
** Content was removed by @zimmerle on 2017-Sep-27

zimmerle:
Dear @marcstern,

Regardless of the content, do you really think that it is a good idea to publish a possible security bypass on GitHub ?

marcstern:
In this case, there is very few information an attacker can use.
In case you have a vulnerability in your application in a JSON request without JSON object, then ModSecurity doesn't protect you. Point There is no way to use that info to tweak an attack to bypass ModSecurity.
This disclosure doesn't introduce an additional risk

marcstern:
I created a (one line) pull request for this: https://github.com/SpiderLabs/ModSecurity/pull/1577

marcstern:
Remark: In my patch, I used an empty ARG name. This looks the most logical approach; however, in a strict environment, you should have a rule forbidding (for non-JSON requests) an empty ARG name.
You obviously have the possibility to fine-tune your rule to not apply it to JSON requests but, if this syntax is totoally usual (but is it really?), then we may use a non-empty name (ex: "unnamed") to not trigger that kind of problem.
Any experience with this syntax (in any direction)?

marcstern:
This patch is trivial and has a high value, why not adding it in the 2.9 trunk?

zimmerle:
I think you meant to make this comment on the related pull request, not in this issue. The pull request can be found at #1577. The pull request is under review, it will be merged once the review is done.

zimmerle:
This is already fixed in v3.0. The solution adopted is different from the ticket #1577. In v3 the json parser understands the array structures into array_n format, where "n" is the position of the element. Also, we moved from string concatenation to a stack structure making the implementation more robust. Fix is still pending for v2.

zimmerle:
Thanks @marcstern.

Merged as part of: 9b6d4b2bb9887e6647be585db1dfb03bd7114925 and 25e5543c7f66c97871d9202cd15b9945de90faf1.

PR #1577, opened 2017-09-28
Include unanmed JSON values in unnamed ARGS

marcstern:
Issue #1576: unanmed JSON values are not parsed because not included in ARGS.
Include these JSON values in unnamed ARGS to include in parsing

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 18 Jul 2017 - 2.9.2
 -------------------
 
+ * Include unanmed JSON values in unnamed ARGS
+   [Issue #1576 - Marc Stern]
  * IIS build refactoring and dependencies update
    [Issue #1487 - @victorhora]
  * Best practice: Initialize msre_var pointers
--- a/apache2/msc_json.c
+++ b/apache2/msc_json.c
@@ -25,8 +25,7 @@ int json_add_argument(modsec_rec *msr, const char *value, unsigned length)
      * to reference this argument; for now we simply ignore these
      */
     if (!msr->json->current_key) {
-        msr_log(msr, 3, "Cannot add scalar value without an associated key");
-        return 1;
+        msr->json->current_key = "";
     }
 
     arg = (msc_arg *) apr_pcalloc(msr->mp, sizeof(msc_arg));


marcstern:
Remark: running in prod for more than a month on several (big) servers

marcstern:
Any chance to accept this PR?

marcstern:
JSON standrd (rfc8259) explicitely states that a JSON body can contain a stand-alone value.
Actually, a lot of libraries and widely used applications are using that on some places.
It's a major issue to not support this.

zimmerle:
This issue was already addressed in v3. To avoid different behaviors in between the version 2 and 3, we need the version 2 to work in the exactly same fashion as v3. There was a comment explain how it works on v3: https://github.com/SpiderLabs/ModSecurity/issues/1576#issuecomment-343899565

It important to keep v2 and v3 to working in the same fashion. Let me know your considerations on how it was implemented on v3.

marcstern:
From a functional point of view, it should be closer than the actual situation, no?
Why not implementing it until a complete rewrite is peformed?

zimmerle:
In branch the v2/dev/issue_1576 there is an alternative where the data structure is somewhat navigable, for instance, the JSON: 
```json
{  
   "foo":"bar",
   "mod":"sec",
   "ops":[  
      [  
         "um",
         "um e meio"
      ],
      "dois",
      "tres",
      {  
         "eins":[  
            "zwei",
            "drei"
         ]
      }
   ],
   "whee":"lhebs"
}
```

Will be mapped to: 
```
ARGS:foo -> bar
ARGS:mod -> sec
ARGS:ops.ops.ops -> um
ARGS:ops.ops.ops -> um e meio
ARGS:ops.ops -> dois
ARGS:ops.ops -> tres
ARGS:ops.ops.eins.eins -> zwei
ARGS:ops.ops.eins.eins -> drei
ARGS:whee -> lhebs
```
This is not comprehensive as the v3 solution, but gives the user an opportunity to navigate in the JSON structure.

Just for the reference, that would be the mapping with the patch suggested here.
```
ARGS:foo -> bar
ARGS:mod -> sec
ARGS:ops -> um
ARGS:ops -> um e meio
ARGS:ops -> dois
ARGS:ops -> tres
ARGS:ops.eins -> zwei
ARGS:ops.eins -> drei
ARGS:whee -> lhebs
```

zimmerle:
Thanks @marcstern.

Merged as part of: 9b6d4b2bb9887e6647be585db1dfb03bd7114925 and 25e5543c7f66c97871d9202cd15b9945de90faf1.

marcstern:
The implemented solution fixes most cases but not all.
Still missed, JSON body with only an int (e.g. 25) or only a string (e.g. "abc").
My pull request solves this.

zimmerle:
Hi @marcstern,

Do you mind to create a new issue with an example to illustrate the scenario?

marcstern:
A JSON request with only a number or a string, nothing else.
Example:
--29000000-B--
Host: test
Content-Length: 2
User-Agent: Mozilla 4.0
Accept: text/html, image/gif, image/jpeg, *; q=.2, */*; q=.2

--29000000-C--
25
--29000000-F--

marcstern:
Any feedback?

marcstern:
Any feedback?

Issue #1559, opened 2017-09-08
modest_audit.log does not log anything, nginx


intelbg:
Hello,
I have compiled nginx 1.12 with libmodsecurity v3, nginx connector and CRS3. Although I can't understand why modest audit log does not log anything. This is the configuration (which is 1:1 with the INSTALL steps).

cat modsec_includes.conf 
include modsecurity.conf
include crs-setup.conf
include rules/REQUEST-900-EXCLUSION-RULES-BEFORE-CRS.conf
include rules/REQUEST-901-INITIALIZATION.conf
include rules/REQUEST-905-COMMON-EXCEPTIONS.conf
include rules/REQUEST-910-IP-REPUTATION.conf
include rules/REQUEST-911-METHOD-ENFORCEMENT.conf
include rules/REQUEST-912-DOS-PROTECTION.conf
include rules/REQUEST-913-SCANNER-DETECTION.conf
include rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf
include rules/REQUEST-921-PROTOCOL-ATTACK.conf
include rules/REQUEST-930-APPLICATION-ATTACK-LFI.conf
include rules/REQUEST-931-APPLICATION-ATTACK-RFI.conf
include rules/REQUEST-932-APPLICATION-ATTACK-RCE.conf
include rules/REQUEST-933-APPLICATION-ATTACK-PHP.conf
include rules/REQUEST-941-APPLICATION-ATTACK-XSS.conf
include rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf
include rules/REQUEST-943-APPLICATION-ATTACK-SESSION-FIXATION.conf
include rules/REQUEST-949-BLOCKING-EVALUATION.conf
include rules/RESPONSE-950-DATA-LEAKAGES.conf
include rules/RESPONSE-951-DATA-LEAKAGES-SQL.conf
include rules/RESPONSE-952-DATA-LEAKAGES-JAVA.conf
include rules/RESPONSE-953-DATA-LEAKAGES-PHP.conf
include rules/RESPONSE-954-DATA-LEAKAGES-IIS.conf
include rules/RESPONSE-959-BLOCKING-EVALUATION.conf
include rules/RESPONSE-980-CORRELATION.conf
include rules/RESPONSE-999-EXCLUSION-RULES-AFTER-CRS.conf

In mod security.conf I have the following configurations about logs:

[...]

Issue #1551, opened 2017-08-31
Nginx libmodsecurity json log format and Send a log without generating a file


bosungs2y:
Hello :)
I want to do HTTPS Morning with Nginx + libmodsecurity + Suricata
However, the project is not proceeding because it is not supported in several parts.
Are you planning to update on the following features? (Q1, Q2)

# Information
```
Nginx(1.11.13)
libModSecurity(ver.3)
```

## Q1 Nginx libmodsecurity json log format
1) SecAuditLogType SecAuditLogType JSON ---> Not support
2) SecAuditLogStorageDir ---> Response Data Not Inserted (Bug)

> {"transaction":{"client_ip":"172.26.137.77","time_stamp":"Thu Aug 31 10:29:25 2017","server_id":"2fa238b0000b474b935dcd3ee0934488d43a2429","client_port":63734,"host_ip":"172.26.137.77","host_port":80,"id":"150414296527.369192","request":{"method":"POST","http_version":1.1,"uri":"/","body":"Document.write(ddd","headers":{"Host":"10.195.26.228","User-Agent":"curl/7.54.0","Accept":"*/*","Content-Length":"18","Content-Type":"application/x-www-form-urlencoded"}},"response":{"http_code":405,"headers":{"Server":"nginx/1.11.13","Date":"Thu, 31 Aug 2017 01:29:25 GMT","Content-Length":"174","Content-Type":"text/html","Connection":"keep-alive"}}}}

## Q2 Nginx libmodsecurity Send a log without generating a file
1) SecAuditLog "|/usr/bin/socat -b 5301072 - UDP-SENDTO:10.195.25.3:8888
```
nginx: [emerg] "modsecurity_rules_file" directive Rules error. File: /usr/local/nginx/conf/modsecurity.conf. Line: 21. Column: 71. Invalid input:  SecAuditLog "|/usr/bin/socat -b 5301072 - UDP-SENDTO:10.195.25.3:8888" in /usr/local/nginx/conf/nginx.conf:40
nginx: [emerg] "modsecurity_rules_file" directive Rules error. File: /usr/local/nginx/conf/modsecurity.conf. Line: 21. Column: 71. Invalid input:  SecAuditLog "|/usr/bin/socat -b 5301072 - UDP-SENDTO:10.195.25.3:8888" in /usr/local/nginx/conf/nginx.conf:40
```
2) Can I create a single file when using SecAuditLogStorageDir? (Client)

[...]

Issue #1547, opened 2017-08-25
concurrent logging files empty


AdamJacobMuller:
Hello,

We are running [ModSecurity@1edd3570e11e9bb2b6d86b249232b24917a4b0ac](https://github.com/SpiderLabs/ModSecurity/commit/1edd3570e11e9bb2b6d86b249232b24917a4b0ac) and [ModSecurity-nginx@abbf2c47f6f3205484a1a9db618e067dce213b89](https://github.com/SpiderLabs/ModSecurity-nginx/commits/abbf2c47f6f3205484a1a9db618e067dce213b89) with nginx 1.13.1.

I'm attaching the rule set + modsecurity configuration as well.

From the `SecAuditLog`:
```
www.cyberprosconsulting.com 207.244.80.239 - - [25/Aug/2017:04:02:37 +0000] "GET / HTTP/1.1" 200 16384 - "Pingdom.com_bot_version_1.4_(http://www.pingdom.com/)" 150363375766.543193 - /var/log/nginx/modsec_audit/33062/12763/20170825/20170825-0402/20170825-040237-150363375766.543193 0 0.000000 md5:d41d8cd98f00b204e9800998ecf8427e
```
The files that appear are all zero bytes, which modsecurity appears to understand/expect as it says its writing 0 bytes and the md5 hash is d41d8... (the md5 hash of an empty string).

[rules.zip](https://github.com/SpiderLabs/ModSecurity/files/1250804/rules.zip)

zimmerle:
Make sure you have YAJL support enabled, otherwise the logs file will be empty.

AdamJacobMuller:
https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#secauditlogformat

Does this option no longer exist at all for ModSecurity 3 with the only option being for json-format logging?

zimmerle:
Hi @AdamJacobMuller This option is available on the latest version. You can choose between JSON and old native format. Also, you can compile libModSecurity to support the JSON. Just install the yajl developer packages and re-compile libModSecurity. The compilation process should detect the developer packages and enable the support by default.

AdamJacobMuller:
hi @zimmerle,

I'm not even seeing the keyword SecAuditLogFormat in the source for v3/master, should I be using another branch?

Issue #1493, opened 2017-07-07
Don't AuditLog Json format logging to one file in libmodsecurity?


truman-test:
To save the json format log in libmodsecurity, i know need to set it as follows: **SecAuditLogType Parallel**
( Parallel configuration is probably the same as setting up Concurrent. )
( https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#secauditlogtype )

However, when I set it to **SecAuditLogType: Parallel**, SecAuditLogStorageDir setting should be set, so can not logging to one file.

So, I set it to **SecAuditLogType: Serial** to logging in one file. 
Serial can not be saved in Json format. Do you plan to support it?

Source Code: https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/audit_log/writer/serial.cc

====

# https://github.com/SpiderLabs/ModSecurity-nginx/issues/47#issuecomment-309748770

zimmerle:
Hi @truman-test,

That is an interesting question. A regular JSON is a self contained document. It demands a clear **start** and a clear **end**. For example:

```
{  
   "modsec":"cool"
}
```
**Example 1.** We have: _{_ and _}_, starting and closing the document.

```
[  
   {  
      "modsec":"cool"
   }
]
```
**Example 2.** We have: _[_ and _]_, starting and closing the document.

So, having a single entry in a single file seems fine as it is written in a single shot. To have all the logs in a single JSON is more complicated, as it may demand to use the same techniques that are used on streaming JSON. It could be, for instance, a simple concatenated JSON, but I am afraid that it may not be what you are looking for.

What is the target application that you are interested to use to read those logs?

truman-test:
Hello @zimmerle , 

I use libmodsecurity to do full logging(Request / Response) on Nginx + HTTPS traffic
( I need a json format full logging single file.)

The logged files are sent to OpenNIDS (ex Suricata, Snort) using tail or socat for security monitoring.


+ HTTPS Traffic Monitoring
Nginx+libmodsecurity (HTTPS Traffic Logging Json format Single File) -> OpenSouceNIDS

zimmerle:
Hi @truman-test,

Can you provide the links for the documentation with the expected JSON type?

truman-test:
Hello @zimmerle ,
I want is not special, same as the existing modsecurity2.9 + json type provided by Apache

link : https://gist.github.com/zimmerle/b22660ab33bd6df444fd

ex) 
[...]

PR #1583, opened 2017-10-06
Enable support for AuditLogFormat (JSON/Native)

victorhora:
Should solve issues: #1493 and #1453. 

Also should help at #1559, #1551, #1547, #1484, #1483 and #1395.


intelbg:
With this branch audit logging doesn't work to me too, neither with serial logging type, neither with concurrent/parallel.

intelbg:
I also added SecAuditLogFormat JSON and logging doesn't work too. I used branch v3/dev/json_auditlog. Can you please tell me how can I trace the problem? I tried to see something with strace, but found nothing.

intelbg:
Sorry for my 3 post, but I found the reason. It's the bug in nginx connector for custom 403 page. https://github.com/SpiderLabs/ModSecurity-nginx/issues/55. If I comment the custom 403 page block in nginx logging works perfectly, but only in Serial mode, not in concurrent mode. Will it me fixed as I see that there isn't solution.

zimmerle:
Merged! Thanks ;)

intelbg:
@zimmerle so should I recompile now again with the master branch to solve the issue or I didn't understand you?

zimmerle:
Hi @intelbg,

@victorhora's pull request was about adding the support for [SecAuditLogFormat](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#SecAuditLogFormat). With this patch it is possible for the user to choose between JSON and Regular/Old format.  Independent of [SecAuditLogType](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#secauditlogtype).

Before @victorhora's patch, the format was set according to the type. 

Notice that this patch may not be related to the issue that you are facing, therefore I am merging this as it adds a new functionality on top of ModSecurity v3. I've tested the patch and it seems to be working as expected.

intelbg:
@zimmerle thank you about your reply! I understand. Did you tested it with concurrent type of logging because with it didn't work for me. I hope that the other problem I reported will be solved too, because they are dependent.

intelbg:
In addition, the problem with concurrent logging was my mistake. It works. Remains the problem with custom 403 page only.

PR #1611, opened 2017-11-06
Fix fallback for non-mapped Unicode char on v2

allanbomsft:
Same as https://github.com/SpiderLabs/ModSecurity/pull/1609 but for ModSec version 2.


zimmerle:
Closing this. Further details at: #1609.

PR #1612, opened 2017-11-06
Windows build, fixed duplicate YAJL dir in script

allanbomsft:
Hi Felipe. I see you recently did this change https://github.com/SpiderLabs/ModSecurity/commit/c917df0f2a8acb0fb31262a5ad222d6d78394fe3 , but I think there might be a problem with the YAJL path. When building via build_modsecurity.bat directly, all is fine, because the path becomes ....\yajl\build\\yajl-2.1.0 . But when building via build_release.bat , build_yajl.bat gets invoked, which sets the YAJL_DIR env variable, and so the path becomes ....\yajl\build\yajl-2.1.0\yajl-2.1.0 and the build fails.

Changing build_modsecurity.bat to let YAJL align with the other dependencies, where only the %DEPENDENCIES_DIR% env variable is used, and the rest is hardcoded.

### DIFF
--- a/iis/build_modsecurity.bat
+++ b/iis/build_modsecurity.bat
@@ -15,21 +15,21 @@ set CURRENT_DIR=%cd%
 cd ..\apache2
 del *.obj *.dll *.lib
 del libinjection\*.obj libinjection\*.dll libinjection\*.lib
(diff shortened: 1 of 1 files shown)


victorhora:
Confirm it fixes the issue with the automated build script. 

We might need to adjust the Windows buildbot builds scripts to take this into consideration.

victorhora:
Merged at https://github.com/SpiderLabs/ModSecurity/commit/e97799c9bc5dbf614e317ac3269ce8e7fd6c0176

Thanks! :)

PR #1613, opened 2017-11-06
Let body parsers observe SecRequestBodyNoFilesLimit

allanbomsft:
Previously, modsecurity_request_body_store would keep feeding the body parsers (JSON/XML/Multipart) even after the SecRequestBodyNoFilesLimit limit was met. This change prevents this. Also, modsecurity_request_body_end now returns an error code when the limit is met, so that a message can be logged for this event.


zimmerle:
Merged! Thanks!

PR #1614, opened 2017-11-07
Disallow 'block' as a default action

michaelgranzow-avi:
Currently, it is possible to specify 'block' as the default disruptive
action for any phase. If there is a rule for a phase with default
action 'block' which in turn uses 'block' as its disruptive action,
i.e., wants to use the default disruptive action, and the rule gets
triggered, libmodsecurity enters an infinite recursion with subsequent
segmentation violation.

Include a regression test for this scenario.

### DIFF
--- a/src/parser/seclang-parser.cc
+++ b/src/parser/seclang-parser.cc
@@ -877,7 +877,7 @@ namespace yy {
 
 
     // User initialization code.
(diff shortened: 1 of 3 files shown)


zimmerle:
Hi @michaelgranzow-avi,

Thank you for the patch. Indeed the segfault is not welcomed. However, we still need to have the block working in the same fashion that we had on v2. Therefore, I have proposed a different fix. Available here: 6842d4bba8fa596f6141f5abcbb7d0b636c76eb0

PR #1624, opened 2017-11-28
Few types change can be safely

dcarlier-afilias:
used afterwards (for apr_sleep for example it is an uint64_t-ish type as
parameter).


victorhora:
Hi @dcarlier-afilias 

Thanks for your contribution!

Did you got compiler warnings about these ones? :)

I'm also thoughtful about changing the type of rule_id. I'm thinking there would be other places in the code where rule_id is used and would also need to change the type from int to long.

ghost:
I did get compiler warnings indeed for those lines.

ghost:
@victorhora What do you think of these now ?

marcstern:
Wouldn't it be better to always use long instead of (sometimes) unsigned long?
Especially as there are some tests <= 0.
Otherwise looks good, except for build/compile where change don't look related to this PR, right?

airween:
@marcstern:

I see this is a pretty old PR, and then there wasn't any CI workflow. If you think this is a useful modification, could you sent this PR again in a new PR? Of course we can mention the original author and give the credit in CHANGELOG.

PR #1636, opened 2017-12-20
Correction remove_by_tag and remove_by_msg

Minasu:
Hello,

When I use remove_by_tag, when the condition is true, return 1 causes all other rules to stop processing.
I have a patch that in my case seems to answer my problem.
Hope it can be useful to you too.

Minasu


zimmerle:
Merged! thanks!

PR #1640, opened 2017-12-26
Fix msc_who_am_i() to return pointer to a valid C string

defanator:
Previously this function was unusable as it returned pointer to some garbage data.


zimmerle:
Merged! Thanks.

PR #1651, opened 2018-01-11
IIS, buffer request body before taking lock

allanbomsft:
This is specific to IIS.
Buffer request body before taking lock, in order to prevent slow clients from keeping the lock, preventing all other traffic from flowing.


allanbomsft:
This one is serious. You should consider it.

zimmerle:
Thanks! Merged!

Issue #1663, opened 2018-01-26
Huge memory consumption and slow nginx configtest/start/restart with ModSecurity v3 and ModSecurity-nginx  


vaLski:
Hellow everybody.

I am trying to make modsecurity 3 + modsecurity-nginx connector working on my setup with ~2000 vhosts and ~3000+ rules.

Compilation configuration etc. are all fine **but** when I try to run nginx **configtest**/**start**/**restart** it it is taking ages (5,8,10 **minutes**) for each of the operations to complete.

Also during the configtest and after the restart nginx is consuming **huge** amount of memory talking GBs here.

The problem(s):

nginx -t is taking ages
nginx restart/reload is taking even more time
nginx is consuming huge amount of memory (talking GBs here)

At first it sounds very similar to https://github.com/SpiderLabs/ModSecurity/issues/1546 but it is not. The the patches mentioned there are already applied in my build.

Also I am not really sure if this is a memory leak bug at all or something that is not very well optimized for use cases like mine.

I am also using apache 2.4.x with modsecurity 2.x with the **very same** rule-set and configtest/start/restart operations are completed in matter of (sub)seconds with the same amount of virtual hosts. 

Here is a quick look at the ltrace of nginx -t which I think might be useful for you. It is showing excessive amount of calls to **msc_rules_merge** which is taking most of the configtest time.

root@server.com:/etc/nginx# ltrace  -f -c nginx -t
nginx: the configuration file /etc/nginx/nginx.conf syntax is ok
nginx: configuration file /etc/nginx/nginx.conf test is successful
"% time     seconds  usecs/call     calls      function
------ ----------- ----------- --------- --------------------
 32.87 1783.733281         157  11349307 strncmp
 28.18 1529.493827  1529493827         1 __libc_start_main
 18.44 1000.688939       23966     41753 msc_rules_merge
  5.63  305.725415         162   1882658 strcmp
  4.99  270.902396         158   1711981 memset
  2.35  127.432618         171    743696 memcpy
  1.77   95.819104         160    598387 malloc

If you need more information please let me know. I will be happy to assist.

The setup:

- number of modsecurity rules: 3355
- number of nginx virtual hosts:  1779
- ModSecurity v3 VERSION - c1cd668acb820824e791ba802396c60900eae9c1
- ModSecurity v3 build options
`
./configure \
    --prefix=/usr/local \
    --enable-shared=yes \
    --with-geoip=no \
    --with-ssdeep=no \
[...]

PR #1667, opened 2018-01-26
headers/modsecurity/rules_properties.h - temporary prevent (O)n issue…

vaLski:
… while parsing modsec rules - this should fix https://github.com/SpiderLabs/ModSecurity/issues/1663

### DIFF
--- a/headers/modsecurity/rules_properties.h
+++ b/headers/modsecurity/rules_properties.h
@@ -430,19 +430,19 @@ class RulesProperties {
             std::vector<modsecurity::Rule *> *rules_from = from+i;
             for (size_t j = 0; j < rules_from->size(); j++) {
                 Rule *rule = rules_from->at(j);
-                for (size_t z = 0; z < rules_to->size(); z++) {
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @vaLski,

Thank you for the patch.

Disable the rule id duplication may not be the right approach here as it is necessary due to other reasons. Also, the integer comparison inside the loop should not be the responsible for the numbers on #1663. It may trigger a problem in a race condition, but the problem is likely to be elsewhere. Therefore, I am closing this issue without merge.

vaLski:
@zimmerle I totally agree that disabling this code is not the best approach to solve this issue. Hence my C skills are limited this was the easiest approach for me to try.

Despite that fact, I am sure that this is the code that is triggering the issue I described in #1663 and more specifically the problem with the slow syntax check (nginx -t) and slow nginx start/reload.

Since I applied this patch (code comment) I am not experiencing this issue. Try this with 3k mod security rules and 1000+ vhosts and you will see that too. I bet a beer this is the code that is causing the slowdown during start/restart/reload/configtest.

PR #1691, opened 2018-02-27
Fix LMDB compile error

airween:
I run this problem: https://github.com/SpiderLabs/ModSecurity/issues/1586.
This patch fixed it.


zimmerle:
Merged! Thanks!

PR #1694, opened 2018-03-04
Add missing verify*** operator statements to parser

victorhora:
Add missing references in the parser.

Solves the ``Operator: verifyCPF is not yet supported.`` and ``Operator: verifySSN is not yet supported.`` errors if one tries to use [verifyCPF](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#verifyCPF) or [verifySSN](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual#verifySSN) with libModSecurity.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.0.x - YYYY-MMM-DD (To be released)
 -------------------------------------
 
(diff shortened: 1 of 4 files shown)


zimmerle:
Merged. thanks ;)

PR #1696, opened 2018-03-06
Prettier error messages for unsupported configurations (UX)

victorhora:
Rationale is being explicit on which configurations are currently supported and allowing a smoother transition between v2 to v3, mainly for older or highly modified setups that uses unusual configuration directives.
Also, some unsupported directives that uses CONFIG_VALUE (e.g. SecConnEngine) now allows the user to keep the configuration directive as long as its set to 'Off'.


src/parser/seclang-parser.yy
```diff
@@ -1521,13 +1656,52 @@ expression:
...
+        YYERROR;
+      }
     | CONGIG_DIR_SEC_STATUS_ENGINE
+/* Parser error disabled to avoid breaking default installations with modsecurity.conf-recommended
+        driver.error(@0, "SecStatusEngine is not yet supported.");
+        YYERROR;
+*/
     | CONFIG_DIR_UNICODE_MAP_FILE
+/* Parser error disabled to avoid breaking default installations with modsecurity.conf-recommended
+        driver.error(@0, "SecUnicodeMapFile is not yet supported. utils::string::x2c");
```
zimmerle: utils::string::x2c ? kkk

victorhora: Oops! :P

zimmerle:
merged. thanks.

PR #1701, opened 2018-03-09
Temporarily accept invalid MULTIPART_SEMICOLON_MISSING operator

victorhora:
MULTIPART_SEMICOLON_MISSING is now accepted by the parser and referred to  MultipartMissingSemicolon. Ex: ```SecRule MULTIPART_SEMICOLON_MISSING "!@eq 0" "id:'200011',phase:2,msg:'MULTIPART_SEMICOLON_MISSING'"``` results in:

```Matched "Operator `Eq' with parameter `0' against variable `MULTIPART_MISSING_SEMICOLON' (Value: `1' ) [id "200011"] [rev ""] [msg "MULTIPART_SEMICOLON_MISSING"]```

Suggested temporary addition to the parser related with https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/995, https://github.com/SpiderLabs/owasp-modsecurity-crs/pull/1023 and to avoid issues like https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1032 and https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1021.

### CODE REVIEW

**Comment by defanator** [comment]
@victorhora @zimmerle wow. Are you guys going to merge this one before next libmodsecurity release?

On a related note (not sure if you have seen this in slack) - currently libmodsecurity fails to load default crs-setup.conf with the following error:

```
test@vagrant:~/ModSecurity$ sudo nginx -t
nginx: [emerg] "modsecurity_rules_file" directive Rules error. File: /etc/nginx/modsec/owasp-crs/crs-setup.conf. Line: 96. Column: 43. SecCollectionTimeout is not yet supported.  in /etc/nginx/nginx.conf:75
nginx: configuration file /etc/nginx/nginx.conf test failed
```

This is not [directly] related to CRS, but still worth to fix somehow (comment that directive in crs-setup.conf.default?)

victorhora:
hey @defanator, merging this one is my idea to avoid those issues happening too often. I'm not sure if @zimmerle likes this idea too much as it's a dumb/fake variable that gets pointed to the right one in the end... but I don't see an alternative as of now as it might not be backported on CRS 3.0.x

About the issue with the unsupported directive, I've suggested a change at [v3/dev/fix-seccol_timeout_err_crs](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/fix-seccol_timeout_err_crs) branch. It will make the error go away until the feature is addressed.

zimmerle:
Well. I would fix it on OWASP CRS. ModSec v3 points to the error, while v2 silent fails. IMHO the fact that no one is noticing is not a reason to not fix. Ultimately it is a missing check.

Anyhow, this pull request seems to bring a benefit to the less experienced user which is always good. It was indeed a good idea. Merged. Thank you @victorhora.

Yes @defanator, this will be part of our next release.

PR #1702, opened 2018-03-09
Adds capture action to detectXSS

victorhora:
Adds capture action in the same basis of https://github.com/SpiderLabs/ModSecurity/pull/1488 to mimic the same behaviour as v2


zimmerle:
Merged. Thanks :)

PR #1737, opened 2018-04-10
Define m_secmarker_skipped as an integer type

p0pr0ck5:
There's no reason to treat this this as a double, since it represents a human-readable data value that is only meaningful as an integer. In doing so we write cleaner audit logs and save a small amount of space.

### DIFF
--- a/headers/modsecurity/rules.h
+++ b/headers/modsecurity/rules.h
@@ -80,7 +80,7 @@ class Rules : public RulesProperties {
 
  private:
     int m_referenceCount;
-    double m_secmarker_skipped;
+    uint8_t m_secmarker_skipped;
 };
 
 #endif
--- a/test/test-cases/regression/secmarker.json
+++ b/test/test-cases/regression/secmarker.json
@@ -110,7 +110,7 @@
     },
     "expected": {
       "audit_log": "",
-      "debug_log": "Out of a SecMarker after skip 6.000000 rules.",
+      "debug_log": "Out of a SecMarker after skip 6 rules.",
       "error_log": ""
     },
     "rules": [


zimmerle:
Hi @p0pr0ck5,

Thanks for the patch. Indeed no need do show the decimal part to the user. However, I am afraid this patch will not write a cleaner audit logs or save any considerable amount of space as this specific variable is only used in the debug logs.

As a matter of fact, as of 6f92c8914a822f440b2280802b408a1c2e4c66c2 this variable is complete disable if ModSecurity is compiled without the debug logs. 

#### m_secmarker_skipped Life Cycle
- declared here:
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/headers/modsecurity/rules.h#L83

- initialized here:
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/headers/modsecurity/rules.h#L51
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/headers/modsecurity/rules.h#L57

- incremented here:
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/src/rules.cc#L185

- displayed here:
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/src/rules.cc#L188-L189

- reset here:
https://github.com/SpiderLabs/ModSecurity/blob/a1b6cceb65978c76271e22d41d4a380b381f7caa/src/rules.cc#L191

As you can see, all those references are related to the debug logs. Let me know if you have any further question.

p0pr0ck5:
Whups, I had meant to say debug logs, not audit logs. My apologies.

PR #1738, opened 2018-04-11
Fix STATUS var parsing and accept STATUS_LINE var for v2 backward compatibility

victorhora:
STATUS variable seemed to expecting an invalid parameter. 

[STATUS_LINE](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#STATUS_LINE) is now accepted by the parser to allow backwards compatibility with users migrating from v2 to v3. 

The variable is referring to the implementation of ::Status (m_variableResponseStatus) to avoid having the same data in memory.


zimmerle:
Merged! thanks!

PR #1740, opened 2018-04-12
Execute all tests with 'make check'

p0pr0ck5:
Replace a hard-written list of test files in Makefile.am (that is missing a number of tests) with a list generated via bootstrap build.sh.


p0pr0ck5:
Travis will fail until #1739 is merged (because of the broken test), so these two PRs likely need to be manually merged together.

build/append-tests.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+
+printf "TESTS=" >> Makefile.am
+
+T=$(find test/test-cases -iname *.json)
```
zimmerle, line 5: This will put everything in a single line. IMHO very confuse to read. 

There is a script commented on the Makefile.am that does the same thing but in an elegant fashion:
https://github.com/SpiderLabs/ModSecurity/blob/v3/master/Makefile.am#L87

Output for this patch:
https://gist.github.com/zimmerle/b8be2a60fbf4efe0670fc27beb5ab69d

Output of original script:
https://gist.github.com/zimmerle/051978cb61950d244bbd6edfa3bdf397

Do you mind to share with us, why do you think that aggregates everything in the same line will be better?

zimmerle:
General comment: this is under review waiting for the input of @p0pr0ck5.

p0pr0ck5:
Thanks @zimmerle, let me clean this up following the merge of #1739.

p0pr0ck5:
@zimmerle this is rebased and ready for review.

zimmerle:
Did you addressed my comment in the review session ?

p0pr0ck5:
@zimmerle I don't understand the question? We didn't review this together, AFAIK?

zimmerle:
@p0pr0ck5 It seems like the review was on draft. Updated now.

build/append-tests.sh
p0pr0ck5: I don't care one way or another if the test names are on a single line or multiple lines. IMO if it's an autogenerated list it doesn't matter.

Putting the Bash into the `Makefile.am`, as it is in https://github.com/SpiderLabs/ModSecurity/blob/v3/master/Makefile.am#L87, doesn't work for obvious reasons. So I decided to put it into `build.sh`. Let me know if you have an idea for a better place. I am not overly familiar with automake tools in this regard.

This PR is really only about getting all the appropriate tests to run when `make test` and Travis are executed, so that authors don't have to manually add file names to `Makefile.am`. I have no preference for how things are formatted, only that it's done. Let me know your thoughts on how best to move forward so we see all tests being executed consistently.

zimmerle: That affects the ability of the users to manipulate the generated-distributed files. I understand that you made the patch targeting a very specific point, but it may lead to a negative impact on others. In order to have the code in the mainline we need to make sure that it is not causing a negative impact to any user.

p0pr0ck5: How does this impact other users? Is your point that you want each test definition on a separate line? That's completely fine with me, as I already mentioned. The only goal here is to ensure that the tests that were written are actually running. There are already a fair number of tests that are not being executed by Travis/`make check`.

p0pr0ck5:
@zimmerle what can we do to help push this along? It would be great to have consistent test coverage as part of the Makefile and Travis CI integrations.

zimmerle:
Hi @p0pr0ck5,

I am failing to find the changes in your pull request. Did you published that somewhere else? Are you still interested to help on that?

p0pr0ck5:
@zimmerle I'm still confused about what specifically you want to see in this PR/you object to. As I noted, Putting the Bash into the Makefile.am, as it is in https://github.com/SpiderLabs/ModSecurity/blob/v3/master/Makefile.am#L87, doesn't work because it generates an invalid Makefile. Hence the reason for my approach.

What changes do you want to see in this PR? I will make them so we can merge this in and have consistent test coverage. Thanks :)

zimmerle:
As of 1527f4e2f23c486d560b534d7c82b020bfd342bb I can confirm that 100% of the tests are running. If you see any inconsistency please point us towards it.

p0pr0ck5:
That doesn't help when adding _new_ test suites, as the author might forget to add them (as has clearly been done many times resulting in the need for `1527f4e`).

The idea is to avoid forcing the author to remember to do a manual step when it could be handled automatically for them. It seems like you're averse to this though, so I won't push on this farther. Thanks.

zimmerle:
I see the value of this as a feature. But, part of my responsibility reviewing those pull requests is to make sure that your contribution fits to 100% of the users in the best way that we can. No point to have a feature that is not helping. Your patch is putting 220 files in a single line (about 14276 characters) making it very hard to be read. A file that is likely to be manipulated manually during the development, also probably to be manipulated by packagers.

I think this feature is important, I just think that there is a better way to implement it by putting a file per line. Apparently you agreed. I've just asked for the code, which I failed to find.

Just to give an example of why it is bad to have all those files in a single line, the file `test/test-cases/data/GeoIP2-City-Test-source.json` is listed as a test case and it is not. Not easy to tell that it is there.

PR #1749, opened 2018-04-20
Update modsecurity.conf-recommended

1Jesper1:
Add application/json as default to SecResponseBodyMimeType config for enabling JSON response parsing as default.

### DIFF
--- a/modsecurity.conf-recommended
+++ b/modsecurity.conf-recommended
@@ -115,7 +115,7 @@ SecResponseBodyAccess On
 # configuration below to catch documents but avoid static files
 # (e.g., images and archives).
 #
-SecResponseBodyMimeType text/plain text/html text/xml
+SecResponseBodyMimeType text/plain text/html text/xml application/json
 
 # Buffer response bodies of up to 512 KB in length.
 SecResponseBodyLimit 524288


victorhora (commented):
Thanks for your contribution @1Jesper1.

But at this moment we are not sure if this is the best approach for the default configuration of ModSecurity.

Usually people are more concerned with request body parsing for malicious data. Response body inspection is normally targeted to find sensitive data (e.g. credit card numbers). 

In addition, Intentionally or unintentionally malformed bodies on response body inspection can harm user experience and also cause performance impacts that not everyone can account for.

But maybe we should add a note to the reference manual to make it clear for users that wish to perform JSON body inspection on response need to add _application/json_ to SecResponseBodyMimeType.

What you think?

1Jesper1:
Thanks for the reply, I understand your point. A note in the reference manual about this would be useful.

victorhora:
Hi @1Jesper1,

The note has been added to the reference manual. Your contribution has been referenced on the revision history here: https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)/_history

Thanks!

PR #1750, opened 2018-04-23
Fix memory leak in modsecurity::utils::expandEnv()

defanator:
Found by ASAN.


src/utils/system.cc
```diff
@@ -126,9 +126,9 @@ std::list<std::string> expandEnv(const std::string& var, int flags) {
                 std::ifstream *iss = new std::ifstream(exp[0], std::ios::in);
                 if (iss->is_open()) {
                     iss->close();
-                    delete iss;
                     vars.push_back(exp[0]);
                 }
+                delete iss;
```
zimmerle, line 131: indeed.

zimmerle:
Merged. Thanks!

PR #1755, opened 2018-04-30
Initialize invalid_count in UrlDecode::evaluate

p0pr0ck5:
The other caller of urldecode_nonstrict_inplace initialized the third param, but UrlDecode::evaluate did not.


zimmerle:
Do you mind to describe what is the point to initialize it?

p0pr0ck5:
Good practice, since the value is unused in this context it doesn't have an impact on control flow. It's largely a quality issue (and keeping its usage in line with the other caller is good for code review purposes IMO).

zimmerle:
Same as #1756

PR #1756, opened 2018-04-30
Fix memory leak when an allocation fails

p0pr0ck5:


zimmerle:
Given the utilization context of this function, do you mind to describe the leak that you mentioned that was fixed?

p0pr0ck5:
This is largely a code quality PR. Static analysis tools will find this easily; I was studying the msc_tree functionality when I noticed this and thought it worth a patch. It's not a high priority issue IMO, just a code cleanup.

zimmerle:
Thank you for the clarification. It is very important to have an accurate description for the pull request. A not so expert user may believe that this is going to solve an issue, but, in a real world scenario, this is not an issue. From the user perspective it wont change anything.

PR #1758, opened 2018-04-30
Fix LDFLAGS for unit tests.

smlx:
Fix a typo that was breaking the unit test build.


smlx:
Specifically: it was breaking the build in Alpine Linux.

zimmerle:
Merged! Thanks!

PR #1759, opened 2018-04-30
Fix setvar parsing of quoted data

victorhora:
Proposed fix for an issue with the [setvar](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#setvar) action that although was accepting quoted data, it didn't seemed to be properly handling it.

Hopefully it should solve https://github.com/SpiderLabs/ModSecurity/issues/1733 :)


zimmerle:
Merged! Thanks!

PR #1760, opened 2018-05-01
Build in parallel.

smlx:
This getconf parameter is supported on both Linux and OSX.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,7 +44,7 @@ before_script:
 script:
   - ./build.sh
   - ./configure $OPTS
-  - make
+  - make -j$(getconf _NPROCESSORS_ONLN)
   - make check
   - make check-static
 


zimmerle:
Hi @smlx,

Thanks for the patch! It seems like it is causing a random compilation issue when the parser generation is enabled. See here: https://travis-ci.org/SpiderLabs/ModSecurity/jobs/373323135 (line 1180). 

Most likely a problem is within the build scripts. Holding the merged till we find what is causing the issue.

victorhora:
Indeed. I've triggered another build from Travis to see if we get different results :)

zimmerle:
Forced the parser generation to happens prior to general make seems to temporary fix it. Not able to reproduced it outside Travis.

https://github.com/SpiderLabs/ModSecurity/blob/a278bb673afaaf7a9ba70c2bd1742db02c5c1bf7/.travis.yml#L47

Thanks @smlx!

PR #1761, opened 2018-05-01
Clean up MD5 hexdigest

p0pr0ck5:
The null terminator is not necessary when using this form of the std::string constructor, and its use was confusing given the extra indent.


zimmerle:
Merged. Thanks!

PR #1765, opened 2018-05-03
Fix memory leak in msc_rules_* C APIs

weliu:
When msc_rules_* APIs fail, the error string will return to indicate the reason. It was a duplicate of an internal string. However, it will be duplicated again after those calls in Nginx connector. For example:

In ngx_http_modsecurity_merge_loc_conf():

rules = msc_rules_merge(c->rules_set, p->rules_set, &error);
if (rules < 0) {
    return strdup(error);
}

Since the type of error is 'const char **', I believe that we should not duplicate it inside those APIs and let the caller decide how to handle it.

### DIFF
--- a/src/rules.cc
+++ b/src/rules.cc
@@ -321,7 +321,7 @@ extern "C" int msc_rules_merge(Rules *rules_dst,
     Rules *rules_from, const char **error) {
     int ret = rules_dst->merge(rules_from);
     if (ret < 0) {
-        *error = strdup(rules_dst->getParserError().c_str());
+        *error = rules_dst->getParserError().c_str();
     }
     return ret;
 }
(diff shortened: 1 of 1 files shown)


p0pr0ck5:
Perhaps we should remove the duplicate `strdup` from the Nginx connector instead? The Apache connector does not call such a duplicate: https://github.com/SpiderLabs/ModSecurity-apache/blob/4e8854c46bd3c216d7834ab94605f45114794791/src/msc_config.c#L169-L183.

weliu:
@p0pr0ck5 Since the Apache connector does not free the returned error string, there still exists memory leak if we only remove duplicate strdup from the Nginx connector.

zimmerle:
Merged! Thanks!

weliu:
@zimmerle Yes, I made a mistake, since the getParserError() return a temporary string, it is not safe for us to return its c_str. Sorry for the trouble I have made.

zimmerle:
No problem. I dint notice at first sight as well.

PR #1769, opened 2018-05-08
Use rvalue reference in ModSecurity::serverLog to avoid string copy

weliu:


zimmerle:
Hi @weliu,

Thank you for the patch. Merged!

PR #1781, opened 2018-05-23
set overrideModeDefault to Allow

### LINKED ISSUE #787: Change default IIS setting for overrideModeDefault in applicationHost.config
Hello,

In the 2.8.0 version of the ModSecurity for IIS installer, it adds a line like the one below to the applicationHost.config file (which governs the global IIS configuration for that host--all websites on the host):

`<section name="ModSecurity" overrideModeDefault="Deny" allowDefinition="Everywhere" /></sectionGroup>`

The effect of this configuration is preventing individual websites running on the host from being able to disable ModSecurity (at the site level). I propose changing the IIS Installer's default overrideModeDefault to "Allow" which would allow individual websites the ability to include a ModSecurity enabled="FALSE", thereby disabling ModSecurity for that site. 

Thank you,
Brian Clark

default-kramer:
So that individual websites can add <ModSecurity ...> to their web.config file. Fixes #787


default-kramer:
I did manage to test this locally, and it works as expected.

zimmerle:
Merged! Thanks!

PR #1786, opened 2018-05-25
Fix race condition in UniqueId::uniqueId()

weliu:
There exists a race condition in UniqueId::uniqueId(). I called msc_init() simultaneously in many threads, so they can have their own instance of ModSecurity. However, the process get core dumped every once in a while. The stack looks like this:

This PR uses std::once_flag to make sure UniqueId::getInstance().fillUniqueId() only called once.

### DIFF
--- a/src/unique_id.cc
+++ b/src/unique_id.cc
@@ -55,6 +55,8 @@
 
 namespace modsecurity {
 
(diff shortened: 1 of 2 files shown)


zimmerle:
Merged! Thanks!

PR #1790, opened 2018-06-01
Add basic pkg-config info

gquintard:


zimmerle:
Hi @gquintard,

Thank you for the patch! I was missing the patch level information on the version. I've made a little modification to display it. Everything is now merged!

Here:
https://github.com/SpiderLabs/ModSecurity/blob/a3980bbea16d0a68fbb01fb2cf9320b48ed1a7f8/modsecurity.pc.in#L8

And here:
https://github.com/SpiderLabs/ModSecurity/blob/a3980bbea16d0a68fbb01fb2cf9320b48ed1a7f8/configure.ac#L25-L26
https://github.com/SpiderLabs/ModSecurity/blob/a3980bbea16d0a68fbb01fb2cf9320b48ed1a7f8/configure.ac#L207-L208

gquintard:
Thanks!

gquintard:
Revisiting that one, shouldn't

```
Cflags: -I@includedir@/@PACKAGE@
```

just be 

```
Cflags: -I@includedir@/
```
?

in the first case you include ``modsecurity/modsecurity.h``, in the second one ``modsecurity.h``. The former seems more right to me.

EDIT: it needs to be ``Cflags: -I@includedir@/`` because MS headers include  ``modsecurity/XXX.h``. I'll open a PR

PR #1792, opened 2018-06-04
Added test cases to multipart parser regression tests

airween:
Added more regression test cases to multipart parser.


airween:
Note, that the Travis CI has failed, because the ModSecurity processes wrongly the new request-body test cases.

PR #1801, opened 2018-06-13
Added some documentation for multipart boundary check

airween:


p0pr0ck5:
It might be helpful to also document this somewhere a bit more consumable, like the wiki? But, this is awesome to have in here either way :D

airween:
Well, this patch affects only the 3.x tree, but looks like the complete 3.x documentation is missing. :)

zimmerle:
Merged! Thanks!

PR #1806, opened 2018-06-18
Remove unsupported configuration parameter

Richardds:
nginx: [emerg] "modsecurity_rules_file" directive Rules error. File: /etc/nginx/modsec/modsecurity.conf. Line: 37. Column: 33. As of ModSecurity version 3.0, SecRequestBodyInMemoryLimit is no longer supported. Instead, you can use your web server configurations to control those values. ModSecurity will follow the web server decision. in /etc/nginx/sites-enabled/default.conf:3

### DIFF
--- a/modsecurity.conf-recommended
+++ b/modsecurity.conf-recommended
@@ -38,12 +38,6 @@ SecRule REQUEST_HEADERS:Content-Type "application/json" \
 SecRequestBodyLimit 13107200
 SecRequestBodyNoFilesLimit 131072
 
-# Store up to 128 KB of request body data in memory. When the multipart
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @Richardds,

Thank you for your patch. The configuration in question is no longer valid for version 3, which is hosted here:

https://github.com/SpiderLabs/ModSecurity/blob/4e3a1f71535bd4c7d218754562a4fe53287b8689/modsecurity.conf-recommended#L1

The one that you have patched is the configuration for version 2 in which this entry still valid. Therefore, I am closing this pull request.

PR #1807, opened 2018-06-19
Adds support for ctl:requestBodyProcessor=URLENCODED

### LINKED ISSUE #1797: ModSecurity 3.0 fails with ctl:requestBodyProcessor=URLENCODED
Enforcing URLENCODED body processor results in startup failure:

`Expecting an action, got:  ctl:requestBodyProcessor=URLENCODED"`

Enforcing JSON body processor works alright.

CRS 3.1 is going to have a rule that enforces the URLENCODED body processor when there is a body, but no content-type. Eventually, we would like to have this rule in the recommended-rules, but for the time being, we plan to introduce it into CRS 3.1 as a bypass prevention.

With this bug / missing feature, modsec 3.0 fails to load the upcoming CRS 3.1.

victorhora:
Adds support for URLENCODED body processor as defined at [WWWFormUrlEncoded](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/transaction.cc#L560-564). Should fix #1797


zimmerle:
https://gist.github.com/zimmerle/2e047fb9f60bce10a5596d1d2995bac4

src/actions/ctl/request_body_processor_urlencoded.cc
```diff
@@ -27,7 +27,7 @@ namespace ctl {
 
 bool RequestBodyProcessorURLENCODED::evaluate(Rule *rule,
     Transaction *transaction) {
-    transaction->m_requestBodyProcessor = Transaction::WWWFormUrlEncoded;
+    transaction->m_requestBodyType = Transaction::WWWFormUrlEncoded;
```
victorhora: Hey @zimmerle, although the configuration was being accepted by the parser, there was a problem with the pull request as the [condition to actually extract the arguments](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/transaction.cc#L745-L747) was not being satisfied due to the usage of the m_requestBodyType as opposed to m_requestBodyProcessor variable which is used for the XML and JSON body processors. My bad 😋

victorhora:
Merged as of https://github.com/SpiderLabs/ModSecurity/commit/f999f54eda0674fe3f453bf939b9abcd2aa0290a and https://github.com/SpiderLabs/ModSecurity/commit/aa158ceef3fe88e52679a70cda45adc08667991c

PR #1810, opened 2018-06-20
Fix variables case sensitiveness

### LINKED ISSUE #1808: setvar tx.paranoia_level on v3/master ignored
Referencing here an issue open on owasp-modsecurity-crs that it seems related to libmodsecurity v3/master:  https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1127

victorhora:
Changes the behaviour of the collection backend to fix an issue where variables were case sensitive in some scenarios. Should fix #1808.

### DIFF
--- a/src/collection/backend/in_memory-per_process.cc
+++ b/src/collection/backend/in_memory-per_process.cc
@@ -17,7 +17,7 @@
 #include "src/collection/backend/in_memory-per_process.h"
 
 #ifdef __cplusplus
-#include <string>
+#include <string.h>
 #include <iostream>
 #include <unordered_map>
 #include <list>
@@ -106,7 +106,7 @@ void InMemoryPerProcess::resolveMultiMatches(const std::string& var,
         }
     } else {
         for (auto &a : *this) {
-            if (a.first.compare(0, var.size(), var) == 0) {
+            if (strncasecmp(a.first.c_str(), var.c_str(), a.first.size()) == 0) {
                 l->insert(l->begin(), new VariableValue(&m_name, &var, &a.second));
             }
         }
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.0.3 - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+ - Fix variables case sensitiveness
+   [Issue #1808 - @victorhora, @theMiddleBlue, @michaelgranzow-avi]
  - Allow empty strings to be evaluated by regex::searchAll
    [Issue #1799, #1785 - @victorhora, @XuanHuyDuong, @zimmerle]
  - Adds basic pkg-config info


theMiddleBlue:
Tested with `setvar:tx.paranoia_level=1`, it works! Thanks @victorhora !

src/collection/backend/in_memory-per_process.cc
```diff
@@ -106,7 +106,7 @@ void InMemoryPerProcess::resolveMultiMatches(const std::string& var,
         }
     } else {
         for (auto &a : *this) {
-            if (a.first.compare(0, var.size(), var) == 0) {
+            if (strncasecmp(a.first.c_str(), var.c_str(), a.first.size()) == 0) {
```
michaelgranzow-avi, line 109: Ignoring the change from case-sensitive to case-insensitive (which is the correct fix):

The old code compares the first var.size() characters, the new code compares the first a.first.size() characters. If these two sizes are not the same, this is not equivalent.

Consider var = "tx.paranoia", key = "tx.paranoia_level": the old comparison is true, the new one false.

I don't know why the logic was changed from full comparison to partial comparison in the first place (it looks wrong to me), but with this commit we're changing the logic again.

It would be good to learn about the motivation of the original change (https://github.com/SpiderLabs/ModSecurity/commit/892beb53603a44a3b7cdcbfe83eff9e8c51c9307). One aspect is that it turned a hash lookup into a linear search, so it may well have performance impact.

airween: *Consider var = "tx.paranoia", key = "tx.paranoia_level": the old comparison is true, the new one false.*

Sorry for the question, but why is the old way accepted? In my opinion, any str**N**cmp is incomprehensible. If I want to set a variable with name tx.paranoia, I **want** to search it exactly, not the ones which starts with it. :)

I mean, the old any new method is wrong :)

michaelgranzow-avi: @airween : I agree with you, as I said in my comment. For the purpose of this PR, however, I was just pointing out the change in logic that is not related to case sensitivity, which is the only thing this commit should change.

I've raised the question of comparing only the first N bytes on the original PR.

airween: @michaelgranzow-avi : sorry, may be you misunderstood me. I don't understand why is there any str**n**... comparing, instead of "simple" strcmp()/strcasecmp()? Why just need to compare the first N byte, why not the two strings as they are, without any length argument?

michaelgranzow-avi: @airween, I know what you mean and I agree with you. I've opened https://github.com/SpiderLabs/ModSecurity/issues/1818 and added a patch.

michaelgranzow-avi: Proposed fix: https://github.com/SpiderLabs/ModSecurity/pull/1820 - this would make the current PR https://github.com/SpiderLabs/ModSecurity/pull/1810 unnecessary.

zimmerle:
fixed by: d810de9166794be2bee972fffb0e09cb0b8d6ec5

Issue #1809, opened 2018-06-20
v3/master Lua unknown error


theMiddleBlue:
Hi,

I've a problem using `exec:` or defining a `SecRuleScript` with a test Lua script. Following my configurations and test:

Rule:
```
SecRule REQUEST_FILENAME "^/test" "phase:2,id:41,pass,exec:/opt/modseclua/test.lua

# or

SecRuleScript "/opt/modseclua/test.lua" "phase:5,pass"
```

My test Lua script:
```
function main()
	-- m.log(1, "Hello world!");
	return nil;
end
```

In both cases I got an empty response and an unknown error makes nginx workers to crash:

modsecurity debug log:
```
[4] (Rule: 41) Executing operator "Rx" with param "^/test" against REQUEST_FILENAME.
[9] Target value: "/test" (Variable: REQUEST_FILENAME)
[9] Matched vars updated.
[4] Rule returned 1.
[9] (SecDefaultAction) Running action: log
[9] Saving transaction to logs
[9] (SecDefaultAction) Running action: auditlog
[4] (SecDefaultAction) ignoring action: deny (rule does not cotains block)
[9] (SecDefaultAction) Running action: status
[4] Not running disruptive action: pass. SecRuleEngine is not On
[4] Running (non-disruptive) action: exec
[8] Running script... /opt/modseclua/test.lua
```

nginx error.log
```
2018/06/20 14:05:12 [alert] 10192#0: worker process 10276 exited on signal 11
```

I've try to run my script and it works:
```
root@mywebsite:~# cat /opt/modseclua/test.lua 
function main()
	print("test")
	-- m.log(1, "Hello world!");
	return nil
end

main()
root@mywebsite:~# 
root@mywebsite:~# 
root@mywebsite:~# lua5.2 /opt/modseclua/test.lua 
test
root@mywebsite:~# 
```

Any idea? Am I doing something wrong?

thanks.

p0pr0ck5:
@theMiddleBlue, what version of Nginx and modsecurity are you running? Can you show

- the output of the libmodsecurity `./configure`
- the output of `nginx -V`
- the core dump backtrace (see https://www.nginx.com/resources/wiki/start/topics/tutorials/debugging/#core-dump)

theMiddleBlue:
./configure
```
ModSecurity - v3.0.2-67-g4e3a1f71 for Linux
 
 Mandatory dependencies
[...]

PR #1814, opened 2018-06-21
Implement support for Lua 5.1

p0pr0ck5:


src/engine/lua.h
```diff
@@ -101,4 +101,27 @@ static const struct luaL_Reg mscLuaLib[] = {
...
+        lua_pushstring(L, l->name);
+        for (i = 0; i < nup; i++)  /* copy upvalues to the top */
+          lua_pushvalue(L, -(nup + 1));
+        lua_pushcclosure(L, l->func, nup);  /* closure with those upvalues */
+        lua_settable(L, -(nup + 3));
+    }
+
+    lua_pop(L, nup);  /* remove upvalues */
+}
+#endif
```
thibaultcha, line 125: _Maybe_ the intent is to eventually promote support for LuaJIT with this PR (totally speculating here), but if so, and if I recall correctly, some versions of LuaJIT define `LUA_VERSION_NUM` as `501` and implement `luaL_setfuncs`. Should we follow Mike's advice here? See:

https://www.freelists.org/post/luajit/ANN-LuaJIT210beta3,3

thibaultcha: If this isn't a goal nor a long-term goal of this PR, then feel free to ignore this!

p0pr0ck5: Hey @thibaultcha, the goal here is just to get support for 5.1 introduced. Trying to integrate with OpenResty was a bit of an exercise (see #1809), and there's no clean way to do it at this point (it was also the initial impetus for this PR). Eventually LuaJIT support might be nice, but since a number of distros only provide Lua lib packages for 5.1 out of the box, getting 5.1 supported is the sole goal here. As such I'd rather not introduce new symbols unnecessarily and increase the scope of this change.

If there's a large push for LuaJIT support here, perhaps we can revisit in the future, but at this point I do not see much value for that.

thibaultcha: All good :+1:

p0pr0ck5:
@victorhora @zimmerle any movement here?

victorhora:
All good, merged at https://github.com/SpiderLabs/ModSecurity/commit/dee989844914f07e38367fb6b1ead906541c0849. Thanks @p0pr0ck5 :)

src/engine/lua.h
victorhora: There's an attempt to support LuaJIT on some distros at https://github.com/SpiderLabs/ModSecurity/pull/1854. You might want to give it a try to see if it works for you @thibaultcha.

Issue #1818, opened 2018-06-25
TX.partial_match returned when TX.partial requested


michaelgranzow-avi:
```
"SecRuleEngine On",
"SecRule REMOTE_ADDR \"@unconditionalMatch\" \"id:1,deny,setvar:TX.partial_match=1,chain\"",
"SecRule TX.partial \"@gt 0\" \"id:2,t:lowercase,t:none,status:444\""
```
The second (chained) rule should never fire because there is no variable 'TX.partial', only 'TX.partial_match'. But it does fire:

```
[8] Saving variable: TX:partial_match with value: 1
[4] Rule returned 1.
[4] Executing chained rule.
[4] (Rule: 2) Executing operator "Gt" with param "0" against TX:partial.
[9] Target value: "1" (Variable: TX:partial)
[9] Matched vars updated.
[4] Rule returned 1.
```

zimmerle:
Hi @michaelgranzow-avi,

Thank you. We already saw your comments on victor`s pull request. It is on our queue to be fixed.

michaelgranzow-avi:
This change fixes both issues (the partial match and the performance degradation) by going back to the original code:

```
commit ec90bad44e6bf27eeae98a9ca5474b303fec1110 (HEAD -> Issue-1818_partial-match_v3-master)
Author: michaelgranzow-avi <michael@avinetworks.com>
Date:   Mon Jun 25 21:43:33 2018 +0200

    Variable names must match fully, not partially; also revert to hash table lookup instead of linear search

diff --git a/src/collection/backend/in_memory-per_process.cc b/src/collection/backend/in_memory-per_process.cc
index 65e3371d..f627aaef 100644
--- a/src/collection/backend/in_memory-per_process.cc
+++ b/src/collection/backend/in_memory-per_process.cc
@@ -105,10 +105,9 @@ void InMemoryPerProcess::resolveMultiMatches(const std::string& var,
             l->insert(l->begin(), new VariableValue(&m_name, &i.first, &i.second));
         }
     } else {
-        for (auto &a : *this) {
-            if (a.first.compare(0, var.size(), var) == 0) {
-                l->insert(l->begin(), new VariableValue(&m_name, &var, &a.second));
-            }
+        auto range = this->equal_range(var);
+        for (auto it = range.first; it != range.second; ++it) {
+            l->insert(l->begin(), new VariableValue(&m_name, &var, &it->second));
         }
     }
 }
[...]

PR #1820, opened 2018-06-26
#1818: Variable names must match fully, not partially; also revert to…

michaelgranzow-avi:
… hash table lookup instead of linear search; add test case

Propsed fix for https://github.com/SpiderLabs/ModSecurity/issues/1818


zimmerle:
fixed by: d810de9166794be2bee972fffb0e09cb0b8d6ec5

Issue #653, opened 2014-01-30
multipart boundary valid characters too strict.


diseq:
According to rfc2046 the characters allowed in the boundary are following:

```
 boundary := 0*69<bchars> bcharsnospace

 bchars := bcharsnospace / " "

 bcharsnospace := DIGIT / ALPHA / "'" / "(" / ")" /
                  "+" / "_" / "," / "-" / "." /
                  "/" / ":" / "=" / "?"
```

modsecurity msc_multipart.c checks for the following:

```
    switch(c) {
        /* Special characters not allowed. */
           case '(' :
           case ')' :
           case '<' :
           case '>' :
           case '@' :
           case ',' :
           case ';' :
           case ':' :
           case '\\' :
           case '"' :
           case '/' :
           case '[' :
           case ']' :
           case '?' :
           case '=' :
               return 0;
               break;
    }
```
- so it should be the following:
  
  ```
  switch(c) {
      /* Special characters not allowed. */
         //case '(' :  // too strict
         //case ')' :  // too strict
         case '<' :
         case '>' :
         case '@' :
         //case ',' :  // too strict
         case ';' :
         //case ':' :  // too strict
         case '\\':
         case '"' :
         //case '/' :  // too strict
         case '[' :
         case ']' :
         //case '?' :  // too strict
         //case '=' :  // too strict
             return 0;
             break;
  }
  ```

victorhora:
I believe #1821 should fix this?

victorhora:
Fixed at  https://github.com/SpiderLabs/ModSecurity/commit/7af8363fd46fd5a0bb21cc6b623d6b344f27be2a

Thanks for reporting!

PR #1821, opened 2018-06-27
Multipart boundary parsing as per RFC

allanbomsft:
Update the multipart parser to allow characters in boundaries as per RFC2046 page 21.

### DIFF
--- a/apache2/msc_multipart.c
+++ b/apache2/msc_multipart.c
@@ -695,40 +695,27 @@ static int multipart_boundary_characters_valid(char *boundary) {
 
     if (p == NULL) return -1;
 
-    while((c = *p) != '\0') {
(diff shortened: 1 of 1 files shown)


victorhora:
This would fix #653?

victorhora:
According to the [RFC2046 page 22](https://tools.ietf.org/html/rfc2046#page-22), indeed the parser seemed to be more strict than it should. I believe this is safe to be merged.

As I'm not a fan of writing test cases for v2 (:P), I've written a test case for v3 and adapted this patch at https://github.com/SpiderLabs/ModSecurity/pull/1955.

victorhora (approved):

zimmerle:
merged ;) thanks.

PR #1823, opened 2018-06-27
IIS, buffer request body before taking lock

allanbomsft:
IIS, buffer request body before taking lock. This is to prevent bad clients from sending requests slowly and thereby holding the lock indefinitely.


allanbomsft:
duplicate PR...

PR #1826, opened 2018-06-29
IIS: no lock on ProcessRequest. No reload of config. (#24)

allanbomsft:
IIS: No lock on ProcessRequest. This allows for concurrent requests to happen on IIS.

Also removed the auto reload of config, as this is dangerous to do now where one thread might be using the config while another thread might be reloading it. It's worth sacrificing auto reload for the huge benefit of concurrent requests.

In this pull request the diff alignment around the if(pConfig->m_Config == NULL) section is hard to read, but what I've basically done is taken away the parts that dealt with checking for timestamps.


zimmerle:
Merged! Thanks!

PR #1847, opened 2018-07-22
Fix pc file

gquintard:
the include path can't have @PACKAGE@ at the end as we include headers
with

 #include <modsecurity/modsecurity.h>

so it's already in there


zimmerle:
Merged! Thanks!

PR #1852, opened 2018-07-26
Fix OpenBSD build

### LINKED ISSUE #1841: OpenBSD doesn't have the header wordexp.h
Hi, I'm trying to create a port of libmodsecurity for OpenBSD. Unfortunately, `src/utils/string.cc` and `src/utils/system.cc`need the header `wordexp.h` which is not available on OpenBSD. Can you make a workaround for this?. The OpenBSD developers [will never implement this API](http://openbsd-archive.7691.n7.nabble.com/patch-libc-wordexp-support-td159368.html).

victorhora:
OpenBSD build was breaking due to the lack of _wordexp.h_. Patch fixes the build by using glob.h when building on OpenBSD. Should fix #1841.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.0.3 - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+ - Fix OpenBSD build
+   [Issue #1841 - @victorhora, @zimmerle, @juanfra684]
  - Variable names must match fully, not partially. Match should be case
    insensitive.
    [Issue #1818, #1820, #1810, #1808 - @michaelgranzow-avi, @victorhora,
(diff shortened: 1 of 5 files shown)


victorhora:
Maybe we should consider using _glob.h_ by default on all platforms as a matter of best practice due to security concerns with _wordexp_. That being said, according to the docs, glob.h has "No tilde expansion or parameter substitution is done; if you want these, use wordexp(3)."

[GLOB_TILDE flag ](https://www.gnu.org/software/libc/manual/html_node/More-Flags-for-Globbing.html#More-Flags-for-Globbing)could be used for tilde expansion, but as this would require further changes I refrained from doing this at the moment.

victorhora:
Wondering if something like: 
```
Include test-cases/data/config_example2*.*
```

Would be a valid test scenario to prove that expandEnv() is also functionally correct when using glob.

Thoughts?

zimmerle:
Merged! Thanks!

Issue #1884, opened 2018-08-30
GeoLookup does not work


everping:
I tried to block IPs from a specific country but GeoLookup does not work.

- OS: Ubuntu 16.04.5 LTS
- modsecurity v3.0.3
- modSecurity-nginx v1.0.0
- nginx v1.14.0
- libgeoip1, libgeoip-dev, geoip-bin was installed

nginx virtual host

```
    modsecurity on;
    modsecurity_rules '
        # Include the recommended configuration
        Include /etc/nginx/modsec/modsecurity.conf

        # OWASP CRS v3 rules
        Include /usr/local/owasp-modsecurity-crs-3.0.0/crs-setup.conf
        Include /usr/local/owasp-modsecurity-crs-3.0.0/rules/*.conf

        # Audit log
        SecAuditLog /var/log/modsec/mysite.com/audit.log

        # Custom rules
        Include /etc/nginx/modsec/mysite.com/*.conf
    ';

```

/usr/local/owasp-modsecurity-crs-3.0.0/crs-setup.conf
```
...
SecGeoLookupDB util/geo-location/GeoIP.dat
...
```
/etc/nginx/modsec/mysite.com/main.conf
```
SecRule REMOTE_ADDR "@geoLookup" "chain,id:1,drop,msg:'Non-VN IP address'"
SecRule GEO:COUNTRY_CODE "!@streq VN"
```

And the debug log I got

```
[4] (Rule: 1) Executing operator "GeoLookup against REMOTE_ADDR.
[9] Target value: "x.x.x.x" (Variable: REMOTE_ADDR)
[4] Rule returned 0.
```

This means the remote address could not be looked up. Is there anyone can tell me what is wrong here?

victorhora:
@everping 

Can you tell us what is the output of your configure script when setting up libModSecurity? It should be something like this:

```
ModSecurity - v3.0.2-78-g90197bd for Linux

 Mandatory dependencies
   + libInjection                                  ....v3.0.2-78-g90197bd
   + SecLang tests                                 ....90197bd

 Optional dependencies
   + GeoIP/MaxMind                                 ....found
      * (GeoIP) v1.6.9
         -lGeoIP  , -I/usr/include/
   + LibCURL                                       ....found v7.47.0
      -L/usr/lib/x86_64-linux-gnu -lcurl,  -DWITH_CURL_SSLVERSION_TLSv1_2 -DWITH_CURL
   + YAJL                                          ....found v2.1.0
[...]

PR #1895, opened 2018-09-11
Fix parser to support GeoLookup with MaxMind

victorhora:
Should fix issue where [SecGeoLookupDb](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#SecGeoLookupDb) directive and [@geoLookup](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#geoLookup) operator was not working as expected when using libMaxMind. 

Also a small change to libmaxmind.m4 to allow the script to find the headers on Ubuntu based distros where PkgConfig is not available. Should fix issue #1884

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.0.3 - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+ - Fix parser to support GeoLookup with MaxMind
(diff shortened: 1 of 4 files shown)


zimmerle:
Merged! thanks!

Issue #1888, opened 2018-09-04
Are SecMarkers Broken in 3.0.2?


stevendore:
I am working on loading rules with [modsecurity 3.0.2](https://github.com/SpiderLabs/ModSecurity/commit/90197bdd99eef6b874f9b09a413d9ded06ec26f3). I am using the rules from[ `v3.2/dev`](https://github.com/SpiderLabs/owasp-modsecurity-crs/commit/9ec587c0d0bf14557ceab35785488ba8c1b28c12). When attempting to include `modsecurity.conf-recommended`, `crs-setup.conf` and `rules/*.conf`, I am getting the error, `Rule id: 0 is duplicated` from `/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf` onward. After messing around I tried removing `SecMarker ...` and I was then able to load the rules (baring the, I believe, known errors with `rules/REQUEST-912-DOS-PROTECTION.conf` and `rules/REQUEST-914-FILE-DETECTION.conf` [1]. Here is a link to the [filebin](https://filebin.net/rglofqcruf2llkno) after removing `SecMarker`. Is this the actual issue or am I getting these working by doing something wrong?

[1] 
```
Rules error. File: /owasp-modsecurity-crs/rules/REQUEST-912-DOS-PROTECTION.conf. Line: 187. Column: 29. Expecting an action, got:  ,\

Rules error. File: /owasp-modsecurity-crs/rules/REQUEST-914-FILE-DETECTION.conf. Line: 70. Column: 85. Expecting a variable, got:  :  S}: %{MATCHED_VARS}',\
```

stevendore:
Here is a diff for the rule changes:
```
diff --git a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
index f44c90e..c6823ce 100644
--- a/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
+++ b/rules/REQUEST-903.9001-DRUPAL-EXCLUSION-RULES.conf
@@ -377,6 +377,3 @@ SecRule REQUEST_FILENAME "@endsWith /admin/config/services/rss-publishing" \
     pass,\
     nolog,\
     ctl:ruleRemoveTargetByTag=CRS;ARGS:feed_description"
-
-
[...]

PR #1901, opened 2018-09-15
Fix RulesProperties::appendRules()

stevendore:
After a discussion on #1888 I noticed `RulesProperties::appendRules()` was not checking for duplicate IDs as well as throwing an error if there were `secMarkers` in more than one file (when calling any combination of `rules->load()`, `rules->loadFromUri()` or `rules->loadRemote()` more than once). To fix the `secMarker` issue, the if statement on [rules_properties.h:441](https://github.com/SpiderLabs/ModSecurity/blob/c2bc6952651707f773280b7b8188db814aa11718/headers/modsecurity/rules_properties.h#L441) just needed to be negated.

This function also doesn't accurately check for duplicate IDs. The check can be circumvented by putting the rule in a different phase. To fix this, the ruleId list (`v` in this function) has to be populated completely before checking it against the other list.

Before the patch is applied the example below will fail to load rules due to multiple files containing `secMarkers`

test01.conf
```
SecRule REQUEST_FILENAME "@endsWith /wp-login.php" "id:1,phase:2"
SecRule REQUEST_FILENAME "@endsWith /private" "id:2,phase:1"
SecMarker END-ONE
```

test02.conf
```
SecRule UNIQUE_ID "@rx ^." "id:3,phase:3"
SecMarker END-TWO
```

test03.conf
```
SecRule REQUEST_FILENAME "@endsWith /admin" "id:4,phase:6"
SecMarker END-THREE
```

```bash
x@x:/code/ModSecurity/tools/rules-check$ ./modsec-rules-check /code/test_conf/*.conf
 : /code/test_conf/test01.conf  --  Loaded 9 rules.
 : /code/test_conf/test02.conf  --  Loaded -1 rules.
    Rule id: 0 is duplicated

 : /code/test_conf/test03.conf  --  Loaded -1 rules.
    Rule id: 0 is duplicated

Test failed.
```

After removing the `SecMarkers` from test02 and test03 the test will pass.

```
x@x:/code/ModSecurity/tools/rules-check$ ./modsec-rules-check /code/test_conf/*.conf
 : /code/test_conf/test01.conf  --  Loaded 9 rules.
 : /code/test_conf/test02.conf  --  Loaded 1 rules.
 : /code/test_conf/test03.conf  --  Loaded 1 rules.
Test ok.
```
And if you force a duplicate `ruleId` by changing the Id in test03 to 1, the test will still pass.

After the patch all of these issues are dealt with in the correct manner.
The test will pass with the original conf files above and will fail if you change the ID in test03 to 1.


djellemah:
+1 on this. I'm running into the same issue (duplicate ID errors when merging Rules instances via the C api) and this pull request fixes that.

victorhora:
Merged as of https://github.com/SpiderLabs/ModSecurity/commit/2cebba911699e3fb05b141b869f213b9754e878a.

Thanks for your contribution! :)

PR #1910, opened 2018-09-20
Allow LuaJIT 2.1 to be used

### LINKED ISSUE #1909: ModSecurity 3 no longer compiles with LuaJIT 2.1
A recent change in ModSecurity seems to have broken compatibility with LuaJIT (which is compatible with Lua 5.1).  The following is the error when compiling:

```
In file included from ../src/rule_script.h:22:0,
                 from seclang-parser.yy:20,
                 from seclang-parser.cc:46:
../src/engine/lua.h: In function 'void luaL_setfuncs(lua_State*, const luaL_Reg*, int)':
../src/engine/lua.h:111:13: error: 'void luaL_setfuncs(lua_State*, const luaL_Reg*, int)' was declared 'extern' and later 'static' [-fpermissive]
 static void luaL_setfuncs(lua_State *L, const luaL_Reg *l, int nup) {
             ^~~~~~~~~~~~~
In file included from /usr/include/luajit-2.1/lua.hpp:5:0,
                 from ../src/engine/lua.h:17,
                 from ../src/rule_script.h:22,
                 from seclang-parser.yy:20,
                 from seclang-parser.cc:46:
/usr/include/luajit-2.1/lauxlib.h:88:18: note: previous declaration of 'void luaL_setfuncs(lua_State*, const luaL_Reg*, int)'
 LUALIB_API void (luaL_setfuncs) (lua_State *L, const luaL_Reg *l, int nup);
                  ^~~~~~~~~~~~~
make[3]: *** [Makefile:2013: parser/libmodsecurity_la-seclang-parser.lo] Error 1
```

victorhora:
Simple change to allow for LuaJIT 2.1 to be used with ModSecurity. 

The issue happened due to luaL_setfuncs() already being defined in LuaJIT 2.1 (see [lauxlib.h#L88](https://github.com/LuaJIT/LuaJIT/blob/v2.1/src/lauxlib.h#L88)).

Commit dee9898#diff-f120b67754e2c3ff8a98d4e3007f12efR111 defines this func as it is being used by the Lua engine inside ModSecurity (lua.cc#L138) but it is lacking in Lua 5.1 to make it compatible.

An extra if defined condition (WITH_LUA_JIT_2_1) was added to cover for LuaJIT 2.1.

Should fix #1909.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.0.3 - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+ - Allow LuaJIT 2.1 to be used
+   [Issue #1909 - @victorhora, @mdunc]
  - Fix: function m.setvar in Lua scripts and add testcases
    [Issue #1859 - @nowaits, @victorhora]
  - Fix SecResponseBodyAccess and ctl:requestBodyAccess directives
--- a/build/lua.m4
+++ b/build/lua.m4
@@ -68,7 +68,8 @@ else
         case $LUA_PKG_VERSION in
            (5.1*) LUA_CFLAGS="-DWITH_LUA_5_1 ${LUA_CFLAGS}" ; lua_5_1=1 ;;
            (5.2*) LUA_CFLAGS="-DWITH_LUA_5_2 ${LUA_CFLAGS}" ; lua_5_2=1 ;;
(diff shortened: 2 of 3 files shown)


victorhora:
Merged as of https://github.com/SpiderLabs/ModSecurity/commit/20ef01d75c3c4ec5ed32ef416d7db9af3a2fcd99

PR #1917, opened 2018-09-27
IIS: Remove body prebuffering again. Unneeded due to no lock on modse…

allanbomsft:
IIS: Remove body pre-buffering again, which I had introduced a few months back. This is no longer needed, because now we have also removed the locking around modsecProcessRequest, so when modsecProcessRequest calls ReadBodyCallback which might block if the client is sending slow, it is no longer holding the lock.

The implementation I introduced a few months ago, where we pre-buffer, has a bug. It causes a freeze if SecRequestBodyAccess is set to Off. This is because if the body is read from IIS, IIS also expects the body the be written back, which I had forgotten. Reverting the pre-buffering fixes this.

Tested by running 150 concurrent diverse requests and ensuring the expected return value.

This is the commit we are undoing: https://github.com/SpiderLabs/ModSecurity/commit/18af2597779c0e8947e9541104f6100c1667ba22

### DIFF
--- a/iis/mymodule.cpp
+++ b/iis/mymodule.cpp
@@ -32,12 +32,6 @@
 
 #include "winsock2.h"
 
-// Used to hold each chunk of request body that gets read before the full ModSec engine is invoked
-typedef struct preAllocBodyChunk {
-    preAllocBodyChunk* next;
-    size_t length;
-    void* data;
-} preAllocBodyChunk;
 
 class REQUEST_STORED_CONTEXT : public IHttpStoredContext
 {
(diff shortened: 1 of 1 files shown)


victorhora (commented):
I believe this is safe to be merged. Pushing this PR into a test branch: https://github.com/SpiderLabs/ModSecurity/commit/904ab510b53f72e11ef41f8b84ac0121c40fb1fc

This should trigger IIS buildbots and we can merge after as per procedure :)

Thansk :)

victorhora:
http://www.modsecurity.org/developers/buildbot/builders/Win8%20-%20%28VS%202012%2064b%29%20-%20IIS/builds/133

victorhora:
Merged here: https://github.com/SpiderLabs/ModSecurity/commit/a55a9481b3eb29495814f80c12b654216be14674

Thank :)

Issue #1916, opened 2018-09-27
[V3] callback function not receiving http return code


stevendore:
Working with the callback function the `message` always contains a `%d` instead of an error code. 
ex: `ModSecurity: Access denied with code %d (phase 2). Matched "Operator Rx with parameter a...`. 

Even though in the audit log, the line will contain the code instead of `%d`
ex: `ModSecurity: Access denied with code 200 (phase 2). Matched "Operator Rx with parameter a...`

I am invoking the call back by:

```C
#include <ctype.h>
#include <glob.h>
#include <arpa/inet.h>

#include <modsecurity/intervention.h>
#include <modsecurity/rules.h>

// data is an output stream that is needed to print
static void msc_callback(void *data, const void *message)
{
	func(data, "%s", (const char *)message);
}
int main() {
const char *err;

ModSecurity *modsec = msc_init();
Rules *rules = msc_create_rules_set();
Transaction *txn = msc_new_transaction(waf->modsec, waf->rules, (void *)data);
msc_set_log_cb(modsec, (void *)msc_callback);

if (msc_rules_add_file(rules, pglob.gl_pathv[0], &err) < 0) {
        printf("rules error: %s", err);
}

/*
... add and process request/response headers and bodies ...
*/

msc_process_logging(txn);
}
```

After some digging through the code for the process for the callback I found that the callback is only called through [`Transaction::serverLog()`](https://github.com/SpiderLabs/ModSecurity/blob/bc3d3f19154793e23568103b931a15168b34d768/src/transaction.cc#L1783) which contains the function[ `ModSecurity::serverLog(m_logCbData, rm)`](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/modsecurity.cc#L189). ModSecurity has no access/reference to the `Transaction::m_httpCodeReturned` so it would have had to been passed by ` Transaction::serverLog()` with `this->m_httpCodeReturned` to be able to be passed into the[ `RuleMessage::log()`](https://github.com/SpiderLabs/ModSecurity/blob/bc3d3f19154793e23568103b931a15168b34d768/src/rule_message.cc#L65) to generate the callback message. I believe this patch can (and has in small testing) fix it : 

```diff
diff --git a/headers/modsecurity/modsecurity.h b/headers/modsecurity/modsecurity.h
index 8fa9e59..6157dae 100644
--- a/headers/modsecurity/modsecurity.h
+++ b/headers/modsecurity/modsecurity.h
[...]

PR #1919, opened 2018-09-28
[V3] Allow the callback function to receive and log a http return code

stevendore:
As discussed in #1916, there is an issue where the callback function will not insert the http return code into the message even though the audit log contained the correct message.
ex: callback: `ModSecurity: Access denied with code %d (phase 2). Matched "Operator Rx with parameter a....`
auditlog: `ModSecurity: Access denied with code 200 (phase 2). Matched "Operator Rx with parameter a...`

After some digging through the code for the process for the callback I found that the callback is only called through [`Transaction::serverLog()` ](https://github.com/SpiderLabs/ModSecurity/blob/bc3d3f19154793e23568103b931a15168b34d768/src/transaction.cc#L1783)which contains the function[ `ModSecurity::serverLog(m_logCbData, rm)`](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/modsecurity.cc#L189). ModSecurity has no access/reference to the `Transaction::m_httpCodeReturned` so it would have had to been passed by `Transaction::serverLog()` with `this->m_httpCodeReturned` to be able to be passed into the [`RuleMessage::log()`](https://github.com/SpiderLabs/ModSecurity/blob/bc3d3f19154793e23568103b931a15168b34d768/src/rule_message.cc#L65) to generate the callback message. After this patch I am able to get the http return code into the callback function message parameter.

Though I am unsure how/if props should be set or in `ModSecurity::serverLog()`. I am also unsure in the case that `RuleMessageLogProperty` is set, how to best pass the `m_httpCodeReturned` to the rule message or callback function.

Additionally I noticed in [`ModSecurity::serverLog()` ](https://github.com/SpiderLabs/ModSecurity/blob/bc3d3f19154793e23568103b931a15168b34d768/src/modsecurity.cc#L207)there might be bug due to the repeated code in the if statement.

### DIFF
--- a/headers/modsecurity/modsecurity.h
+++ b/headers/modsecurity/modsecurity.h
@@ -289,7 +289,7 @@ class ModSecurity {
      */
     void setServerLogCb(ModSecLogCb cb, int properties);
 
-    void serverLog(void *data, std::shared_ptr<RuleMessage> rm);
+    void serverLog(void *data, std::shared_ptr<RuleMessage> rm, int httpCodeReturned);
 
     const std::string& getConnectorInformation();
 
--- a/src/modsecurity.cc
+++ b/src/modsecurity.cc
@@ -186,7 +186,7 @@ const std::string& ModSecurity::getConnectorInformation() {
     return m_connector;
 }
 
-void ModSecurity::serverLog(void *data, std::shared_ptr<RuleMessage> rm) {
+void ModSecurity::serverLog(void *data, std::shared_ptr<RuleMessage> rm, int httpCodeReturned) {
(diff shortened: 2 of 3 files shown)


zimmerle:
Hi @steven-j-wojcik,

As of v3.0.3 the logging schema was restructured, as a consequence this issue was mitigated. I am sorry to close this without a merge. Thanks for the contribution. I am putting your name as a contributor for the previous change.
https://github.com/SpiderLabs/ModSecurity/blob/af137442d5389fb24c5dece2dd951a2ce9e9c9aa/CHANGES#L31-L32

PR #1920, opened 2018-09-30
The database should be secured

shivamk01:
Digital Signature certificates should be generated using elliptic curve digital signature algorithm which is very secure so that the system cannot be copied or manipulated and the integrity is maintained.


victorhora:
@shivamk01 I believe there's something wrong with this Pull request. It seems like you're trying to merge the [v3/dev/test](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/test) branch with the [v3/dev/test](https://github.com/SpiderLabs/ModSecurity/tree/v2/master) branch?

> Digital Signature certificates should be generated using elliptic curve digital signature algorithm which is very secure so that the system cannot be copied or manipulated and the integrity is maintained.

As far as I know, there's no "Digital Signature certificates" being generated by ModSecurity so I will be closing this Pull request as I'm assuming it was created by error/accident.

PR #1949, opened 2018-11-07
Replaced log locking using mutex with fcntl lock

Cloaked9000:
Switching from mutexes to file locking for logging. Fixes https://github.com/SpiderLabs/ModSecurity/issues/1927


zimmerle:
Hi @Cloaked9000,

Sorry for the delay in looking into this. There is indeed a very clever approach. Thank you. Merged.

defanator:
@Cloaked9000 kudos for this one!

PR #1955, opened 2018-11-14
Make the boundary check less strict as per RFC2046

victorhora:
This PR is based on #1821 but directed at the v3 branch and with a test case to validate the changes.

According to the [RFC2046 page 22](https://tools.ietf.org/html/rfc2046#page-22), indeed the parser seemed to be more strict than it should. I believe this is safe to be merged.


zimmerle:
Merged! thanks

PR #1966, opened 2018-11-28
Fix "make dist" target to include default configuration

defanator:
This change allows to grab fully functional distribution archive from arbitrary future commit, so there will no need to gather `modsecurity.conf` and its prerequisites (namely `unicode.mapping`) separately.


zimmerle:
merged! thanks.

PR #1978, opened 2018-12-10
Complete merging of particular rule properties

defanator:
Closes SpiderLabs/ModSecurity-nginx#142 issue.

### DIFF
--- a/headers/modsecurity/rules_properties.h
+++ b/headers/modsecurity/rules_properties.h
@@ -343,10 +343,12 @@ class RulesProperties {
 
         if (from->m_requestBodyLimit.m_set == true) {
             to->m_requestBodyLimit.m_value = from->m_requestBodyLimit.m_value;
+            to->m_requestBodyLimit.m_set = true;
         }
 
         if (from->m_responseBodyLimit.m_set == true) {
             to->m_responseBodyLimit.m_value = from->m_responseBodyLimit.m_value;
+            to->m_responseBodyLimit.m_set = true;
         }
 
         if (from->m_requestBodyLimitAction != PropertyNotSetBodyLimitAction) {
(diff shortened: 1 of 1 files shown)


defanator:
The PR may be incomplete, see 
https://github.com/SpiderLabs/ModSecurity-nginx/issues/142#issuecomment-445799318 (and below) for details.

zimmerle:
Hi @defanator,

Thank you for the report :) that is an interesting subject, because it lead us to a bigger discussion. Back on v2, some configuration items had o global scope even when placed inside a virtual host. That is annoying as hell, as it was counterintuitive. Reading the name of the configuration is/was not possible to tell (or to expect) when/how/if it is going to work.

There are some technical limitations that are important, for instance, very earlier on the request - when the vhost and/or the destination folder is not know -  is the place that we want to threat _slow dos_. Therefore this particular configuration will no be able to target a single vhost, but the entire server.

That may not be the case of SecRequestBodyLimit. Yet, it is on the same set of the configurations class that will be good to be not-global. IMHO, the only configurations that should be global are the ones that  can't be different due to technical limitations. Working towards it. On **v3/dev/reload** we are have a differentiation between what is a rule and what is a configuration.

As **v3/dev/reload** is not yet ready for production, let me see what can i do to circumvent this specific problem.

defanator:
@zimmerle in other words, a configuration like the one described in 
https://github.com/SpiderLabs/ModSecurity-nginx/issues/142#issuecomment-445799318 definitely should be considered as valid, right?

Let me know if you have any other concerns that may (or may not) be related to v3/dev/reload in this context (i.e. if that branch is going to be heavily refactored and will unlikely go into 3.0.4, I'd merge this one into v3/master as it fixes the initial issue reported at https://github.com/SpiderLabs/ModSecurity-nginx/issues/142 - it wasn't about _redefining_ values, it's about the fact that the documented behavior of SecRequestBodyLimit does not work entirely if `modsecurity_rules_file` is specified in top contexts).

zimmerle:
Merged! Thanks :)

defanator:
@zimmerle oh, I was going to update the PR to do actual merges like here - 
https://github.com/SpiderLabs/ModSecurity-nginx/issues/142#issuecomment-445841791 - so I'll test this once again and submit a new one.

Also, here's the full list of directives which do not get merged:

```
SecRuleEngine
SecRequestBodyAccess
SecResponseBodyAccess
SecXmlExternalEntity
SecUploadKeepFiles
SecTmpSaveUploadedFiles
SecRequestBodyLimit
SecResponseBodyLimit
SecRequestBodyLimitAction
SecResponseBodyLimitAction
SecUploadFileLimit
SecUploadFileMode
```

Issue #1983, opened 2018-12-14
Use of AC_CHECK_FILE prevents cross compilation.


chuckwolber:
When cross compiling modsecurity via a Yocto Project bitbake recipe, the use of AC_CHECK_FILE in configure.ac causes an error:

`configure: error: cannot check for file existence when cross compiling`

I am working on a fix and should be submitting a pull request shortly.

PR #1984, opened 2018-12-14
Use of AC_CHECK_FILE prevents cross compilation.

chuckwolber:
The use of AC_CHECK_FILE causes the following error when cross compiling:

  configure: error: cannot check for file existence when cross compiling

The solution is to check for the file directly instead of using a macro.

Resolves: #1983


zimmerle:
Merged! Thanks!

PR #1992, opened 2018-12-29
base64 decode

tdoubley:
1. the base64decode not check the valid of input string，
if input is a invalid base64 string `VGVzdENhc2U=VGVzdENhc2U=`, we will get a `empty string`, I think it coulde lead false positive.

2. the input string len is strlen(data.c_str()), if input is `VGVzdENhc2U=\\u0000VGVzdENhc2U=`,  the variable len equal to 12, we will loss part of data. I'm not sure if it was a mistake?

### DIFF
--- a/src/operators/pm_from_file.cc
+++ b/src/operators/pm_from_file.cc
@@ -25,6 +25,23 @@
 namespace modsecurity {
 namespace operators {
 
+bool PmFromFile::isComment(const std::string &s) {
+    if (s.size() == 0) {
+	  return true;
+    }
+    size_t pos = s.find("#");
+    if (pos != std::string::npos) {
+        for (int i = 0; i < pos; i++) {
+	    if (!std::isspace(s[i])) {
+		return false;
+	    }
+	}
+    } else {
+	return false;
+    }
+
+    return true;
+}
 
 bool PmFromFile::init(const std::string &config, std::string *error) {
     std::istream *iss;
@@ -50,7 +67,9 @@ bool PmFromFile::init(const std::string &config, std::string *error) {
     }
 
     for (std::string line; std::getline(*iss, line); ) {
-        acmp_add_pattern(m_p, line.c_str(), NULL, NULL, line.length());
+        if (isComment(line) == false) {
+            acmp_add_pattern(m_p, line.c_str(), NULL, NULL, line.length());
+	}
     }
 
     while (m_p->is_failtree_done == 0) {
--- a/src/operators/pm_from_file.h
+++ b/src/operators/pm_from_file.h
@@ -36,6 +36,8 @@ class PmFromFile : public Pm {
         : Pm(n, std::move(param)) { }
 
     bool init(const std::string &file, std::string *error) override;
+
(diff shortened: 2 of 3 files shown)


src/operators/pm_from_file.cc
```diff
@@ -25,6 +25,23 @@
...
+	    if (!std::isspace(s[i])) {
+		return false;
+	    }
+	}
+    } else {
+	return false;
+    }
+
+    return true;
+}
```
zimmerle, line 44: Hi @tdoubley,

I think this specific commit is related to other issue :)  Should I take it in consideration? or is something that you are still working on?

zimmerle:
Hi @tdoubley, 

Thank you for the patches.

Reading your code I notice that the real change is considering the return of ```mbedtls_base64_decode```; if equals to *MBEDTLS_ERR_BASE64_INVALID_CHARACTER* return the original data instead of an empty string. Not sure if that is the best approach given the fact that it may lead the user to think that the decoded string is the string itself.

The operator modificatior *&* can be used in order to verify is the output was consistent or not.

I also notice that we return the original data if there is a memory allocation problem. We may want to change that as well.

zimmerle:
Addressing your concern about the addition to the null byte. That is indeed a known evasion technique. To address that the user can make usage of [removeNulls](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#removeNulls) transformation. Together with the [multimatch](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#multimatch) gives the user a huge surface to work with.

src/operators/pm_from_file.cc
tdoubley: Hi @zimmerle  

I have just read the latest code ,  it‘s surly releated to [issue1948](https://github.com/SpiderLabs/ModSecurity/issues/1948), i think those issue can be solved by this commit.

victorhora:
> the real change is considering the return of `mbedtls_base64_decode`; if equals to _MBEDTLS_ERR_BASE64_INVALID_CHARACTER_ return the original data instead of an empty string. Not sure if that is the best approach given the fact that it may lead the user to think that the decoded string is the string itself.
> 
> The operator modificatior _&_ can be used in order to verify is the output was consistent or not.
> 
> I also notice that we return the original data if there is a memory allocation problem. We may want to change that as well.

> Addressing your concern about the addition to the null byte. That is indeed a known evasion technique. To address that the user can make usage of removeNulls transformation. Together with the multimatch gives the user a huge surface to work with.

There's also the fact that we have [base64DecodeExt](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#base64DecodeExt) which uses a more forgiving implementations of Base64 decoding which should cover the use case that @tdoubley patch covers?

zimmerle:
Hi @tdoubley,

Do you still have an interest in work with this pull request?

zimmerle:
Thank for your interest. Please open if you are still interested in the contribution.

PR #1997, opened 2019-01-11
fix logging issues

arminabf:
mod_security does not allow per-module log level configuration nor to log module information into error_log. For example, following does not work

`LogLevel security2:info`
`ErrorLogFormat "{"\"module\":\"%m\", \"level\":\"%l\", \"message\":\"%M\"}"`

In the later example no module information is logged. Commit 1f4b951 fixes these issues.

In addition, custom log formats are not applied for messages being logged at startup. Commit b9e8317 fixes some (even though not all) log messages where only the default log format is applied.

Please note that there are more log messages having that problem but that would need far-reaching changes. For example in the function msc_status_engine_call() that is also  used by standalone/api.c where no server_rec is available.

### DIFF
--- a/apache2/mod_security2.c
+++ b/apache2/mod_security2.c
@@ -105,39 +105,39 @@ static int server_limit, thread_limit;
 *
 * \param mp Pointer to memory pool
 */
-static void version(apr_pool_t *mp) {
+static void version(apr_pool_t *mp, server_rec *s) {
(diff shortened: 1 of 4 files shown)


marcstern:
Replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3166

Issue #1996, opened 2019-01-10
Support for RE2 regular expression engine


WGH-:
RE2 (https://github.com/google/re2) is a regular expression engine written in C++ and developed by Google.

Unlike libpcre, RE2 runtime is always linear to the size of the input. It comes at some cost: the engine inherently doesn't support certain features, like backreferences and lookaround assertions.

Still, even the CRS has some regular expressions in its rules that run very slowly (minutes) on certain inputs, but are supported by RE2 and run much faster (seconds) on the same inputs.

The idea of adding RE2 support to ModSecurity has been floating around for a while, and there's quite a few mentions of people working on things related to it:
* SpiderLabs/owasp-modsecurity-crs#1255
* SpiderLabs/owasp-modsecurity-crs#1130

I have been unable to find neither issue nor public pull request/fork/branch of ModSecurity with RE2 support.

I have a more-or-less working proof-of-concept patch for ModSecurity with optional RE2 support (with fallback to libpcre if the regexp fails to compile on RE2).

Is someone else working on this? Because if yes, it's probably not worth it to duplicate efforts. If not, I think I can clean up my patch a bit and submit it here as a PR.

zimmerle:
Hi @WGH-,

Thinking in that possibility, in ModSecurity v3 we have this as plug able interface. So that we could support provide for different regular expression engines. It just demands a single file change, here:

https://github.com/SpiderLabs/ModSecurity/blob/3c1fba278c14fe9b63cff80a3ae32df82ba042ac/src/utils/regex.cc#L41-L129

Ideally the regular expression back-end could be selected with a compilation flag, selected during the configuration.

[Hyperscan](https://github.com/intel/hyperscan) can be also taken into consideration.

WGH-:
It seems nobody is sitting on an unpublished patch (or at least doesn't admit to), so I'll try to clean up and submit mine this week.

zimmerle:
Great to hear that you have a patch already. Let us known if you need further assistance.

zimmerle:
@WGH- starts the work in that feature with an excellent patch submitted at #2003. Patch is already merged.

zimmerle:
Hi, @WGH-,

[...]

PR #2003, opened 2019-01-16
Refactor regex code

WGH-:
While proceeding with preparation of my RE2 patch (#1996), I found quite a few odd things with regex code, and decided to attempt to fix them first:

 * Lack of encapsulation.
 * Non-method functions for matching without retrieving all groups.
 * Regex class being copyable without proper copy-constructor (potential UAF
   and double free due to pointer members `m_pc` and `m_pce`).
 * Redundant `SMatch::m_length`, which always equals to `match.size()` anyway.
 * Weird `SMatch::size_` member which is initialized only by one of the three matching
   functions, and equals to the return value of that function anyways.
 * Several places in code having `std::string` value instead of reference.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -37,6 +37,8 @@ v3.0.4 - YYYY-MMM-DD (to be released)
    [Issue #1943 - @victorhora, @allanbomsft]
  - Fix buffer size for utf8toUnicode transformation
    [Issue #1208 - @katef, @victorhora]
+ - Refactoring on Regex and SMatch classes.
+   [@WGH-]
 
 
 v3.0.3 - 2018-Nov-05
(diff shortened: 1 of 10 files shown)


zimmerle:
Hi @WGH-,

Thank you for the patch. Having a look at it tomorrow.

zimmerle:
Patch merged! Made minimal cosmetic changes to make the style check script happy again.

```
$ make check-coding-style | grep regex 
warning: ./src/utils/regex.cc:43:  { should almost always be at the end of the previous line  [whitespace/braces] [4]
warning: ./src/utils/regex.cc:112:  Closing ) should be moved to the previous line  [whitespace/parens] [2]
```

I do have suggestions for more refactoring surround this particular implementation. I will send it over you via patch.

mdunc:
I think this caused my build to fail when compiling ModSecurity for nginx.  Configured with the `--with-lmdb` option.  

```
collection/backend/lmdb.cc: In member function ‘virtual void modsecurity::collection::backend::LMDB::resolveRegularExpression(const string&, std::vector<const modsecurity::VariableValue*>*, modsecurity::Variables::KeyExclusions&)’:
collection/backend/lmdb.cc:541:38: error: use of deleted function ‘modsecurity::Utils::Regex::Regex(const modsecurity::Utils::Regex&)’
```

Edit: Created an issue for this https://github.com/SpiderLabs/ModSecurity/issues/2008

PR #2007, opened 2019-01-22
Add support for capturing group test cases

WGH-:
This enables unit tests to compare the matching groups as well, not just binary match-no match.

The format of JSON test case is extended, new `"re_groups"` parameter is added:
```json
{
      "type" : "op",
      "name" : "rx",
      "param" : "(abc)(def)",
      "input" : "abcdef",
      "ret" : 1,
      "re_groups": ["abcdef", "abc", "def"]
},
{
      "type" : "op",
      "name" : "rx",
      "param" : "(abc)(def)",
      "input" : "abcdef   abcdef",
      "ret" : 1,
      "re_groups": ["abcdef", "abc", "def", "abcdef", "abc", "def"]
}
```

New tests cases will need to be added to https://github.com/SpiderLabs/secrules-language-tests, of course.

### DIFF
--- a/test/unit/unit.cc
+++ b/test/unit/unit.cc
@@ -41,6 +41,10 @@
 using modsecurity_test::UnitTest;
 using modsecurity_test::ModSecurityTest;
 using modsecurity_test::ModSecurityTestResults;
+using modsecurity::ModSecurity;
+using modsecurity::Rule;
+using modsecurity::Rules;
+using modsecurity::Transaction;
 using modsecurity::actions::transformations::Transformation;
 using modsecurity::operators::Operator;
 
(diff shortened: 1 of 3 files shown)


zimmerle:
Hi @WGH-,

Are you sending the test cases as well?

WGH-:
Hi! Yes, I have some test-cases as well. I might find more interesting test cases while working on RE2, which I'm currently doing, that's why I haven't sent them yet.

While we're at it, do you know why `searchAll` have such an awkward interface? I.e. flattened list of all groups of all matches (see second test case).

zimmerle:
v3 was basically con

> Hi! Yes, I have some test-cases as well. I might find more interesting test cases while working on RE2, which I'm currently doing, that's why I haven't sent them yet.
> 
> While we're at it, do you know why `searchAll` have such an awkward interface? I.e. flattened list of all groups of all matches (see second test case).

We are having a discussion on Slack about it :+1:

WGH-:
Once https://github.com/SpiderLabs/secrules-language-tests/pull/4 is merged, I'll update this one to point to different submodule commit.

WGH-:
On second thought, it might be better to just merge this alongside wth re2 branch, once it's completed.

Issue #2008, opened 2019-01-25
Compilation fails with --with-lmdb


mdunc:
**Describe the bug**

When compiling with LMDB, compilation fails with an error regarding Regex

**Logs and dumps**

Output of:
 Compilation error:
```
collection/backend/lmdb.cc: In member function 'virtual void modsecurity::collection::backend::LMDB::resolveRegularExpression(const string&, std::vector<const modsecurity::VariableValue*>*, modsecurity::Variables::KeyExclusions&)':
collection/backend/lmdb.cc:541:38: error: use of deleted function 'modsecurity::Utils::Regex::Regex(const modsecurity::Utils::Regex&)'
     Utils::Regex r = Utils::Regex(var);
                                      ^
In file included from ../src/variables/variable.h:28:0,
                 from ../src/collection/backend/lmdb.h:39,
                 from collection/backend/lmdb.cc:17:
../src/utils/regex.h:57:5: note: declared here
     Regex(const Regex&) = delete;
     ^
Makefile:2608: recipe for target 'collection/backend/libmodsecurity_la-lmdb.lo' failed
make[3]: *** [collection/backend/libmodsecurity_la-lmdb.lo] Error 1
make[3]: Leaving directory '/build/ModSecurity/src'
Makefile:3196: recipe for target 'all-recursive' failed
make[2]: *** [all-recursive] Error 1
make[2]: Leaving directory '/build/ModSecurity/src'
Makefile:1009: recipe for target 'all' failed
make[1]: *** [all] Error 2
make[1]: Leaving directory '/build/ModSecurity/src'
Makefile:1014: recipe for target 'all-recursive' failed
make: *** [all-recursive] Error 1
```

**To Reproduce**

Easily reproducible in a fresh Ubuntu 16.04 Docker container, but I imagine other distros are also affected.

`docker run --rm -ti ubuntu:16.04 bash`

```
#install build dependencies
apt update
apt -y install build-essential git libpcre3-dev libltdl-dev libtool zlib1g-dev autoconf
mkdir /build
cd /build

#compile and install lmdb
git clone https://github.com/LMDB/lmdb.git
cd lmdb/libraries/liblmdb
make && make install && ldconfig

#compile and install ModSecurity
cd /build
git clone https://github.com/SpiderLabs/ModSecurity.git
cd ModSecurity
git checkout v3/master
git submodule init
git submodule update
./build.sh
./configure --with-lmdb
make && make install
```

**Expected behavior**

[...]

PR #2011, opened 2019-01-27
Fix LMDB compilation

WGH-:
See #2008 

When I re-enabled LMDB testing in Travis, I also discovered a regression test that fails when LMDB is enabled, and fixed it as well. The fix is ugly, but so is the LMDB backend code that stores data in hardcoded location relative to the current working dir `¯\_(ツ)_/¯`.


zimmerle:
Thank you for your contribution. We don't encourage users to use LMDB. It is there for advanced users.

Issue #2013, opened 2019-01-31
Build commands replaced with pkg-config in yocto project bitbake builds


chuckwolber:
When building modsecurity v3 in a (Yocto) bitbake recipe for embedded environments, the builds can fail because xml2-config and pcre-config are deprecated in favor of pkg-config.

I am working on a fix to replace the xml2-config and pcre-config references in the m4 macro files with pkg-config. I should have a pull request shortly.

victorhora:
Let us know if you need support with a pull request @chuckwolber. Thanks :)

chuckwolber:
Thanks, and sorry for the delay. I will try to get it fixed shortly.

zimmerle:
Hi, @chuckwolber 

Any progress on this?

PR #2014, opened 2019-01-31
Replace xml2-config, pcre-config with pkg-config.

chuckwolber:
Yocto deprecates xml2-config and pcre-config for pkg-config. As pkg-config is
commonly found (and is used in other parts of the modsecurity v3), this change
should be compatible with existing build environments.

This addresses issue #2013 .

### DIFF
--- a/build/libxml.m4
+++ b/build/libxml.m4
@@ -1,10 +1,11 @@
-dnl Check for LIBXML2 Libraries
-dnl CHECK_LIBXML2(ACTION-IF-FOUND [, ACTION-IF-NOT-FOUND])
+heck for LIBXML2 Libraries
 dnl Sets:
+dnl  LIBXML2_VERSION
 dnl  LIBXML2_CFLAGS
-dnl  LIBXML2_LIBS
+dnl  LIBXML2_CPPFLAGS
+dnl  LIBXML2_LDADD
+dnl  LIBXML2_LDFLAGS
 
-LIBXML2_CONFIG=""
 LIBXML2_VERSION=""
 LIBXML2_CFLAGS=""
 LIBXML2_CPPFLAGS=""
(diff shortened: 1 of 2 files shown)


chuckwolber:
Looks like there are some test failures. ```make check``` seems to go through the motions in my own environment, but does not appear to run any actual tests:

```
============================================================================
Testsuite summary for modsecurity 3.0
============================================================================
# TOTAL: 0
# PASS:  0
# SKIP:  0
# XFAIL: 0
# FAIL:  0
# XPASS: 0
# ERROR: 0
============================================================================
```

I suspect I am missing a few environment tools, but I have not dug too deeply into the details. In any case, I am happy to update the patch if I can get some guidance on why the tests are failing. I also do not spend much time working with autotools, so it is definitely possible that I missed something obvious.

zimmerle:
Hi @chuckwolber,

There is this flag that is manually added called: "-DWITH_LIBXML2". This flag tells ModSecurity to use the libxml during the compilation. It seems like you are missing this flag in your patch.

About your tests not being executed, it is likely that you are missing the yajl during your compilation/configuration. Without JSON ModSecurity is not able to run any test.

chuckwolber:
Updated patch to include "-DWITH_LIBXML2" and it is still failing several test cases. Any guidance you have would be appreciated.

zimmerle:
Hi @chuckwolber,

It seems that ```-D WITH_LIBXML2``` was put in the wrong place. Look at [here](
https://travis-ci.org/SpiderLabs/ModSecurity/jobs/505532935#L752). Also, all the test that was failing is due to the absence of XML. By fixing that XML, is likely to have all test passing.

zimmerle:
I am closing this as the review comments are still open for a while now. Feel free to re-open whenever it is possible to work on it again.

PR #2016, opened 2019-02-01
Fixed data collecting in multipart parsing

airween:
* the FILES_NAMES collection was empty after the end of multipart parsing, fixed it
* added regression test case


airween:
Looks like the existing test cases were wrong, checked same variable values. The second commit (https://github.com/SpiderLabs/ModSecurity/pull/2016/commits/7c6c610e2d6470b9b4eb8c9d768d710c6f55df0d) fixes them, now all build completed.

zimmerle:
Merged! Thanks!

PR #2017, opened 2019-02-03
Fixed validateByteRange parsing method

airween:
There is a bug in @validateByRange operator parsing: if the argument is something like this: "32,34,38,42-59,61,65-90,95,97-122" (see https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/ab24a20faf28156f0495b0c07f2ff37860a3defe/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L1476-L1493 rule), then the 3rd block will evaluated as "38,42-59", not "38" _and_ "42-59".

Also the test file for regression test had been added.


zimmerle:
Merged! Thanks!

PR #2018, opened 2019-02-05
V3/collection re fix

airween:
Looks like the Regex::searchAll() method is not case insensitive (InMemoryPerProcess::resolveSingleMatch() needs that). This bug detected with help of CRS regression test [920450](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.1/dev/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920450.yaml).

Also fix (temporary) the 'ovector' limit bug, which triggered when the regex contains too much brackets, eg. in [CRS rule 942130](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/ab24a20faf28156f0495b0c07f2ff37860a3defe/rules/REQUEST-942-APPLICATION-ATTACK-SQLI.conf#L575).

### DIFF
--- a/src/utils/regex.h
+++ b/src/utils/regex.h
@@ -50,8 +50,7 @@ class SMatch {
 
 class Regex {
  public:
-    explicit Regex(const std::string& pattern_);
-    explicit Regex(const std::string& pattern_, int flags);
+    explicit Regex(const std::string& pattern_, bool caseSensitive = false);
     ~Regex();
 
     // m_pc and m_pce can't be easily copied
--- a/src/utils/regex.h
+++ b/src/utils/regex.h
@@ -27,7 +27,7 @@
 namespace modsecurity {
 namespace Utils {
 
-#define OVECCOUNT 30
+#define OVECCOUNT 50
 
 class SMatch {
  public:
@@ -51,6 +51,7 @@ class SMatch {
 class Regex {
  public:
     explicit Regex(const std::string& pattern_);
+    explicit Regex(const std::string& pattern_, int flags);
     ~Regex();
 
     // m_pc and m_pce can't be easily copied
--- a/src/utils/regex.h
+++ b/src/utils/regex.h
@@ -27,7 +27,7 @@
 namespace modsecurity {
 namespace Utils {
 
-#define OVECCOUNT 50
+#define OVECCOUNT 900
 
 class SMatch {
  public:
--- a/src/collection/backend/in_memory-per_process.cc
+++ b/src/collection/backend/in_memory-per_process.cc
@@ -134,7 +134,7 @@ void InMemoryPerProcess::resolveRegularExpression(const std::string& var,
     //std::string name = std::string(var, var.find(":") + 2,
     //    var.size() - var.find(":") - 3);
     //size_t keySize = col.size();
-    Utils::Regex r(var, PCRE_CASELESS);
+    Utils::Regex r(var, true);
 
     for (const auto& x : *this) {
         //if (x.first.size() <= keySize + 1) {
--- a/src/collection/backend/in_memory-per_process.cc
+++ b/src/collection/backend/in_memory-per_process.cc
@@ -134,7 +134,7 @@ void InMemoryPerProcess::resolveRegularExpression(const std::string& var,
     //std::string name = std::string(var, var.find(":") + 2,
     //    var.size() - var.find(":") - 3);
     //size_t keySize = col.size();
-    Utils::Regex r(var);
+    Utils::Regex r(var, PCRE_CASELESS);
 
     for (const auto& x : *this) {
         //if (x.first.size() <= keySize + 1) {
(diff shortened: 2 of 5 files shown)


airween:
Could somebody re-run the CI build? Looks like it was [failed](https://api.travis-ci.org/v3/job/490059519/log.txt) (by Travis):
```
InstalledDir: /Applications/Xcode-9.4.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin
before_script.1
0.06s$ echo $TRAVIS_OS_NAME
osx
$ [ "$TRAVIS_OS_NAME" != osx ] || brew update
==> Downloading https://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.3.7.mavericks.bottle.tar.gz
######################################################################## 100.0%
==> Pouring portable-ruby-2.3.7.mavericks.bottle.tar.gz
==> Homebrew is run entirely by unpaid volunteers. Please consider donating:
  https://github.com/Homebrew/brew#donations


No output has been received in the last 10m0s
```

Thanks :)

victorhora:
hey @airween, I've triggered a rebuild on the failed build. We should have the results in a moment.

Thanks for your contribution, me or @zimmerle will have a chance to review it soon :)

zimmerle:
Merged ea937cf572dcf9ef5d964fa87dd6aca831d24d77 and 640daa49025a59aee039dfda0a1f34f137415f32. I do have something that I want to discuss about 3334d2a6b26d560ad6d8b820ee6d3fb56e327a75 and 4e1b99ad90a0cfa07bf856984c2c657e12fa2cd2. 

The regular expression implementation is self-contained in the **regex** class. It means that this is the only class which effectively depends on _libpcre_*. Within this patch, the definition _PCRE_CASELESS_ will be spread among other classes making them also dependent on pcre. As of now, it won't be a problem, but, as we aiming to support different regular expression backends, we better avoid this scenario. I would rather prefer to have a new signature to the method -- considering the _PCRE_CASELESS_ flag only inside the regex class or even a modificator on the regular expression string. 


* VerifyCC it is an exception and considered to be a bug to be fixed.

WGH-:
You can make a regular expression case insensitive by prefixing it with `(?i)`. It seems to be supported almost everywhere (I checked libpcre, Python re, RE2, Hyperscan).

airween:
Hi @WGH-, thanks, I know it - but the problem is not that.

The problem is, the comparing variables on collections (in-memory and LMDB) *must be* case *insensitive*.

This condition is not met.
See [this PR](https://github.com/SpiderLabs/ModSecurity/pull/1810), and the documentation.

BTW, the ModSec2 follows this way... so I think we have to align the ModSec3 too.

src/utils/regex.h
```diff
@@ -50,8 +50,7 @@ class SMatch {
 
 class Regex {
  public:
-    explicit Regex(const std::string& pattern_);
-    explicit Regex(const std::string& pattern_, int flags);
+    explicit Regex(const std::string& pattern_, bool caseSensitive = false);
```
zimmerle, line 53: You may not need this to be explicit anylonger.

zimmerle (changes requested):

zimmerle:
@airween as commented by @WGH- the regular expression could be handled this with the ignore case "modificator". In any case, your pull request is currently holding four commits:

- 80409045125387b5a5e1f405855dbf03d245d30b - Removed pcre dependency from outside of regex class
- 3334d2a6b26d560ad6d8b820ee6d3fb56e327a75 - Regex search in collection data case insensitive fix
- 4e1b99ad90a0cfa07bf856984c2c657e12fa2cd2 - Fixed regex case insensitive search bug in LMDB
- 640daa49025a59aee039dfda0a1f34f137415f32 - Added test cases

In 80409045125387b5a5e1f405855dbf03d245d30b you have fixed a problem added in 3334d2a6b26d560ad6d8b820ee6d3fb56e327a75 and  3334d2a6b26d560ad6d8b820ee6d3fb56e327a75. Do you mind to change 3334d2a6b26d560ad6d8b820ee6d3fb56e327a75 and 3334d2a6b26d560ad6d8b820ee6d3fb56e327a75 to actually incorporate 80409045125387b5a5e1f405855dbf03d245d30b ?

airween:
The https://github.com/SpiderLabs/ModSecurity/commit/80409045125387b5a5e1f405855dbf03d245d30b should ignore. Do I revoke it?

I know what @WGH- said and your idea, but the CRS contains regex's without modifications, and modsec2 works with it. I'ld like to follow that way, tha's why I sent this PR.

The other solution would be that all occurrences of necessary regex's need to add the modifier, and works of libmodsec3 should differ from ModSec2.

airween:
Closed this PR, and re-opened a new one [here: #2107 ](https://github.com/SpiderLabs/ModSecurity/pull/2107)

PR #2024, opened 2019-02-11
Allow empty anchored variable to use

airween:
The problem is that when a SecRule contains an negated operator (eg. "!@endsWith .pdf"), and the variable is empty (eg. REQUEST_BASENAME, and the query is a simple "GET /"), then the result of negated evaluate of empty variable will TRUE. Then the rule interpreted as the GET / _IS_ endsWith .pdf, so it will generate a false positive match.

I don't see the reason why didn't allowed the empty variable previously, but this patch fix this bug above, and all other test result keep as PASSED.

See the rule-920200.json regression test file, which inspired the CRS id 920200 and its test.


zimmerle:
Merged! Thanks!

PR #2046, opened 2019-03-13
segfault with assigned user id on OpenShift

arminabf:
When running httpd on OpenShift, by default the server instance will run with an assigned user ID not appearing in the UNIX password file.

There are several places in the code of mod_security where the user name is tried to retrieved by use of apr_uid_name_get(). As the function can not find a corresponding user name it returns with an error (and argument pointer being NULL). Due to missing error handling constant segmentation faults are faced on OpenShift.

As a fallback on OpenShift (and probably other container platforms), the user id should be used instead of the user name.

### DIFF
--- a/apache2/msc_logging.c
+++ b/apache2/msc_logging.c
@@ -234,16 +234,20 @@ static char *construct_auditlog_filename(apr_pool_t *mp, const char *uniqueid) {
      * This is required for mpm-itk & mod_ruid2, though should be harmless for other implementations 
      * It also changes the return statement.
      */
-    char *username;
+    char *userinfo;
+    apr_status_t rc;
     apr_uid_t uid;
     apr_gid_t gid;
     apr_uid_current(&uid, &gid, mp);
-    apr_uid_name_get(&username, uid, mp);
+    rc = apr_uid_name_get(&userinfo, uid, mp);
+    if (rc != APR_SUCCESS) {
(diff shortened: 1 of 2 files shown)


marcstern:
Better to have a centralized function for this:
```
char* get_username(apr_pool_t* mp) {
 char* username;
 apr_uid_t uid;
 apr_gid_t gid;
 int rc = apr_uid_current(&uid, &gid, mp);
 if (rc != APR_SUCCESS) return "apache";
 rc = apr_uid_name_get(&username, uid, mp);
 if (rc != APR_SUCCESS) return "apache";
 return username;
}
```

FedericoHeichou:
Why isn't this merged? Sometimes segmentation fault won't happens and it could lead to other major vulnerabilities.  
In `apache2/re_variables.c:2597` the rc is checked, why it isn't in these other 2 files?

martinhsv:
There are two slightly differing proposals for this issue if apr_uid_name_get() fails:
 - use the numeric value from apr_uid_current
 - use a fixed string such as "apache" (in the related issue)

There are some things I like about the first proposal, but I'm wondering if having an all-digit default could be confusing or obscure for some users.

marcstern:
Using the numeric value from apr_uid_current is OK (and maybe even better).
Both solutions are better than the current bogus behaviour anyway.

martinhsv:
Thanks all, for the contributions on this matter.

Issue #2000, opened 2019-01-11
SecAuditLogParts "H" not being written in some cases


victorhora:
**Describe the bug**

Our friend @defanator have recently reported a strange behaviour where SecAuditLogParts "H" is not being written to the audit_log in some cases. Namely:

```SecRule ARGS:testparam "@contains test" "id:1234,deny,log,status:403"```
- this rule does not produce anything in H section

```SecRule ARGS:testparam "@contains test" "id:1234,block,log,status:403"```
- while this one does

The difference being block action vs deny action.

**Logs and dumps**

```
---C788bS9S---A--
[11/Jan/2019:15:08:25 +0000] 154721930524.840108 127.0.0.1 48688 127.0.0.1 80
---C788bS9S---B--
GET /?testparam=deny123 HTTP/1.1
Host: localhost
User-Agent: curl/7.47.0
Accept: */*

---C788bS9S---D--

---C788bS9S---E--
<html>\x0d\x0a<head><title>403 Forbidden</title></head>\x0d\x0a<body>\x0d\x0a<center><h1>403 Forbidden</h1></center>\x0d\x0a<hr><center>nginx/1.15.7</center>\x0d\x0a</body>\x0d\x0a</html>\x0d\x0a

---C788bS9S---F--
HTTP/1.1 403
Server: nginx/1.15.7
Date: Fri, 11 Jan 2019 15:08:25 GMT
Content-Length: 153
Content-Type: text/html
Connection: keep-alive

---C788bS9S---H--

---C788bS9S---I--

---C788bS9S---J--

---C788bS9S---Z--

---PuQ3VB7M---A--
[11/Jan/2019:15:08:29 +0000] 154721930978.657786 127.0.0.1 33304 127.0.0.1 8080
---PuQ3VB7M---B--
GET /?testparam=block123 HTTP/1.0
Host: 127.0.0.1:8080
Connection: close
User-Agent: curl/7.47.0
Accept: */*

---PuQ3VB7M---D--

---PuQ3VB7M---F--
HTTP/1.0 200
Server: nginx/1.15.7
Date: Fri, 11 Jan 2019 15:08:29 GMT
Content-Length: 3
Content-Type: application/octet-stream
Connection: close

---PuQ3VB7M---H--
[...]

PR #2048, opened 2019-03-15
Make sure m_rulesMessages is filled after successfull match

victorhora:
This fixes issue #2000.

The problem seems to be that the conditions to push ruleMessage below is not met when only [deny](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#deny) action is used:

https://github.com/SpiderLabs/ModSecurity/blob/6d5198b1a60992416c9c8f1d77754e8e5edba996/src/rule.cc#L815-L822

Then when the for loop to append m_rulesMessages to the audit_log string stream is called, the content of this std::list seems to be empty:

https://github.com/SpiderLabs/ModSecurity/blob/6d5198b1a60992416c9c8f1d77754e8e5edba996/src/transaction.cc#L1486-L1489

### DIFF
--- a/src/rule.cc
+++ b/src/rule.cc
@@ -812,7 +812,7 @@ end_exec:
             trans->serverLog(ruleMessage);
 	}
     }
-    else if (m_containsStaticBlockAction && !m_containsMultiMatchAction) {
+    else if (!m_containsMultiMatchAction) {
         /* warn */
         trans->m_rulesMessages.push_back(*ruleMessage);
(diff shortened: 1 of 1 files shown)


defanator:
@victorhora I can confirm that this one fixes an issue reported in #2000.

zimmerle:
merged at: d4dc3dbf2a647942b9b315a413a99cacae90a484

Issue #890, opened 2015-05-23
Spontaneously high CPU usage


celesteking:
This bug is hard to replicate, but I'll try to describe it here.
We've activated modsec on our servers, but with some we're noticing extremely high cpu usage. It manifests itself after running for some time in detection mode. 
We have mlogc logging enabled. 

It might be caused by graceful restart, although I'm not sure -- no way to replicate reliably. 

Here's what happens with a process that's consuming 100% CPU:

```
(gdb) bt
#0  0x00002b36cebad969 in run_child_cleanups () from /usr/local/apache/lib/libapr-1.so.0
#1  0x00002b36cebad9af in cleanup_pool_for_exec () from /usr/local/apache/lib/libapr-1.so.0
#2  0x00002b36cebad9c6 in cleanup_pool_for_exec () from /usr/local/apache/lib/libapr-1.so.0
#3  0x00002b36cebad9c6 in cleanup_pool_for_exec () from /usr/local/apache/lib/libapr-1.so.0
#4  0x00002b36cebad9eb in apr_pool_cleanup_for_exec () from /usr/local/apache/lib/libapr-1.so.0
#5  0x00002b36cebba4e9 in apr_proc_create () from /usr/local/apache/lib/libapr-1.so.0
#6  0x00002b36cf1e86c2 in suphp_script_handler (r=0x1122ceb8) at mod_suphp.c:953
#7  0x00002b36cf1e8ef9 in suphp_handler (r=0x1122ceb8) at mod_suphp.c:569
#8  0x000000000044ac13 in ap_run_handler ()
#9  0x000000000044b4dc in ap_invoke_handler ()
#10 0x00000000004b9ddd in ap_internal_redirect ()
#11 0x00000000004e1f11 in handler_redirect ()
#12 0x000000000044ac13 in ap_run_handler ()
#13 0x000000000044b4dc in ap_invoke_handler ()
#14 0x00000000004b9ddd in ap_internal_redirect ()
#15 0x00000000004b90d3 in ap_die ()
#16 0x00000000004b92ca in ap_process_request ()
#17 0x00000000004b5abd in ap_process_http_connection ()
#18 0x00000000004546cf in ap_run_process_connection ()
#19 0x0000000000454b33 in ap_process_connection ()
#20 0x00000000004e3157 in process_socket ()
#21 0x00000000004e3a82 in worker_thread ()
[...]

PR #2049, opened 2019-03-19
Issue 890 - ModSecurity corrupts the global pool's cleanups linked list with threaded MPMs.

lightsey:
I tested the first commit fixing the re_precomp rules ( msre_op_validateHash_execute and msre_op_rx_execute ) using the example configuration I provided in issue #890.

The second commit fixing the other usage of rule->ruleset->mp in functions that are run during HTTP requests hasn't gone through any real testing. Nothing stands out to me in the code as suggesting that these should be using storage pools that are shared between all the threads in the process, but I'm not that very familiar with ModSecurity's codebase and could be mistaken.

### DIFF
--- a/apache2/re_operators.c
+++ b/apache2/re_operators.c
@@ -784,10 +784,10 @@ static int msre_op_validateHash_execute(modsec_rec *msr, msre_rule *rule, msre_v
                 msr_log(msr, 6, "Escaping pattern [%s]",pattern);
             }
 
-            regex = msc_pregcomp_ex(rule->ruleset->mp, pattern, PCRE_DOTALL | PCRE_DOLLAR_ENDONLY, &errptr, 
+            regex = msc_pregcomp_ex(msr->mp, pattern, PCRE_DOTALL | PCRE_DOLLAR_ENDONLY, &errptr,
                     &erroffset, msc_pcre_match_limit, msc_pcre_match_limit_recursion);
             if (regex == NULL) {
-                *error_msg = apr_psprintf(rule->ruleset->mp, "Error compiling pattern (offset %d): %s",
+                *error_msg = apr_psprintf(msr->mp, "Error compiling pattern (offset %d): %s",
                         erroffset, errptr);
                 return 0;
             }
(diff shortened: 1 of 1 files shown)


markblackman:
This is a perfect fit for CGI problem we are seeing in #2101  and we will test your PR

lightsey:
cPanel has shipped this patch to customers for several months now without any apparent problems.

https://github.com/CpanelInc/mod_security2

marcstern:
> 
> 
> cPanel has shipped this patch to customers for several months now without any apparent problems.
> 
> https://github.com/CpanelInc/mod_security2

Apparently not. The file in the source doesn't have that patch (https://github.com/CpanelInc/mod_security2/blob/master/SOURCES/modsecurity-2.9.3.tar.gz)

markblackman:
Although I updated #2101 at the time, last year, I will also point out here that this patch fixed our thread problem. Mod_security 2.x was not written with thread safety in mind and there was at least too much unsafe sharing of global pools between threads in an apache 2.4 event MPM.

lightsey:
> > cPanel has shipped this patch to customers for several months now without any apparent problems.
> > https://github.com/CpanelInc/mod_security2
> 
> Apparently not. The file in the source doesn't have that patch (https://github.com/CpanelInc/mod_security2/blob/master/SOURCES/modsecurity-2.9.3.tar.gz)

The patch is applied while building the RPM. The tar.gz file in this repo should be a pristine copy of the original modsecurity tarball.

https://github.com/CpanelInc/mod_security2/blob/master/SOURCES/0003-Store-temporaries-in-the-request-pool-for-regexes-co.patch

https://github.com/CpanelInc/mod_security2/blob/ecbb7b9802f628a89610b2dcce23961161c2fc31/SPECS/ea-apache24-mod_security2.spec#L56

marcstern:
Indeed.
So, running for 11 months in prod on CPanel should be considered as robust, no?

marcstern:

zimmerle:
Merged! Thanks @lightsey

PR #2057, opened 2019-04-03
Fix variables output in debug logs

jleproust:
Just a little fix for variable lists in debug logs. Without this, printing a `Variables` object in the debug log actually only printed one variable of the list, not the whole pipe-separated list.


zimmerle:
Merged! Thanks!

PR #2058, opened 2019-04-03
Store values extracted from json to ARGS_POST

jleproust:
Variables extracted from the body of requests in multipart or urlencoded contents are stored in both the ARGS/ARGS_NAMES and ARGS_POST/ARGS_POST_NAMES, so we can choose to match or exclude them with ARGS_GET/ARGS_POST.

However, when the body of the request is JSON, its content can be optionally extracted to variables only in the ARGS and ARGS_NAMES collections. The ARGS_POST/ARGS_POST_NAMES collections stay empty.

My change stores them into both like multipart/urlencoded bodies, for consistency and simpler rules.

The only side effect is a change in the debug log line where the source is used, JSON is replaced with POST.

### DIFF
--- a/src/request_body_processor/json.cc
+++ b/src/request_body_processor/json.cc
@@ -142,7 +142,7 @@ int JSON::addArgument(const std::string& value) {
     }
 
 
-    m_transaction->addArgument("JSON", path + data, value, 0);
+    m_transaction->addArgument("POST", path + data, value, 0);
 
     return 1;
 }


zimmerle:
@jleproust thank you for the patch.

I understand the rationality purposed; however, it is interesting to consider the particular aspects of a JSON; I would like to highlight the possibility for a JSON to have a catastrophic amount of keys. A rule meant to inspect the ARGS_POST may not take into consideration all those keys. That behavior could lead to an undesired consumption of the server resources.

Considering this use case, a possibility to tell apart what is ARGS_POST and JSON sounds mandatory. Yet a rule that wants to inspect both collections can easily target: "JSON|ARGS_POST".

Please re-open if you think that this decision should be revisited.

jleproust:
Hi @zimmerle ! Thanks for the feedback, let me gather my memories from 2 years ago :sweat_smile: 

I was well aware of the JSON extraction performance issue, I even contributed the patch to mitigate them, long before it was even acknowledged as a DoS vuln. However using JSON as a collection instead of ARGS_POST does not help in all the cases, sometimes you want to match json keys content specifically, and I don't think we could this way. The application I wanted to protect had different arguments in query string and request body, and I wanted to process form POST and JSON content the same, while handling query parameters otherwise. This seemed like a good idea, whereas I don't think your proposition would allow it without duplicating rules (which comes with other performance issues). Especially I don't think CRS matches JSON by default, but I'm not sure after such a long time.

zimmerle:
Let me exemplify how to inspect both collections without duplicating the rules.

#### Rule 1 - Inspecting only ARGS_POST
```
SecRule ARGS_POST "id:1"
```

#### Rule 2 - Inspecting only JSON
```
SecRule JSON "id:2"
```

#### Rule 3 - Inspecting both: ARGS_POST and JSON
```
SecRule ARGS_POST|JSON "id:3"
```

If we merge the purposed patch, all the use cases above will be concentrated in one: inspect both collections simultaneously. In this fashion, we are going to lose some functionality. Not to mention that targets can be [updated by ID](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#SecRuleUpdateTargetById); thus, for some _vhost_ or _paths_, it is possible to add JSON inspection to a previously Rule that was targeting ARGS_POST.

Issue #2019, opened 2019-02-08
nginx + v3: high CPU usage on rule 200001


theMiddleBlue:
I'm not sure if this issue was already reported. Anyway:

What reported here https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1133 seems to be related to the rule 200001 and the `requestBodyProcessor=JSON` on each request with `Content-Type: application/json`.

With a long JSON string in the request body, the rule 200001 takes a very long time to process it, and the nginx worker CPU usage stuck at 100% until it finishes.

**Logs and dumps**
Using an 800 Kb JSON file as request body, and sent with content-type `application/json` the request took 83 seconds:
```
curl -s -H 'Content-Type: application/json' -d @test.json 'http://localhost' -w @curl-format.txt
host localhost upstream: 
    time_namelookup:  0.005550
       time_connect:  0.006125
    time_appconnect:  0.000000
   time_pretransfer:  0.011318
      time_redirect:  0.000000
 time_starttransfer:  0.014967
                    ----------
         time_total:  83.675212
```

Just replacing the content-type value with `application/x-www-form-urlencoded`, the request took just 2 seconds:

```bash
curl -s -H 'Content-Type: application/x-www-form-urlencoded' -d @test.json 'http://localhost' -w @curl-format.txt
host localhost upstream: 
    time_namelookup:  0.005729
       time_connect:  0.006164
    time_appconnect:  0.000000
   time_pretransfer:  0.006292
      time_redirect:  0.000000
 time_starttransfer:  0.012784
                    ----------
         time_total:  2.071719
```

**To Reproduce**
- download [test.json](https://gist.githubusercontent.com/theMiddleBlue/6081d350bb8c2a5ac331c769cbd68bcd/raw/0a5a4f587fb27289b375c09c321f0c258d48bd5c/test.json)
- create curl-format.txt as following
```
    time_namelookup:  %{time_namelookup}\n
       time_connect:  %{time_connect}\n
    time_appconnect:  %{time_appconnect}\n
   time_pretransfer:  %{time_pretransfer}\n
      time_redirect:  %{time_redirect}\n
 time_starttransfer:  %{time_starttransfer}\n
                    ----------\n
         time_total:  %{time_total}\n
```
- send request to nginx with content-type `application/json` and the test.json content as the request body:

```bash
$ curl -v -H 'Content-Type: application/json' -d @test.json 'http://localhost' -w @curl-format.txt
```
**Server:**
[...]

PR #2060, opened 2019-04-04
Mitigate DoS vulnerability by limiting number of extracted arguments

jleproust:
A relatively small json payload (<1MB) can generate up to a hundred thousands keys and values, which can then take ages (>30 minutes) to process with OWASP's CRS 3 for example. The parser is not the issue, no particular rule is the issue, it's just the pure quantity of values.

Obviously, with the nginx connector, this completely locks the worker for the whole duration, so it is a DoS vulnerability.

This is the underlying issue behind #2019 and I guess, https://github.com/SpiderLabs/ModSecurity-nginx/issues/89 too.

That's why I implemented a `SecArgumentsLimit` value, disabled by default, allowing to set a limit to the number of extracted arguments. When the limit is reached, subsequent values are ignored. A limit of 1000 arguments should be enough for most applications, and reduces the maximum processing time to well under a second. Requests with more than 1000 arguments can be matched and rejected if wanted with a rule like `SecRule &ARGS "@gt 1000"`.

In addition, I made the JSON processor cancel the parsing when the limit is reached.


feromax:
Hi there.  Anyone have an ETA on when this will be merged?  Thank you :)

zimmerle:
Closing this in favor of #2234.

Issue #2062, opened 2019-04-06
Add new operator @verifySVNR


Rufus125:
Adding a new operator that detects Austrian social security numbers.

Meaning and calculation:
https://de.wikipedia.org/wiki/Sozialversicherungsnummer#%C3%96sterreich
https://de.wikipedia.org/wiki/Sozialversicherungsnummer#Berechnung
Unfortunately, I could only find links in German.

I have already written the code and will open a PR.

zimmerle:
This was merged already :) thank you.

PR #2063, opened 2019-04-06
New operator verifySVNR that finds Austrian social security numbers

Rufus125:
Relates to #2062 
I also created some tests and a PR, therefore the travis build will fail: https://github.com/SpiderLabs/secrules-language-tests/pull/5


zimmerle:
Hi @Rufus125,

Thank you for your contribution. Instead of merging the requested branch, I am picking commit by commit in order to keep our tree as clean as possible.

For instance, I've just merged bb56288d8511549e07ac4753b233154a1e7713ab into the tree, not actually related to the pull request, yet a valid contribution.

src/operators/verify_svnr.h
```diff
@@ -0,0 +1,55 @@
...
+    /** @ingroup ModSecurity_Operator */
+    explicit VerifySVNR(std::unique_ptr<RunTimeString> param)
+        : Operator("VerifySVNR", std::move(param)) {
+        m_re = new Regex(m_param);
+    }
+
+    ~VerifySVNR() {
+        delete m_re;
+    }
+    bool evaluate(Transaction *transaction, Rule *rule,
```
zimmerle, line 31: You do not need to specify those here. VerifySVNR inherit from Operator which already contains code to do exactly the same -

https://github.com/SpiderLabs/ModSecurity/blob/6d266fae8549ac425ded31692f6d7766ca0bdc72/src/operators/operator.h#L121

It is safe to remove this evaluate declaration. It is better to remove, so we will have a cleaner code in case of a refactoring or major change.

src/operators/verify_svnr.h
```diff
@@ -0,0 +1,55 @@
...
+        const std::string &input) override {
+        return evaluate(transaction, NULL, input, NULL);
+    }
+    bool evaluate(Transaction *transaction,
```
zimmerle, line 35: Same for the evaluate above. No need to co-exist here.

src/operators/verify_svnr.h
```diff
@@ -0,0 +1,55 @@
...
+        const std::string &input) override {
+        return evaluate(transaction, NULL, input);
+    }
+    bool evaluate(Transaction *transaction, Rule *rule,
+        const std::string& input,
+        std::shared_ptr<RuleMessage> ruleMessage) override;
+
+    int convert_to_int(const char c);
```
zimmerle, line 43: Those could be private. Their usage is restricted to the very own class.

test/test-cases/regression/operator-verifysvnr.json
```diff
@@ -0,0 +1,46 @@
+[
```
zimmerle: Thumbs up for the test case :) that really speed up my review. Thank you.

src/operators/verify_svnr.cc
```diff
@@ -0,0 +1,114 @@
...
+    std::list<SMatch> matches;
+    bool is_svnr = false;
+    int i;
+
+    if (m_param.empty()) {
+        return is_svnr;
+    }
+
+    for (i = 0; i < input.size() - 1; i++) {
+        printf ("Decimal i: %d\n", i);
```
zimmerle: Although that is a very useful debug information, it may not a good idea for production I would consider removing it.

src/operators/verify_svnr.cc
```diff
@@ -0,0 +1,114 @@
...
+#include "src/operators/verify_svnr.h"
+
+#include <string>
+
+#include "src/operators/operator.h"
+
+namespace modsecurity {
+namespace operators {
+
+int VerifySVNR::convert_to_int(const char c)
```
zimmerle, line 14: I see that this function is also used in verify cpf. It seems to me that in a further step, we can move it to utils and make both to share the same base code.

Not need to be done now, that could be an optimization.

test/test-cases/regression/operator-verifysvnr.json
```diff
@@ -0,0 +1,46 @@
...
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing Operator :: @verifysvnr (1/2)",
```
zimmerle: This would be 1 of 1.

zimmerle (changes requested):

zimmerle:
Hi @Rufus125,

I've made all the changes that I have suggested on the review and merged the patch. Do you mind to double check to see if I am not missing something?

Thank you!

PR #2073, opened 2019-04-18
Limit more log variables to 200 characters

jleproust:
To avoid generating too long log lines which end up truncated (like in nginx's error log), and missing important bits. Most variables are already truncated to 100-200 characters, but logdata and the request can be huge too.


zimmerle:
Hi @jleproust,

Thank you for your contribution. Sorry about the delay to get it merged. The changes are on v3/master now.

Issue #2091, opened 2019-05-14
ProcessPartial can truncate request bodies in combination with mod_proxy_ajp and mod_wl


rainerjung:
Version: 2.9.3

Symptom: when uploading files to a backend using mod_proxy_ajp or mod_wl, the request body (file contents) can get truncated. The problem is reproducible when using

SecRuleEngine DetectionOnly
SecRequestBodyAccess On
SecRequestBodyLimitAction ProcessPartial

and setting SecRequestBodyNoFilesLimit, SecRequestBodyLimit and SecRequestBodyInMemoryLimit to a value smaller than the file size. In addition I could only easily reproduce the problem when not including any CRS rules. I can provide a full config plus simple test app if needed.

The problem happens when the rquest body limit is reached and partial body processing is active. If bytes are overflowing to disk, then the last read attempt from the mod_security input filter will return 0 bytes and remove the input filter from the filter chain. Following retrievals from the filter chain will correctly return the remaining bytes.

But mod_wl and mod_prox_ajp react on a 0 byte return from the request body read by ending the request body transfer to the backend. For mod_wl the reason is not clear (closed source), but the result is a "503 Service Unavailable" error. For mod_proxy_ajp the reason is, that a zero byte AJP request body packet indicates the end of the body in the AJP protocol, so the backend doesn't pull the remaining parts of the body and instead starts sending the response.

I will create a simple PR. The patch tracks, whether the mod_security input filter actually returns data or not. In case it does not return data and removes itself from the chain, it will delegate to the remaing filter chain instead of just returning APR_SUCCESS.

Regards,

Rainer

zimmerle:
Fixed at #2092.

Issue #1827, opened 2018-07-02
mod_security 2.9.2 appears to truncate POST bodies when using mod_proxy_ajp and mod_jk when using ProcessPartial


markblackman:
We configure our POST body limit as follows

```
  <IfModule mod_security2.c>
    SecRuleEngine DetectionOnly

    SecRequestBodyAccess On
    SecRequestBodyLimitAction ProcessPartial

    SecRequestBodyNoFilesLimit 4096
    SecRequestBodyLimit 4096
    SecRequestBodyInMemoryLimit 4096
  </IfModule>
```

We can then see on the backend that the resulting request is truncated to the SecRequestBodyLimit 4096

zimmerle:
Hi @markblackman,

The configuration  **SecRequestBodyLimit** limits the amount of bytes of the request body to be inspected. That in the expected behavior.

markblackman:
Thanks. Sorry, if it wasn't clear. I am not referring to the inspection behaviour. I am referring to the proxying behaviours. I can completely live with only inspecting 4k bytes out of 8k bytes in an HTTP body and that's what I expect from ProcessPartial. What I can not live with and do not expect, is the 8K body getting truncated to 4k when it's passed to the backend Tomcat servers. Is this well known or well documented behaviour, truncating after inspection?

markblackman:
To be even clearer, with ProcessPartial I expect the entire body to be proxied regardless of how much is inspected. Is this expectation misplaced?

zimmerle:
What is the version of your ModSecurity?

markblackman:
[Mon Jul 02 17:51:54.562328 2018] [:notice] [pid 2380366:tid 140467527632640] ModSecurity for Apache/2.9.2 (http://www.modsecurity.org/) configured.

zimmerle:
Is there any version mismatch in the dependencies?

markblackman:
None that are flagged by modsecurity itself.

```
[Mon Jul 02 18:22:44.359483 2018] [:notice] [pid 2463776:tid 140163929712384] ModSecurity for Apache/2.9.2 (http://www.modsecurity.org/) configured.
[Mon Jul 02 18:22:44.359495 2018] [:notice] [pid 2463776:tid 140163929712384] ModSecurity: APR compiled version="1.6.3"; loaded version="1.6.3"
[...]

Issue #1471, opened 2017-06-26
Multipart: Final boundary missing.


tomsommer:
I have a customer uploading a 200mb mp4 file through AJAX in Drupal.

With mod_security2 enabled, this results in a stopped 57% upload. The following is visible in the debug log:
```
[26/Jun/2017:09:55:17 +0200] [XXXX][1] Request body (Content-Length) is larger than the configured limit (134217728).
[26/Jun/2017:09:55:17 +0200] [XXXX][4] Second phase starting (dcfg 6d31318).
[26/Jun/2017:09:55:17 +0200] [XXXX][1] Request body (Content-Length) is larger than the configured limit (134217728).
[26/Jun/2017:09:55:17 +0200] [XXXX][4] Input filter: Reading request body.

....

[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] Input filter: Bucket type HEAP contains 8000 bytes.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] Input filter: Bucket type HEAP contains 192 bytes.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] Input filter: Bucket type HEAP contains 6408 bytes.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] Input filter: Bucket type HEAP contains 7300 bytes.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] Input filter: Bucket type HEAP contains 7300 bytes.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][4] Multipart parsing error: Multipart: Final boundary missing.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][4] Input filter: Completed receiving request body (length 134219089).
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][4] Starting phase REQUEST_BODY.
[26/Jun/2017:09:55:18 +0200] [XXXXlZepdfwwbDzafVgPaKmpJnA][9] This phase consists of 0 rule(s).
[...]

PR #2092, opened 2019-05-14
When the input filter finishes, check whether we returned data

rainerjung:
during the last read and if not, delegate to the remaining filter
chain.

Without that, ProcessPartial for the request body breaks forwarding
of uploaded files using mod_proxy_ajp and mod_wl.

See issue #2091.


zimmerle:
Related to:

- #1600 and #1471 - @tomsommer ([1](https://github.com/SpiderLabs/ModSecurity/issues/2093#issuecomment-493396588))
- #1827 - @markblackman ([2](https://github.com/SpiderLabs/ModSecurity/issues/2093#issuecomment-494875382))

zimmerle:
Merged! Thank you.

Issue #2093, opened 2019-05-14
Don't run request body completion checks when ProcessPartial and the input filter has't seen all of the body


rainerjung:
Version: 2.9.3

Only run completion checks for request bodies, if we have
actually seen all of the body. In case ProcessPartial is configured
and our input filter only processed part of the request body, checking
for a complete request body makes no sense.

I will provide a pull request.

dune73:
Thank you Rainer.

tomsommer:
Relates to https://github.com/SpiderLabs/ModSecurity/issues/1600 https://github.com/SpiderLabs/ModSecurity/issues/1471

markblackman:
Also relates to #1827

zimmerle:
> Relates to #1600 #1471

> Also relates to #1827

All these three issues seem to be a consequence of ModSecurity been holding the filter data whenever the limit threshold is reached. However, in this very issue @rainerjung is not treating this specific problem. He did treat on #2092, therefore I am making a comment there mention the relation with the listed issues.

zimmerle:
Hi @rainerjung,

Thank you for your patch.

I do have some concerns that, if you don't mind, I would like to discuss and understand your point of view.

**SecRequestBodyLimitAction** can assume two different values: _Reject_ and _ProcessPartial_. When it is marked to _Reject_ I understand that we may not need to inspect the request body contents. However, in the case of _ProcessPartial_ we need to go over the **request data** no matter what. Because the user may have configured ModSecurity to validate a JSON, XML, or multipart considering the limit values. In other words: He may be looking for a JSON error to drop a request (or any other action). If we remove that check, we won't be processing the content partially, rather ignoring the content. Therefore not generating the error which is expected by the end user.

What do you think about that?

zimmerle:
@rainerjung: Waiting for your considerations here.

@tomsommer && @markblackman: in case you are interested, you are invited to discuss as well.

@dune73: it seems that you have agreed with @rainerjung's original idea, what is your opinion considering my comments?

markblackman:
In my case I don't currently have the requirement to validate a full request body even in the processpartial case, so I am content with the implicit ignore, but I can imagine others will not be. You could also just print a warning to that effect ("ProcessPartial limit exceeded, validation not done") in the logs.  

We are already using Rainer's patch to very positive effect and these will go into production systems tomorrow, in fact, so the question is academic for me. We desperately needed these fixes for our use case and we have no interest in validation cases yet.

[...]

PR #2094, opened 2019-05-14
Only run completion checks for request bodies, if we have

rainerjung:
actually seen all of the body. In case ProcessPartial is configured
and our input filter only processed part of the request, checking
for a complete request body makes no sense.

See issue #2093.


zimmerle:
There is a discussion on issue #2093 about this pull request.

Issue #2099, opened 2019-05-20
CRS and v3 incompatibility due to ruleRemoveByTag


theMiddleBlue:
Hi all,

it seems that we've a compatibility problem with CRS and v3 due to missing `ctl:ruleRemoveTargetByTag` and `ctl:ruleRemoveByTag`. Can you confirm the following two issues?

- https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1418
- https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1419

If yes, is there a plan on when they will be implemented on v3?

thank you!

theMiddleBlue:
<del>additional bit: it seems that `ctl:ruleRemoveById` works only if the first argument is an id and not a range, example:</del>

- <del>`ctl:ruleRemoveById=1234` work</del>
- <del>`ctl:ruleRemoveById=900000-999999` doesn't work (used in some CRS exclusion rule files)</del>

thanks.

theMiddleBlue:
all problems seems to be related just to `ctl:ruleRemoveTargetByTag` and `ctl:ruleRemoveByTag` (thanks @airween). I'm trying to understand why some of my exclusion rules that use ranges in `ctl:ruleRemoveById` doesn't work as expected on v3.

airween:
Okay, the ruleRemoveById issue had been cleared :).

Here is a possible solution for ruleRemoveByTag:
https://github.com/airween/ModSecurity/tree/v3/issue2099
https://github.com/SpiderLabs/ModSecurity/compare/v3/master...airween:v3/issue2099

Please check that if works as well or you found any other problem (only with ruleRemoveByTag ctl action).

I'll continue with the ruleRemoveTargetByTag investigation.

theMiddleBlue:
Thank you @airween 

after testing your patch it seems that all works like a charm! Maybe I've tested `ctl: ruleRemoveTargetByTag` with a buggy rule because it works too!

thank you so much for your support, as always!

airween:
PR sent: https://github.com/SpiderLabs/ModSecurity/pull/2102

zimmerle:
Hi @theMiddleBlue,

The issue within **ctl:ruleRemoveByTag** is indeed confirmed. There was a missing check on the rules execution loop. @airween provided a pull request on #2102, which is under revision; Waiting for @airween appreciation.

[...]

PR #2102, opened 2019-05-23
Adds missing check for runtime ctl:ruleRemoveByTag (Issue #2099)

airween:
This PR fixed the remained bug from issue #2099:

Added some test cases to prevent the similar problems.

### DIFF
--- a/src/rules.cc
+++ b/src/rules.cc
@@ -239,17 +239,19 @@ int Rules::evaluate(int phase, Transaction *t) {
                 }
             }
 
-            for (auto &z : t->m_ruleRemoveByTag) {
-                if (rule->containsTag(z, t) == true) {
-                    ms_dbg_a(t, 9, "Skipped rule id '" \
-                        + std::to_string(rule->m_ruleId) \
-                        + "'. Skipped due to a ruleRemoveByTag action.");
-                    remove_rule = true;
-                    break;
+            if (t->m_ruleRemoveByTag.empty() == false) {
+                for (auto &z : t->m_ruleRemoveByTag) {
+                    if (rule->containsTag(z, t) == true) {
+                        ms_dbg_a(t, 9, "Skipped rule id '" \
+                            + std::to_string(rule->m_ruleId) \
+                            + "'. Skipped due to a ruleRemoveByTag action.");
+                        remove_rule = true;
+                        break;
+                    }
                 }
-            }
-            if (remove_rule) {
-                continue;
+                if (remove_rule) {
+                    continue;
+                }
             }
 
             rule->evaluate(t, NULL);
--- a/src/rules.cc
+++ b/src/rules.cc
@@ -248,6 +248,9 @@ int Rules::evaluate(int phase, Transaction *t) {
                     break;
                 }
             }
+            if (remove_rule) {
+                continue;
+            }
 
             rule->evaluate(t, NULL);
             if (t->m_it.disruptive == true) {
--- a/test/test-cases/regression/issue-2099.json
+++ b/test/test-cases/regression/issue-2099.json
@@ -1,0 +1,195 @@
+[
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing ctl:ruleRemoveById - issue 2099",
+    "expected":{
+      "http_code":200
+    },
+    "client":{
+      "ip":"200.249.12.31",
+      "port":123
+    },
+    "request":{
+      "headers":{
+        "Host":"localhost",
+        "User-Agent":"curl/7.38.0",
+        "Accept":"*/*"
+      },
+      "uri":"/remote.php/webdav?bar=foo",
+      "method":"GET",
+      "body": ""
+    },
+    "server":{
+      "ip":"200.249.12.31",
+      "port":80
+    },
+    "rules":[
+        "SecRuleEngine On",
+        "SecRequestBodyAccess On",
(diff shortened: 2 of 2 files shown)


src/rules.cc
```diff
@@ -248,6 +248,9 @@ int Rules::evaluate(int phase, Transaction *t) {
                     break;
                 }
             }
+            if (remove_rule) {
```
zimmerle: Hi @airween,

I am assuming that you have copy && paste the **remove_rule** check from the block just above. Indeed, that was missing. However you may have pasted in the wrong place, that may lead to unnecessary loops. Can you make it exactly in the same fashion as the others checks?

zimmerle (commented):
Hi @airween, 

I have started a review in the code.

Question: Why you have sent this patch against that branch? As already explained to you we are working on 3.1. It seems like your patch is compatible only with the master branch.

zimmerle:
_General comments_:

- Try to give significative names to pull requests, so it can be searchable by other users.
```
A good title could be: Adds missing check for runtime SecRuleRemoveByTag (Issue #2099)
```

- Whenever you make a reference to an issue, use the hash marking otherwise Github won't notice the reference.
```
Instead of 2099 use #2099
```

src/rules.cc
airween: Hi @zimmerle, sorry, I don't know what do you think about - why is this a wrong place, and which loops could occures?

Do you mean that it needs a plus condition before the for loop, as the previous variables?

airween (commented):
> Hi @airween,
> 
> I have started a review in the code.
> 
> Question: Why you have sent this patch against that branch? As already explained to you we are working on 3.1. It seems like your patch is compatible only with the master branch.

Where should I sent the patch?

I just read about the 3.1 from blogposts... not remember that we've discussed about the flow of sending of patches for "old" version :)

Edit: and one more note: the original issue affects 3.0 - I think it's better to apply to this first.

airween:
> _General comments_:
> 
>     * Try to give significative names to pull requests, so it can be searchable by other users.
> 
> 
> ```
> A good title could be: Adds missing check for runtime SecRuleRemoveByTag (Issue #2099)
> ```
> 
>     * Whenever you make a reference to an issue, use the hash marking otherwise Github won't notice the reference.
> 
> 
> ```
> Instead of 2099 use #2099
> ```

Sorry, and thanks for the tips - I'll try to keep it mind :)

Edit: these had been modified too.

src/rules.cc
airween: Do you think about like this?
https://github.com/SpiderLabs/ModSecurity/pull/2102/commits/5497b86e911fcf8e5195a8e87df25edc408d9219

zimmerle: Yes, exactly. There are three loops that aiming to do the very same thing. There is no reason to have one different for the others.

zimmerle:
> > Hi @airween,
> > I have started a review in the code.
> > Question: Why you have sent this patch against that branch? As already explained to you we are working on 3.1. It seems like your patch is compatible only with the master branch.
> 
> Where should I sent the patch?
> 
> I just read about the 3.1 from blogposts... not remember that we've discussed about the flow of sending of patches for "old" version :)
> 
> Edit: and one more note: the original issue affects 3.0 - I think it's better to apply to this first.

Sure @airween. I think that reiterates the need to talk over GitHub issues instead of Hangouts/Skype and Slack. So that we don't lose track of what is going on with the project. With that, we are keeping the info not only ourselves but for others who may want to be up to date and participating in the discussions.

No worries I will modify your patch to fit 3.1.

airween:
> > > Hi @airween,
> > > I have started a review in the code.
> > > Question: Why you have sent this patch against that branch? As already explained to you we are working on 3.1. It seems like your patch is compatible only with the master branch.
> > 
> > 
> > Where should I sent the patch?
> > I just read about the 3.1 from blogposts... not remember that we've discussed about the flow of sending of patches for "old" version :)
> > Edit: and one more note: the original issue affects 3.0 - I think it's better to apply to this first.
> 
> Sure @airween. I think that reiterates the need to talk over GitHub issues instead of Hangouts/Skype and Slack. So that we don't lose track of what is going on with the project. With that, we are keeping the info not only ourselves but for others who may want to be up to date and participating in the discussions.
> 
> No worries I will modify your patch to fit 3.1.

Hi @zimmerle - thanks for details.

I've double checked on Slack (a long time ago we didn't used Skype and Hangouts :)), and mailing list. On the list, I've found a mention for 3.1 in a mail, nothing more.

Of course, it's important to keep the control over the project.

I've never checked out the 3.1 - sounds interesting. Would you like me to see that too?

src/rules.cc
airween: Right, thanks for the clarification.

zimmerle:
Hi @airween, 

Thank you for your patch. I have made some modifications in your patch in order to:

(a) merging all the commits that you were offering fixies within the very same branch;
(b) baptise the commits with more meaningful description;
(c) Adding the test case to the test case list.

The result is here:
 - 2c136a2d9cda6bac4123fee22905a16fbe55b622
 - 7a93bea8f753f45bb3c5246b464977998ea5b462
 - db298696fa9f6f8eb7a14c31ff914474b6e2f6c8
 - b5823d4e0c53375d56f7e5089d6c6bec948ac339

Since you are a long-term contributor, it is important to highlight **(c)**.  If the test case is not on the make file list, it is not executed by the buildbots. Therefore not automagically tested. List is still manual, little trickier to get it done right.

Here is the list:
https://github.com/SpiderLabs/ModSecurity/blob/b5823d4e0c53375d56f7e5089d6c6bec948ac339/Makefile.am#L152


This patch is now merged!

airween:
Hi @zimmerle, thanks for the tips - I'll try to keep them in mind :).

Note, that I've added many test cases (may be for all relevant modifications), but never met whit this issue. Thanks for the notification. :)

PR #2103, opened 2019-05-28
Added missing check for ctl:ruleRemoveByTag for v3.1 (Issue #2099)

airween:
* added missing check for ctl:ruleRemoveByTag for v3.1
* added regression test file

### DIFF
--- a/src/rules_set.cc
+++ b/src/rules_set.cc
@@ -257,7 +257,7 @@ int RulesSet::evaluate(int phase, Transaction *t) {
                 }
             }
 
-            if (ruleWithActions) {
+            if (ruleWithActions && t->m_ruleRemoveByTag.empty() == false) {
                 for (auto &z : t->m_ruleRemoveByTag) {
                     if (ruleWithActions->containsTag(z, t) == true) {
                         ms_dbg_a(t, 9, "Skipped rule id '" \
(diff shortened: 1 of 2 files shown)


zimmerle:
Hi @airween,

It is my understanding that you have submitted the same code for the 3.0.3 tree. I am not merging this because since it was merged there, once I rebase the code will be available anyway. Therefore not necessary to merge this one.

Having said that, I am closing this pull request.

airween:
Hi @zimmerle,

> It is my understanding that you have submitted the same code for the 3.0.3 tree.
may be you misunderstand - maybe I do :).

I've submitted this patch to v3/master branch, but in this PR I submit to v3/dev/3.1-experimental. That's an another branch, not the 3.0.3.

> I am not merging this because since it was merged there, once I rebase the code will be available anyway. Therefore not necessary to merge this one.

right, I didn't know how would like to apply this fix to the 3.1 tree. As I see, this fix doesn't exists yet in 3.1:
https://github.com/SpiderLabs/ModSecurity/blob/13e073ccee32d9422c93729c6ee16e282ac34854/src/rules_set.cc#L259-L271

The another reason why I sent it here, that you asked me first at #2099, why did I sent the fix to v3/master, because you're working on 3.1 - therefore I'ld like help you with this.

> Having said that, I am closing this pull request.

Okay, it's no problem :).

PR #2107, opened 2019-06-01
V3/collection re fix

airween:
Looks like the Regex::searchAll() method is not case insensitive (InMemoryPerProcess::resolveSingleMatch() needs that). This bug detected with help of CRS regression test [920450](https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.1/dev/util/regression-tests/tests/REQUEST-920-PROTOCOL-ENFORCEMENT/920450.yaml).

The problem is, the comparing variables on collections (in-memory and LMDB) must be case insensitive.

This condition is not met.
See this [PR](https://github.com/SpiderLabs/ModSecurity/pull/1810), and the documentation.

BTW, the ModSec2 follows this way... so I think we have to align the ModSec3 too.

I'ld like to apply this patch to v3/master, because I found here, and I think the v3.1 regex module is different.


michaelgranzow-avi:
IMO this bug is not a duplicate as the other bug does not involve regular expression search of variables, but this one does.

src/utils/regex.cc
```diff
@@ -38,18 +38,21 @@ namespace modsecurity {
...
 
-Regex::Regex(const std::string& pattern_)
+Regex::Regex(const std::string& pattern_, bool caseSensitive)
     : pattern(pattern_.empty() ? ".*" : pattern_) {
     const char *errptr = NULL;
     int erroffset;
+    int flags = (PCRE_DOTALL|PCRE_MULTILINE);
 
-    m_pc = pcre_compile(pattern.c_str(), PCRE_DOTALL|PCRE_MULTILINE,
+    if (caseSensitive == true) {
```
michaelgranzow-avi, line 47: The name of the variable is wrong: we add the CASELESS flag if caseSensitive is true. Either keep the name, pass false in the invocations of the constructor changed above and test for !caseSensitive here; or change the name to ignoreCase.

As written it will confuse countless readers :(

airween: Thank for your remark - you're right.

Because this PR contains so many unnecessary commit, and this name convention is really illogical, I created a new one - see #2297.

Now I close this PR. Thanks for all helps to everybody.

airween:
Closing this PR, reopened in #2297.

Issue #2110, opened 2019-06-04
v3/master: ruleRemoveTargetById doesn't work with range id


theMiddleBlue:
Hi,

it seems that a "range id" can't be used on `ctl:ruleRemoveTargetById=<range>;ARGS:foo`, it works only by using a specific rule id. 

**regression test**
```json
[
  {
    "enabled":1,
    "version_min":300000,
    "title":"Testing ctl:ruleRemoveTargetById - issue 1444",
    "expected":{
      "http_code":200
    },
    "client":{
      "ip":"127.0.0.1",
      "port":123
    },
    "request":{
      "headers":{
        "Host":"localhost",
        "User-Agent":"curl/7.38.0",
        "Accept":"*/*"
      },
      "uri":"index.php?foo=bar&z=xxx",
      "method":"GET",
      "body": ""
    },
    "server":{
      "ip":"127.0.0.1",
      "port":80
    },
    "rules":[
        "SecRuleEngine On",
        "SecRule ARGS:foo \"@rx ^bar$\" \"id:100,phase:1,ctl:ruleRemoveTargetById=1000-1999;ARGS:z\"",
        "SecRule ARGS:z \"@rx ^xxx$\" \"id:1010,phase:1,deny,status:403\""
    ]
  }
]
```

**executing test**
```bash
ModSecurity 3.0.3 - tests
(options are not available -- missing GetOpt)

  # File Name                                         Test Name                                                             Passed?   
--- ---------                                         ---------                                                             -------   
  1 issue-1444.json                                   Testing ctl:ruleRemoveTargetById - issue 1444                         failed!

Test failed. From: test/test-cases/regression/issue-1444.json.
Test name: Testing ctl:ruleRemoveTargetById - issue 1444.
Reason: 
HTTP code mismatch. expecting: 200 got: 403

Debug log:
[15596415139.550631] [] [4] Initializing transaction
[15596415139.550631] [] [4] Transaction context created.
[15596415139.550631] [] [4] Starting phase CONNECTION. (SecRules 0)
[15596415139.550631] [] [9] This phase consists of 0 rule(s).
[15596415139.550631] [] [4] Starting phase URI. (SecRules 0 + 1/2)
[15596415139.550631] [index.php?foo=bar&z=xxx] [4] Adding request argument (GET): name "foo", value "bar"
[15596415139.550631] [index.php?foo=bar&z=xxx] [4] Adding request argument (GET): name "z", value "xxx"
[...]

PR #2114, opened 2019-06-04
Adds support to multiple ranges in ctl:ruleRemoveTargetById (#2110)

j0k2r:


CHANGES
```diff
@@ -1,6 +1,8 @@
 v3.0.4 - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+ - Adds support to multiple ranges in ctl:ruleRemoveTargetById
+   [Issue #2110 - @theMiddleBlue, @zimmerle]
```
zimmerle: @j0k2r why you did put my name on it and where is yours? :)

j0k2r: Oh, sorry I don't know the ModSecurity CHANGES rules, I took the the issue "participants"

dune73:
Is this PR covering the use of multiple comma-separated IDs too? (-> `ctl:ruleRemoveTargetById=1000,1999`)

If it does, then adding a test case for that use would be helpful.

j0k2r:
No, it only parse the Id range

j0k2r:
@dune73, After doing some tests, it seems that neither `ctl:ruleRemoveById` nor `ctl:ruleRemoveTargetById` supports setting multiple IDs.

I'm not a Lexer expert but apparently with the https://github.com/SpiderLabs/ModSecurity/blob/cbd15ec1382f7e2c3e19a93ea9ca3e8b4ced44f6/src/parser/seclang-scanner.ll#L447 **ruleRemoveBy** does not support (,)

Don't know if I'm right.

dune73:
Thank you for checking. Much appreciated. So many subtle shortcomings with v3...

zimmerle:
Thank you @j0k2r. Added this as part of #2430. Merged on 3.1-experimental

Issue #2100, opened 2019-05-20
Deocdage of UTF8 with utf8toUnicode() => urlDecodeUni() is wrong


ghost:
Hi,

**Describe the bug**

I have an issue with urlDecodeUni(). It seems that this function does
corruption on decoded strings. I detect the problem with a Russian
POST (luckyly it was crap).

I receive this url-encoded string:

```
   %D0%95%D1%81%D0%BB%D0%B8+ [...]
```

Once decoded, this string is UTF8 Russian message. Modsec/OWASP rules
decodes with the function utf8toUnicode(), it seems that this function
decodes UTF8 and encode it in Microst Url encoded style. It works, this
is the result:

```
   %u0415%u0441%u043b%u0438
```

Once encoded in Microsoft url-encoded style, we decode this encoding with
"urlDecodeUni()", and I obtain the following data (^U is \x15 with "binary"
representation):

```
   ^UA;8 20A
```

This is not a representation of the original message. Reading the code
of the function I understand the "urlDecodeUni()" removes the 2 most
significant bytes of 16 bit character. So:

```
   %u0415 becomes \x15 => ^U
   %u0441 becomes \x41 => "A" (ascii)
   %u043b becomes \x3b => ";" (ascii)
   %u0438 becomes \x38 => "8" (ascii)
```

The incoming message is corrupted, and ModSec matches whatever string.

So, my Russian blogpost which is simply a SPAM match the OWASP rule
941120 and it is blocked. In facts, the request doesn't contains technical
attack.

Maybe the rule "urlDecodeUni()" should provides UTF8 data ? UTF8 works
in 8 bit.

**Logs and dumps**

NOTE: origin log contains binary characters. Is here represented by ^U (\x15)

```
[...]

[155836496791.552370] [/wp-comments-post.php] [9] Target value: "1683" (Variable: ARGS:comment_post_ID)
[155836496791.552370] [/wp-comments-post.php] [9]  T (0) t:utf8toUnicode: "%u0415%u0441%u043b%u0438 %u0432%u0430%u0441 %u0438%u043d%u0442%u0435%u0440%u0435 (18560 characters omitted)"
[155836496791.552370] [/wp-comments-post.php] [9]  T (0) t:urlDecodeUni: "^UA;8 20A 8=B5@5ACNB :0G5AB25==K5 B>20@K, :>B>@K5 5ABL H0=A 8A?>;L7>20BL 2 :>A<5B (3755 characters omitted)"
[...]

PR #2119, opened 2019-06-13
Adds neutralizeSafeEncoding and utf8toChar transformations to fix utf8/base64/hex32 false positives

j0k2r:
Hi guys,

Sometimes, when we protect multiple international websites, we have to accept many UTF8 encoding and we can't have the possibility to set the right **UnicodeMapping** codepage every time.

With this PR we are proposing to add two transformations to solve and facilitate working with `utf8/base64/hex32` entries and limit false positives.

- **utf8toChar**: 
	- Replace all valid UTF8 characters > `0xff` by a single "**a**". This transformation doesn't modify ASCII characters.
	- Replace UTF8 overlong characters, this allows to detect attack escaped by overlong encoding. If the overlong is <= `0xff`, it is replaced by the corresponding byte.
	- If the transformation encounter any invalid UTF8 sequence, it returns the original input.
	- Example:

		Input string (containing a random "**ON**" word, which trigger XSS FP):
			`"test+%D1%8F%D0%B2%D0%BB%D1%8F%D1%8E%D1%82%D1%81%D1%8F+%D0%BD%D0%B5"`

		Output string:
			`"test aaaaaaaa aa"`

		And no false positive is triggered.

- **neutralizeSafeEncoding**:
	- Fix false positives due to random string pattern in base64 or hexadecimal strings.
	- If the full input string represent a known encoding (`base64, base64i, hex32, HEX32`), all the characters are replaced by a tag describing the escaped encoding, and fill other characters with "**a**".
	- If none of this known encoding is matched, it returns the original input.
	- Example:

		Input string (containing a random "**SELECT**" word, which trigger SQLi FP):
			`"Fk2SELECTIUaEGsgww==" `

		Output string:
			`"b64aaaaaaaaaaaaaaaaa"`

	And no false positive is triggered.

Related discussions:
- #2100
- https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/794
- https://github.com/SpiderLabs/owasp-modsecurity-crs/issues/1426


zimmerle:
Hi @j0k2r,

Avoid evasion is a tricky job. Having transformation more forgiven may not be a good idea. The rationale about having the Encoding functionality is to have ModSecurity understanding the inputs precisely as they are understood by the end application.

If I analyze the input differently from the web application, it is not exactly meaningful; it could give as two different scenarios: false positives or false negatives. It seems to me that by having the transformations more forgiveness, you are in favor of having more false negatives than false positives. That may be OK in your use case scenario. Not sure, however, if that is what is expected for most of the users.

I prefer to think that the challenge of writing a generic rule set - agnostic to the application, platform or language - is a challenge that should be taken care at the rule set itself; Ideally, we should not reduce the capability of the engine in favor of reducing the false positives. Instead, we could work by improving our rules to cover those specific scenarios better. Does it make sense to you?

PR #2130, opened 2019-07-10
Small fixes in log messages to help debugging the file upload

airween:
This patch contains some small not important but useful fix:

- change variable name in Multipart class destructor: the variable holds the m_uploadKeepFiles value, not the opposite
- when unlink() (remove temporary files) returns with 0, the message was "Failed to delete file", which is wrong
- when temporary file was created, not the used create mode showed

### DIFF
--- a/src/request_body_processor/multipart.cc
+++ b/src/request_body_processor/multipart.cc
@@ -70,7 +70,7 @@ Multipart::Multipart(std:: string header, Transaction *transaction)
 
 Multipart::~Multipart() {
     ms_dbg_a(m_transaction, 4,
-        "Multipart: Cleanup started (remove files " \
+        "Multipart: Cleanup started (keep files set to " \
         + RulesProperties::configBooleanString(
             m_transaction->m_rules->m_uploadKeepFiles) \
         + ")");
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @airween,

In the comments, you mentioned - 

>  change variable name in Multipart class destructor: the variable holds the m_uploadKeepFiles value, not the opposite

The name of the variable seems to be unchanged. However, there is a message change. That is really what you meant to do?

airween:
Hi @zimmerle,

> The name of the variable seems to be unchanged. However, there is a message change. That is really what you meant to do?

sorry, I wasn't accurate when I wrote the comment. I meant about `SecUploadKeepFiles` variable, and not for the name but for the meaning of this variable. If this configuration directive is set to `On`, then the engine will log:

`Multipart: Cleanup started (remove files True)`

but the IMHO the expected message should:

`Multipart: Cleanup started (keep files set to True)`

So, the _variable name_ in this context refers to the **keep files** (see SecUpload**KeepFiles**), not **remove** (eg. as SecUpload**RemoveFiles**).

You're right, I have to change the comment - do you have any idea, what can I write that's more accurate?

zimmerle:
Thank you @airween. The commit is Ok. The comment that you made here was not accurate, leading the impression that you have forgotten a commit or something similar. Now it is clarified. No need for you to do anything. I will proceed with the merging.

zimmerle:
Merged! Thank you!

PR #2134, opened 2019-07-17
Use long long instead of int for comparison, increment, decrement operations and persistent collections

Nakaner:
We are using ModSecurity at a download service for large files (hundreds of MB to multiple GB) to manage usage quotas (currently the number of connections per IP address). ModSecurity is able to manage traffic quotas by IP address but this is currently limited to 2^31 bytes because the counters are use the `int` type only. Using 64 bit integers allows to use ModSecurity to enforce traffic quotas beyond 4 GB.

Support for 64 bit integer is lacking in ModSecurity 3 as well. We run version 2 with Apache and this is basically our patch of the Debian package.


marcstern (approved):
using long long is a must

PR #2171, opened 2019-09-26
Cast the constant to the right type - FIX #2156

### LINKED ISSUE #2156: clang: warning: comparison of constant 160 with expression of type 'char' is always false
**Description**

/src/actions/transformations/remove_whitespace.cc:66:44: warning: comparison of constant 160 with expression of type 'char' is always false

#define NBSP 160    // non breaking space char
std::string value(val);
if (isspace(value[i]) || (value[i] == NBSP)) {

airween:
A possible solution for #2156.

### DIFF
--- a/src/actions/transformations/remove_whitespace.cc
+++ b/src/actions/transformations/remove_whitespace.cc
@@ -44,7 +44,7 @@ std::string RemoveWhitespace::evaluate(std::string value,
     // loop through all the chars
     while (i < value.size()) {
         // remove whitespaces and non breaking spaces (NBSP)
-        if (isspace(value[i]) || (value[i] == NBSP)) {
+        if (isspace(value[i]) || (value[i] == static_cast<char>(NBSP))) {
             value.erase(i, 1);
         } else {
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @airween, hide the warning is definitely not the solution for the problem.

airween:
Hi @zimmerle,

could you explain why is this just _hiding of warning_?

I made a small test file to demonstrate my idea:
https://gist.github.com/airween/c2f3ff400e4c4a94dcbed9aec6eca48a

Here is my opinion:
- the std::string is a sequence with element type `char`, therefore each item value could be between -127 and 128
- consider somehow there **is** an item with value 160 - see the example [line 25](https://gist.github.com/airween/c2f3ff400e4c4a94dcbed9aec6eca48a#file-rmwsp-cc-L25).
- when the casting comes (line 26-27), then the values converted from `unsigned char` to `signed char`, so the 160 will -96
- it's same what the `static_cast<>` does

In the example you can see that's not just the warning hide, it really removes the NBSP from sequence.

May be I'm wrong, but I'm very curious your opinion.

Issue #2033, opened 2019-02-27
Segmentation fault ctl:ruleRemoveTargetById without target


studersi:
Leaving out the target for `ctl:ruleRemoveTargetById` causes segmentation faults.

This has already been reported and fixed for `ctl:ruleRemoveTargetByTag`: https://github.com/SpiderLabs/ModSecurity/issues/1353.

My guess is that `ctl:ruleRemoveTargetByMsg` is affected as well but I did not test this.

Add a rule exclusion containing `ctl:ruleRemoveTargetById=941130` instead of `ctl:ruleRemoveTargetById=941130;ARGS:xxxx`

 - ModSecurity 2.9.3
 - Apache 2.4.38
 - Linux

studersi:
What is the status on this issue, do you need any more information?

zimmerle:
Hi, @studersi 

We don't have an ETA to fix this yet. It should be fixed, although the priority is not high as this does not directly impact run time. This seems to be a very welcome task to someone that wants to learn a little bit more about ModSecurity internals.

studersi:
This was fixed in https://github.com/SpiderLabs/ModSecurity/pull/2189. Closing the issue.

Issue #1353, opened 2017-03-14
Apache segmentation fault


LukeP21:
Specifying ctl:ruleRemoveTargetByTag= with only 1 parameter causes Apache to pass a configtest but seg fault.

Example:              
ctl:ruleRemoveTargetByTag=attack-disclosure

Tested on both:
Apache/2.4.23 with modsecurity-2.9.1
Apache/2.2.15 with modsecurity-2.9.1

I'm aware it should be ctl:ruleRemoveByTag=attack-disclosure just made a mistake and noticed seg faults.

zimmerle:
Closed as of: 9ac9ff8

PR #2189, opened 2019-10-23
This commit closes the issue #2033.

studersi:
Added a fix for issue #2033 based on #1353 and https://github.com/SpiderLabs/ModSecurity/commit/9ac9ff8223c0a9d7723a312204659ff72c9cc685.

The fix is basically the same as is used for #1353, adding a sanity check after the arguments are parsed. I added the check for `ctl:ruleRemoveTargetById` as well as `ctl:ruleRemoveTargetByMsg`.

I'm currently unable to check this in the environment where it first occurred and I currently cannot reproduce the error as it was reported to me.

Note: This change also fixes inconsistent whites pace on the original fix.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD mmm YYYY - 2.9.x (to be released)
 ------------------------------------
 
+ * Adds a sanity check before use ctl:ruleRemoveTargetById and ctl:ruleRemoveTargetByMsg.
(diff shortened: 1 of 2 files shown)


zimmerle:
Hi @studersi, 

Thank you for your contribution. I've changed the commit name to reflect what you have placed on the CHANGELOG.

studersi:
Perfect, thank you!

PR #2190, opened 2019-10-23
Fix test issue-1974

martinhsv:
A small typo that causes the related unit tests to not function correctly.  Note that this change should accompany the related updates to the secrules-language-tests repo.


zimmerle:
Merged! thanks!

Issue #2021, opened 2019-02-09
Match of "eq 0" against "MULTIPART_UNMATCHED_BOUNDARY" required


dwreski:
**Describe the bug**
For unknown reasons, we're see "access denied" errors with mod_security-2.9.2-5 on fedora28

[Fri Feb 08 22:06:50.144344 2019] [:error] [pid 11255:tid 140146947102464] [client 151.106.0.210:54982] [client 151.106.0.210] ModSecurity: Access denied with code 403 (phase 2). Match of "eq 0" against "MULTIPART_UNMATCHED_BOUNDARY" required. [file "/etc/httpd/conf.d/mod_security.conf"] [line "35"] [id "200003"] [msg "Multipart parser detected a possible unmatched boundary."] [hostname "linuxsecurity.com"] [uri "/index.php"] [unique_id "XF5Dyf3kxTYC2M5TGQgxsAAAANU"], referer: https://linuxsecurity.com/index.php?option=com_content&task=new&sectionid=9&itemid=0

**Logs and dumps**

Output of:
 1. DebugLogs (level 9)
 2. AuditLogs
 3. Error logs
 4. If there is a crash, the core dump file.

_Notice:_ Be carefully to not leak any confidential information.

**To Reproduce**
I don't know how to reproduce.

A **curl** command line that mimics the original request and reproduces the problem. Or a ModSecurity v3 test case.

[e.g: curl "modsec-full/ca/..\\..\\..\\..\\..\\..\\/\\etc/\\passwd" or [issue-394.json](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/test/test-cases/regression/issue-394.json)]

**Expected behavior**
This appears to be related to issue #1804 but that was reported to be fixed

**Server (please complete the following information):**
 - ModSecurity version (and connector): [e.g. ModSecurity v3.0.1 with nginx-connector v1.0.0]
 - WebServer: [e.g. nginx-1.15.5]
 - OS (and distro): [e.g. Linux, archlinux]

**Rule Set (please complete the following information):**
 - Running any public or commercial rule set? [e.g. SpiderLabs commercial rules]
 - What is the version number? [e.g. 2018-08-11]

**Additional context**

**Server (please complete the following information):**
 - ModSecurity version (and connector): [e.g. ModSecurity v3.0.1 with nginx-connector v1.0.0]
 - WebServer: [e.g. nginx-1.15.5]
 - OS (and distro): [e.g. Linux, archlinux]

[...]

Issue #1804, opened 2018-06-15
issue with MULTIPART_UNMATCHED_BOUNDARY [id "200003"] v2.9.2


Daijobou:
I get with upload a image this error:

> [:error] ModSecurity: Access denied with code 44 (phase 2). Match of "eq 0" against "MULTIPART_UNMATCHED_BOUNDARY" required. [file "/etc/httpd/conf.d/mod_security.conf"] [line "34"] [id "200003"] [msg "Multipart parser detected a possible unmatched boundary."] [hostname "www.example.com"] [uri "/myupload"] [unique_id "W...w"], referer: https://www.example.com/myupload
> [http:error] AH01579: Invalid response status 44, referer: https://www.example.com/myupload
> 
like https://github.com/SpiderLabs/ModSecurity/issues/652

I upload simple with 
`<form action="myupload" method="post" accept-charset="utf-8" enctype="multipart/form-data"><input type="file" name="image"><button type="submit" name="submit">Upload</button></form>`

one of many many many image-files with this error "--"

![2018-06-15_18_26_19-greenshot](https://user-images.githubusercontent.com/260343/41491196-f758d74e-70f7-11e8-9367-1b6b3a4d5dad.png)

I use modsecurity2.9.2-1.el7 and its set as "off". So why its catch this issue?

System: CentOS Linux 7.5.1804 (Core)‬ with Product Plesk Onyx Version 17.5.3

etc/httpd/conf.d/mod_security.conf

> <IfModule mod_security2.c>
>     
>     # Default recommended configuration
>     SecRuleEngine On
>     SecRequestBodyAccess On
>     SecRule REQUEST_HEADERS:Content-Type "text/xml" \
>          "id:'200000',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=XML"
>     SecRequestBodyLimit 13107200
>     SecRequestBodyNoFilesLimit 131072
>     SecRequestBodyInMemoryLimit 131072
>     SecRequestBodyLimitAction Reject
>     SecRule REQBODY_ERROR "!@eq 0" \
>     "id:'200001', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2"
>     SecRule MULTIPART_STRICT_ERROR "!@eq 0" \
>     "id:'200002',phase:2,t:none,log,deny,status:44,msg:'Multipart request body \
>     failed strict validation: \
[...]

Issue #652, opened 2014-01-30
Rule MULTIPART_UNMATCHED_BOUNDARY, id:'200003' fails on valid multipart/form-data submission


temyers:
The MULTIPART_UNMATCHED_BOUNDARY rule fires, denying the request if the multipart/form data includes "--" at the beginning of a line.  If there is anything else at the start of the line (e.g. " --") then the rule does not trigger.

This issue was identified against a Liferay 6.2 installation fronted by apache+modsecurity, submitting a freemarker code snippet.

A sample curl command:

```
[...]

PR #2193, opened 2019-10-27
Fix MULTIPART_UNMATCHED_BOUNDARY FP errors

airween:
There was a bug and a fix (#1747) in libmodsecurity3.

We ran into same problem with v2, so I made this patch, which fixes this bug as same way like in v3.
Hope that you'll allow, and will fix these issues:
#652 
#1804 
#2021

### DIFF
--- a/apache2/msc_multipart.c
+++ b/apache2/msc_multipart.c
@@ -1070,6 +1070,17 @@ int multipart_process_chunk(modsec_rec *msr, const char *buf,
                 {
                     char *boundary_end = msr->mpd->buf + 2 + strlen(msr->mpd->boundary);
                     int is_final = 0;
+                    /* if it match, AND there was a matched boundary at least,
+                       set the flag_unmatched_boundary to 2
+                       this indicates that there were an opened boundary, which
+                       matches the reference, and here is the final boundary.
+                       The flag will differ from 0, so the previous rules ("!@eq 0")
+                       will catch all "errors", without any modification, but we can
+                       use the new, permission mode with "@eq 1"
+                    */
+                    if (msr->mpd->boundary_count > 0) {
+                        msr->mpd->flag_unmatched_boundary = 2;
+                    }
 
                     /* Is this the final boundary? */
                     if ((*boundary_end == '-') && (*(boundary_end + 1)== '-')) {
--- a/apache2/re_variables.c
+++ b/apache2/re_variables.c
@@ -1607,7 +1607,12 @@ static int var_multipart_unmatched_boundary_generate(modsec_rec *msr, msre_var *
     apr_table_t *vartab, apr_pool_t *mptmp)
 {
     if ((msr->mpd != NULL)&&(msr->mpd->flag_unmatched_boundary != 0)) {
-        return var_simple_generate(var, vartab, mptmp, "1");
+        if (msr->mpd->flag_unmatched_boundary > 0) {
+            return var_simple_generate(var, vartab, mptmp, apr_itoa(mptmp, msr->mpd->flag_unmatched_boundary));
+        }
+        else {
+            return var_simple_generate(var, vartab, mptmp, "1");
+        }
     } else {
         return var_simple_generate(var, vartab, mptmp, "0");
     }
--- a/modsecurity.conf-recommended
+++ b/modsecurity.conf-recommended
@@ -82,8 +82,48 @@ FL %{MULTIPART_FILE_LIMIT_EXCEEDED}'"
 
 # Did we see anything that might be a boundary?
 #
+# Here is a short description about the ModSecurity Multipart parser: the
+# parser returns with value 0, if all "boundary-like" line matches with
+# the boundary string which given in MIME header. In any other cases it returns
+# with different value, eg. 1 or 2.
+#
+# The RFC 1341 descript the multipart content-type and its syntax must contains
+# only three mandatory lines (above the content):
+# * Content-Type: multipart/mixed; boundary=BOUNDARY_STRING
+# * --BOUNDARY_STRING
+# * --BOUNDARY_STRING--
+#
+# First line indicates, that this is a multipart content, second shows that
+# here starts a part of the multipart content, third shows the end of content.
+#
+# If there are any other lines, which starts with "--", then it should be
+# another boundary id - or not.
+#
+# After 2.9.3, there are two kinds of types of boundary errors: strict and permissive.
+#
+# If multipart content contains the three necessary lines with correct order, but
+# there are one or more lines with "--", then parser returns with value 2 (non-zero).
+#
+# If some of the necessary lines (usually the start or end) misses, or the order
+# is wrong, then parser returns with value 1 (also a non-zero).
+#
+# You can choose, which one is what you need. The example below contains the
+# 'strict' mode, which means if there are any lines with start of "--", then
+# ModSecurity blocked the content. But the next, commented example contains
+# the 'permissive' mode, then you check only if the necessary lines exists in
+# correct order. Whit this, you can enable to upload PEM files (eg "----BEGIN.."),
+# or other text files, which contains eg. HTTP headers.
+#
+# The difference is only the operator - in strict mode (first) the content blocked
+# in case of any non-zero value. In permissive mode (second, commented) the
+# content blocked only if the value is explicit 1. If it 0 or 2, the content will
+# allowed.
+#
+
 SecRule MULTIPART_UNMATCHED_BOUNDARY "!@eq 0" \
 "id:'200004',phase:2,t:none,log,deny,msg:'Multipart parser detected a possible unmatched boundary.'"
+#SecRule MULTIPART_UNMATCHED_BOUNDARY "@eq 1" \
+#"id:'200004',phase:2,t:none,log,deny,msg:'Multipart parser detected a possible unmatched boundary.'"
 
 # PCRE Tuning
 # We want to avoid a potential RegEx DoS condition


fzipi:
Still having this problem.. and there is a patch available....

Can someone merge this one?

nickcyberpunk:
Hi, 
is this patch going to be accepted/merged ? 

I'm also facing similar issues on modsec. 

Thanks

martinhsv:
I'm going to close this unmerged.

The code here mirrors the v3 implementation where some significant problem have been identified, including:
- the functionality removes so many use cases from triggering the flag to equal 1 that it's questionable how useful the check for MULTIPART_UNMATCHED_BOUNDARY still is
- while the intention of the implementation was to have a way to avoid signalling that condition when there is legitimate PEM content in one of the parts (i.e. it begins with '--' like a multipart boundary does); this implementation does not do that successfully if the PEM content is in the final of the multipart parts
- even if **none** of the multipart parts has a value beginning with '--', MULTIPART_UNMATCHED_BOUNDARY will incorrectly get set to 2

What really needs to happen with this functionality, is that it should be reviewed again in v3, with either a substantially revised solution implemented, or the '=2' code should be removed entirely. If the former happens, then that solution can be ported to v2.

In the immediate period for v2, likely the best way to handle FPs for PEM data is likely to stop using rule 200004 from modsecurity.conf-recommended for those cases where PEM (or similar) data is expected.

airween:
> The code here mirrors the v3 implementation

Yes, because that implementation had [allowed](https://github.com/SpiderLabs/ModSecurity/pull/1747#issuecomment-396463392) after a discussion.

> where some significant problem have been identified, including:

>     * the functionality removes so many use cases from triggering the flag to equal 1 that it's questionable how useful the check for MULTIPART_UNMATCHED_BOUNDARY still is

so many? Can you show me a few ones?

>     * while the intention of the implementation was to have a way to avoid signalling that condition when there is legitimate PEM content in one of the parts (i.e. it begins with '--' like a multipart boundary does); this implementation does not do that successfully if the PEM content is in the final of the multipart parts

note, that this implementation wasn't created only to avoid the FP of the PEM content. We have ran into several other issues.

>     * even if **none** of the multipart parts has a value beginning with '--', MULTIPART_UNMATCHED_BOUNDARY will incorrectly get set to 2

Yes, this bug has discovered later - but it is still [unmerged](https://github.com/SpiderLabs/ModSecurity/pull/2417).

> What really needs to happen with this functionality, is that it should be reviewed again in v3, with either a substantially revised solution implemented, or the '=2' code should be removed entirely. If the former happens, then that solution can be ported to v2.

right. Is there any plan/priority to fix it?
 
> In the immediate period for v2, likely the best way to handle FPs for PEM data is likely to stop using rule 200004 from modsecurity.conf-recommended for those cases where PEM (or similar) data is expected.

as I wrote above, the PEM file was just an example. In many cases we ran into this problem with a regular PDF/JPG...

I'm not sure it's a good idea to remove that rule.

Btw: I think this fix is even better than the current state, and merge would be better than remove the rule 200004.

martinhsv:
Hello @airween ,

Much of this discussion might have been better located in the new issue for 'either a substantially revised solution implemented, or the '=2' code should be removed entirely.', but since I haven't created it yet, I'll respond to your main points here.

[Re: "...the functionality removes so many use cases from triggering the flag to equal 1..."] "Can you show me ... "

Basically, most cases of **actual** unmatched boundaries no longer result in the flag equal to 1. For example:
```
-----------------------------423513681827540048931513055996
Content-Disposition: form-data; name="a"

1
-----------------------------wrong-not-matched
Content-Disposition: form-data; name="b"

2
-----------------------------423513681827540048931513055996
Content-Disposition: form-data; name="c"

3
-----------------------------423513681827540048931513055996--
```

Re: "... but is still unmerged"

Yes, that's right. And it remains unmerged precisely because it does not address the fundamental design issues with the overall '=2' functionality that I first noted in that item in Oct. 2020.

"Btw: I think this fix is even better than the current state, and merge would be better than remove the rule 200004."

Even given all of the problems that I have outlined? We could engage further on why you think that's the case but that may be better done in the new issue that I plan to create (as mentioned at the top of this response).

"Is there any plan/priority ... "

Not yet. Not everything can be a top priority item. Activity on the project can and should be guided by what are deemed by the community to be high priority items. Indeed, such input is part of the reason that #2606 was implemented in an earlier timeframe than otherwise might have been the case. If it is the opinion of the CRS group that coming to some more permanent resolution on this matter is their most important issue, I will, of course, take that into account.

martinhsv:
Btw, in case it is helpful to you, I have created a pull request with additional test cases that demonstrate the failure to set the variable to '1': #2681

airween:
Hi @martinhsv,

thanks for the answers.

> Much of this discussion might have been better located in the new issue for 'either a substantially revised solution implemented, or the '=2' code should be removed entirely.', but since I haven't created it yet, I'll respond to your main points here.

IMHO it's not a good idea to stay back to a wrong state. (Definition of wrong state: see later)

>> [Re: "...the functionality removes so many use cases from triggering the flag to equal 1..."] "Can you show me ... "
> 
> Basically, most cases of **actual** unmatched boundaries no longer result in the flag equal to 1. For example:

may be we can consider it as philosophical question, but the given example does not contain any unmatched boundaries.

> ```
> -----------------------------423513681827540048931513055996
> Content-Disposition: form-data; name="a"
> 
> 1
> -----------------------------wrong-not-matched
> Content-Disposition: form-data; name="b"
> 
> 2
> -----------------------------423513681827540048931513055996
> Content-Disposition: form-data; name="c"
> 
> 3
> -----------------------------423513681827540048931513055996--
> ```

Consider the request:

```
-----------------------------423513681827540048931513055996
Content-Disposition: form-data; name="a"

== My attachment starts here ==
1
-----------------------------wrong-not-matched
Content-Disposition: form-data; name="b"

2
== My attachment ends here ==
-----------------------------423513681827540048931513055996
Content-Disposition: form-data; name="c"

3
-----------------------------423513681827540048931513055996--
```

Do you still consider it as unmatched boundary? (This is why I think the previous state was wrong.)

>> Re: "... but is still unmerged"
> 
> Yes, that's right. And it remains unmerged precisely because it does not address the fundamental design issues with the overall '=2' functionality that I first noted in that item in Oct. 2020.

Value `2` was a forced solution. I think the definition of _unmatched boundaries_ is still not clear (or wrong). The example above is not that (from my point of view). The correct value in that case should be 0, not 2 (and not 1, of course).
 
>> "Btw: I think this fix is even better than the current state, and merge would be better than remove the rule 200004."
> 
> Even given all of the problems that I have outlined? We could engage further on why you think that's the case but that may be better done in the new issue that I plan to create (as mentioned at the top of this response).

I'm afraid our assumptions are different about "problems" :).

I can't consider the given example as unmatched boundary. So yes, I think that's a better option.

>> "Is there any plan/priority ... "
> 
> Not yet. Not everything can be a top priority item.

Okay, it's no problem. But can we postpone the removing of this feature until the new PR is available?

> Activity on the project can and should be guided by what are deemed by the community to be high priority items. Indeed, such input is part of the reason that #2606 was implemented in an earlier timeframe than otherwise might have been the case. If it is the opinion of the CRS group that coming to some more permanent resolution on this matter is their most important issue, I will, of course, take that into account.

Thanks for the offer. This means, if somebody send a new PR, you will handle it with priority?


Thanks again.

martinhsv:
Hi @airween ,

In my example, the request body does not violate any RFCs or anything like that. From a standards point of view, everything is legitimate. The line with 'wrong-not-matched' is simply the second line of data for the part with the name 'a'. And the line after that (that begins with 'Content-Disposition') is the third line of the 'a' part, and so on.

However, I say with fairly strong conviction that that pattern is a use cases that was intended to signal MULTIPART_UNMATCHED_BOUNDARY=1 when that functionality was introduced to ModSecurity.

But suppose you're not convinced that I'm correct. Then what cases do you think, **should** result in MULTIPART_UNMATCHED_BOUNDARY=1 ? I would suggest that knowing an answer to that question would be a requirement before implementing changes.

This relates to your final question. A PR represents an implementation of some functionality. In assessing it one must ask: does it implement the requirements (without breaking anything)? But if we don't know what the requirements are, then it's pretty hard to conclude that a particular code change meets the requirements.

Fundamentally we have this question: Is there any point to anything akin to what MULTIPART_UNMATCHED_BOUNDARY does? Are there any actual web server technologies where boundary-like strings could result in parser inconsistencies, and hence an opportunity for evasion?

It is possible that there were one or more such cases back in 2007 when the MULTIPART_UNMATCHED_BOUNDARY was originally introduced (although even at that time the functionality was noted as being 'Not very reliable'). But perhaps there are no such cases now and perhaps haven't been for more than a decade. If that is indeed the case, we should probably move in the direction of deprecating MULTIPART_UNMATCHED_BOUNDARY. There's no point having functionality that both doesn't do anything useful and requires management to handle FPs.

airween:
hi @martinhsv,
 
> In my example, the request body does not violate any RFCs or anything like that.

I didn't even say the opposite :)

> From a standards point of view, everything is legitimate. The line with 'wrong-not-matched' is simply the second line of data for the part with the name 'a'. And the line after that (that begins with 'Content-Disposition') is the third line of the 'a' part, and so on.

yes, that's correct.

> But suppose you're not convinced that I'm correct. Then what cases do you think, **should** result in MULTIPART_UNMATCHED_BOUNDARY=1 ? I would suggest that knowing an answer to that question would be a requirement before implementing changes.

That's a good question, and I have been thinking about this too since your last comment. :)

> Fundamentally we have this question: Is there any point to anything akin to what MULTIPART_UNMATCHED_BOUNDARY does?

perhaps no, there isn't.

> Are there any actual web server technologies where boundary-like strings could result in parser inconsistencies, and hence an opportunity for evasion?

I'm not sure, but it may need to be researched. I'm afraid it's very dependent on implementation. I mean, there are as many interpretations as there are implementations. ModSecurity has its own - and as we can see, that's not a real good one.

> It is possible that there were one or more such cases back in 2007 when the MULTIPART_UNMATCHED_BOUNDARY was originally introduced (although even at that time the functionality was noted as being 'Not very reliable'). But perhaps there are no such cases now and perhaps haven't been for more than a decade. If that is indeed the case, we should probably move in the direction of deprecating MULTIPART_UNMATCHED_BOUNDARY. There's no point having functionality that both doesn't do anything useful and requires management to handle FPs.

Absolutely agree this.

May be the solution will that you have wrote above: make the `MULTIPART_UNMATCHED_BOUNDARY` variable deprecated. I'm going to make some comparison and research about how webservers handle the MP requests. Let me check that.

PR #2199, opened 2019-11-07
Fix nolog rules logging to part H

martinhsv:
This change is in a response to issue-2196.  Rules with 'nolog' were incorrectly logging to part H, including for innocent requests.

Note that this change does not address cases where a 1st rule in a chain has nolog but the subsequent rule(s) in the chain do not.  We may follow up with another code change to address that minority situation.


zimmerle:
Merged! Thanks!

PR #2201, opened 2019-11-13
Replace Cookie parsing method

airween:
This PR fixes three issues in libmodsecurity3:

- aligns the cookie parsing for v2: allows the cookie without value, eg `Cookie: foo` will produces a cookie with name `foo`; the old method skipped it
- processes the whole text after the first `=`, eg. `Cookie: foo=bar=something interesting` will produces a cookie with name `foo` and value `bar=something interesting`; the old method produced the value only `bar`
- fix the offset calculation for the logfiles

### DIFF
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -548,20 +548,53 @@ int Transaction::addRequestHeader(const std::string& key,
 
     if (keyl == "cookie") {
         size_t localOffset = m_variableOffset;
+        size_t pos;
         std::vector<std::string> cookies = utils::string::ssplit(value, ';');
         for (const std::string &c : cookies) {
-            std::vector<std::string> s = utils::string::split(c,
-               '=');
-            if (s.size() > 1) {
-                if (s[0].at(0) == ' ') {
-                    s[0].erase(0, 1);
-                }
-                m_variableRequestCookiesNames.set(s[0],
-                    s[0], localOffset);
+            // skip empty substring, eg "Cookie: ;;foo=bar"
+            if (c.empty() == true) {
+                localOffset++; // add length of ';'
+                continue;
+            }
 
-                localOffset = localOffset + s[0].size() + 1;
-                m_variableRequestCookies.set(s[0], s[1], localOffset);
-                localOffset = localOffset + s[1].size() + 2;
+            // find the first '='
+            pos = c.find_first_of("=", 0);
+            std::string ckey = "";
+            std::string cval = "";
+
+            // if the cookie doesn't contains '=', its just a key
+            if (pos == std::string::npos) {
+                ckey = c;
+            }
+            // else split to two substrings by first =
+            else {
+                ckey = c.substr(0, pos);
+                // value will contains the next '=' chars if exists
+                // eg. foo=bar=baz -> key: foo, value: bar=baz
+                cval = c.substr(pos+1);
+            }
+
+            // ltrim the key - following the modsec v2 way
+            while (ckey.empty() == false && ckey.at(0) == ' ') {
+                ckey.erase(0, 1);
+                localOffset++;
+            }
+
+            // if the key is empty (eg: "Cookie:   =bar;") skip it
+            if (ckey.empty() == true) {
+                localOffset = localOffset + c.length() + 1;
+                continue;
+            }
+            else {
+                // handle cookie only if the key is not empty                
+                // set cookie name
+                m_variableRequestCookiesNames.set(ckey,
+                        ckey, localOffset);
+                localOffset = localOffset + ckey.size() + 1;
+                // set cookie value
+                m_variableRequestCookies.set(ckey, cval,
+                        localOffset);
+                localOffset = localOffset + cval.size() + 1;
             }
         }
     }
--- a/test/test-cases/regression/variable-REQUEST_COOKIES.json
+++ b/test/test-cases/regression/variable-REQUEST_COOKIES.json
@@ -2,7 +2,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing Variables :: REQUEST_COOKIES (1/3)",
+    "title":"Testing Variables :: REQUEST_COOKIES (1/5)",
     "client":{
       "ip":"200.249.12.31",
       "port":123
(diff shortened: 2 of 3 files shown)


airween:
I made this patch again to fix the cookie parsing issues, and closed the #2023.

Special thanks for the help to @theMiddleBlue and @fgsch.

theMiddleBlue:
Thanks @airween 

Due to the criticality, is it possible to merge this PR as soon as possible and create a new "v3.0.something" release?

src/transaction.cc
```diff
@@ -548,20 +548,54 @@ int Transaction::addRequestHeader(const std::string& key,
...
+        size_t pos;
         std::vector<std::string> cookies = utils::string::ssplit(value, ';');
         for (const std::string &c : cookies) {
-            std::vector<std::string> s = utils::string::split(c,
-               '=');
-            if (s.size() > 1) {
-                if (s[0].at(0) == ' ') {
-                    s[0].erase(0, 1);
+            // skip empty substring, eg "Cookie: ;;foo=bar"
+            if (c.length() == 0) {
```
martinhsv: c.empty() instead of testing for equality with 0.

src/transaction.cc
```diff
@@ -548,20 +548,54 @@ int Transaction::addRequestHeader(const std::string& key,
...
+            // handle cookie only if the key is not empty
+            if (ckey.length() > 0) {
+                // ltrim the key - following the modsec v2 way
+                while (ckey.length() > 0 && ckey.at(0) == ' ') {
+                    ckey.erase(0, 1);
+                    localOffset++;
+                }
+
+                // if the key is empty (eg: "Cookie:   =bar;" skip it
+                if (ckey.length() == 0) {
```
martinhsv: Again, consider testing ckey.empty(), instead of length equaling 0.

src/transaction.cc
```diff
@@ -548,20 +548,54 @@ int Transaction::addRequestHeader(const std::string& key,
...
+            // if the cookie doesn't contains '=', its just a key
+            if (pos == std::string::npos) {
+                ckey = c;
+            }
+            // else split to two substrings by first =
+            else {
+                ckey = c.substr(0, pos);
+                // value will contains the next '=' chars if exists
+                // eg. foo=bar=baz -> key: foo, value: bar=baz
+                cval = c.substr(pos+1, c.length());
```
martinhsv: The 2nd argument is not required here; it's more straightforward to let the 2nd arg default.

src/transaction.cc
```diff
@@ -548,20 +548,54 @@ int Transaction::addRequestHeader(const std::string& key,
...
+                // ltrim the key - following the modsec v2 way
+                while (ckey.length() > 0 && ckey.at(0) == ' ') {
+                    ckey.erase(0, 1);
+                    localOffset++;
+                }
+
+                // if the key is empty (eg: "Cookie:   =bar;" skip it
+                if (ckey.length() == 0) {
+                    localOffset = localOffset + c.length() + 1;
+                    continue;
```
martinhsv: The flow here is more complicated than it needs to be.  Consider amending as follows:
1) move the left-trimming while loop (currently lines 580-583) to be right after line 576
2) follow that with a single 'if (ckey.empty()) {' where the 'if' block has what's currently at line 587 and the 'else' block has the 591-598 content.

martinhsv (commented):
I don't see any logic problems (and I definitely like that this addresses the ';;' use case!) 

I have made a few suggestions below for your consideration.  I'll defer to @zimmerle as to whether any of them should be considered necessary.

src/transaction.cc
airween: Changed in e14f3dc.

src/transaction.cc
airween: Changed in e14f3dc.

src/transaction.cc
airween: Changed in e14f3dc.

airween (commented):
Hi @martinhsv, thanks for the suggestions. As you can see, I modified all of them. The regression tests are passed at my side, hope you will allow this patch.

zimmerle:
@airween, as you may figure having a commit with the name "Changed @martinhsv remarks" in the ModSecurity tree is not elegant. As stated before, please do not pull request changes on your own pull request. Make the fix in your own commit.

zimmerle:
I understand that this pull request is a continuation of #2023; Since #2023 was on the priority list, I am passing this to the priority list. Wating for the author to continue the review.

zimmerle:
@theMiddleBlue the patch is marked to be part of the effort towards the upcoming release.

theMiddleBlue:
Thanks @zimmerle, I understand and I really appreciate all efforts. This PR addresses multiple vulnerabilities and one of them really critical. That's why I see the need for a security release, what do you think about? 3.1.0 is going to be released soon? otherwise, I think that all 3.0 users should patch asap.

airween:
hi @zimmerle,

> @airween, as you may figure having a commit with the name "Changed @martinhsv remarks" in the ModSecurity tree is not elegant.

I'm sorry, I just wanted to keep track of every step what I made. I think it's more clear when I make a separated commit which contains the modification for the suggestions.

Anyway, I revoked the last commit, and added the changes to the previous commit (with `amend`).

Hope now this will be appropriate.

theMiddleBlue:
@zimmerle if we can meet in a private slack chat, we can show you why this patch is critical and why we think that a security release should be published. I don't want to share more details in a public PR :)

martinhsv:
Hi @theMiddleBlue,

Just as an FYI, if you have sensitive, security-related concerns, one way to communicate these is to use the method mentioned in https://github.com/SpiderLabs/ModSecurity under the 'Security issue' heading.

theMiddleBlue:
Hi @martinhsv 

I just sent an e-mail with all the details. If you want, and if it helps to accelerate a bit, with @airween we can discuss it on Slack.

zimmerle:
> @zimmerle if we can meet in a private slack chat, we can show you why this patch is critical and why we think that a security release should be published. I don't want to share more details in a public PR :)

Thank you @theMiddleBlue. I saw that you guys are already feeding @martinhsv with information.

zimmerle:
Closing this on the favor of #2202 --- #2202 includes @airween fixes.

PR #2203, opened 2019-11-20
Adjusting rule chain logging; adjusting cookie header parsing

martinhsv:


airween:
Hi @martinhsv,

just my not-so-essential opinion, but the 2fb7eef commit has no effect, because in our last affected [commit](https://github.com/SpiderLabs/ModSecurity/blob/038522ad9b69466cf0e8adf32a388ee7ebf0bcfa/src/transaction.cc#L588-L592) we remove the unnecessary spaces from begin of all cookie keys.

Instead of this new method, I think it would make more sense to modify the other trim method:
```
   559              while (!final_cookie_pair.empty() && isspace(final_cookie_pair.back())) {
   560                  final_cookie_pair.pop_back();
   561              }
```
by this:
```
   559              while (!final_cookie_pair.empty() && (isspace(final_cookie_pair.back()) || final_cookie_pair.back() == ';')) {
   560                  final_cookie_pair.pop_back();
   561              }
```

With this we can trim the cookie string like these:
`Cookie: foo=bar; baz=foo     ` // this works now
`Cookie: foo=bar; baz=foo     ;    ` // this would be more useful
`Cookie: foo=bar; baz=foo     ;` // this could gives back same
`Cookie: foo=bar; baz=foo;` // same here

What do you think?

zimmerle:
1b4dcdcec4a0307a20c28fc74f252e6f88d98b9d is now merged and part of v3/master.

zimmerle:
Merged last @martinhsv`s changes. Thank you.

PR #2219, opened 2019-12-12
Accepts JSON without key [Issue #1576]

marcstern:
This accepts a valid JSON request with only a number or a string, nothing else.
Example:
--29000000-B--
Host: test
Content-Length: 2
User-Agent: Mozilla 4.0
Accept: text/html, image/gif, image/jpeg, *; q=.2, /; q=.2

--29000000-C--
25
--29000000-F--

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD mmm YYYY - 2.9.x (to be released)
 ------------------------------------
 
+ * Accepts JSON without key
+   [Issue #1576 - @marcstern]
  * Adds a sanity check before use ctl:ruleRemoveTargetById and ctl:ruleRemoveTargetByMsg.
    [Issue #2033 - @studersi]
  * Fix the order of error_msg validation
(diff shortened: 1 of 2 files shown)


marcstern:
Anything I can do to have this reviewed/accepted?

zimmerle:
Sorry for the delay.

That is in the review queue. Other users who fill interest to have this feature could participate on the discussion here. The attention from the community to a certain feature is taken into consideration to prioritize the merge.

Up Front, I can tell that I am missing test cases for this feature. Not mandatory at this point, but, it will help a lot during the review process.

marcstern:
I cannot find a test for the JSOn parser, as there's for the multipart parser for instance. If there's one, I can add this test case.

marcstern:
I created a new PR with a test: https://github.com/SpiderLabs/ModSecurity/pull/2363
I hope it'll be accepted now.

PR #2234, opened 2020-01-08
Support configurable limit on number of arguments processed

martinhsv:
This is basically the same as pull request 2060, but adds automated tests.


zimmerle:
Making reference to #2060

zimmerle:
@jleproust @martinhsv - Since the configuration directive is just arbitrating on the number of JSON parameters, it would be fair to call it **SecJSONArgumentsLimit** or maybe other name that references the fact that such configuration is related to the JSON implementation?

martinhsv:
The code in the pull request will limit the number of Arguments (within m_variableArgs) no matter how the elements are created.  E.g. if you set the configurable value to '5', and send a request like this:
```
GET /index.html?k1=v1&my_idd=aaa&z=1&y=2&x=3&w=4&aaa=hello HTTP/1.1
```
... then the argument 'aaa' will get omitted/skipped.

So, SecArgumentsLimit, is the right choice to go with the rest of the code.

(The JSON-specific portion of the pull request is a minor extra bit of functionality to stop json-parsing if the argument limit has been reached.)

jleproust:
@martinhsv Thanks a lot for adding the test! It would be really great if we could merge this.

zimmerle:
Thank you guys @martinhsv @jleproust

zimmerle:
Minor update, added: bad5892b93297d2f5284b8a6dc00026b91a57588.

[secargumentslimit.json](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/pull_2234/test/test-cases/regression/secargumentslimit.json) is now part of [Makefile.am](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/pull_2234/Makefile.am). Otherwise it won't run on **make check**.

https://github.com/SpiderLabs/ModSecurity/blob/bad5892b93297d2f5284b8a6dc00026b91a57588/Makefile.am#L187

The new version is on the build bots.

zimmerle:
merged!

jleproust:
Yay thanks!

rdegez:
Hi,

I thought this change made it to v3.0.4 like the changelog in `https://github.com/SpiderLabs/ModSecurity/blob/v3/master/CHANGES` file suggest : 
```
v3.0.4 - 2020-Jan-13
--------------------

 - Support configurable limit on number of arguments processed
   [@jleproust, @martinhsv]
<...>
```

Except that it didn't :-o
It's in master tree but definitely not in v3.0.4 release :-/

I tried to set `SecArgumentsLimit` and the directive was unknown.

Then I checked in the code :

```
~/modsecurity-master$ grep -ri CONFIG_DIR_ARGS_LIMIT * | wc -l
23

~/modsecurity-v3.0.4$ grep -ri CONFIG_DIR_ARGS_LIMIT * | wc -l
0
``` 

Am I missing something ?

Regards,

martinhsv:
Hi @rdegez ,

You are correct, that code is not in v3.0.4, and the CHANGES file is incorrect.  Thanks for raising that.

I will correct it.

martinhsv:
The CHANGES file has been corrected.

mlosapio:
I'm fairly certain this issue is only partially fixed with this patch. I'm happy to reopen an issue or file a new one.. 

Confirmation the patch works when content-type == json: 
```
time curl -d @test.json https://localhost -k  -H 'Content-Type: application/json'
<html>
<head><title>405 Not Allowed</title></head>
<body>
<center><h1>405 Not Allowed</h1></center>
<hr><center>nginx</center>
</body>
</html>

real    1m31.949s
user    0m0.060s
sys     0m0.074s
```

Switch the feature on:

```
sed -i 's/#SecArgumentsLimit 6/SecArgumentsLimit 6/g' /etc/nginx/modsec/modsecurity.conf
systemctl restart nginx
time curl -d @test.json https://localhost -k  -H 'Content-Type: application/json'
<html>
<head><title>400 Bad Request</title></head>
<body>
<center><h1>400 Bad Request</h1></center>
<hr><center>nginx</center>
</body>
</html>

real    0m0.200s
user    0m0.050s
sys     0m0.076s
```

But if you change the Content-Type to something other than json such as "application/x-www-form-urlencoded", the DDoS condition still exists and an nginx thread is consumed.

martinhsv:
Hi @mlosapio ,

This configurable limit accomplishes two things:
1) Overall, it will stop adding key-value pairs to the relevant ModSecurity variables (e.g ARGS) when the configured limit has been reached and
2) For json content specifically, it will also stop json parsing once the limit has been reached.

The main performance issue being addressed here is the _very_ large number of key-value pairs that might be processed by a large number of regular expressions in any sort of generic rule set (whether OWASP CRS or other).  Change 1 accomplishes that even for non-json content.  See the code changes in transaction.cc.

The second change, that applies only to json content, is a more minor benefit.

I'm fairly sure that Change 1 works correctly for a large set of application/x-www-form-urlencoded args.  And I do recall trying that while reviewing the change back in February.

Nevertheless, if you think the number of arguments added to ARGS (etc.) is not being limited for non-json content, feel free to submit a pull request, or otherwise outline what you think is going wrong.

mlosapio:
I’ll spend some time looking at it. The issue I’m seeing is when you do something like this:

curl -d @test.json https://localhost -k  -H 'Content-Type: application/x-www-form-urlencoded' 

While it may be technically incorrect, it still has pretty serious adverse effects. 

My assumption is that it’s skipping json.cc but the thread is still getting bogged down

martinhsv:
Hi @mlosapio ,

I have a suspicion as to what you are encountering.  Please contact me at the email address listed in https://github.com/SpiderLabs/ModSecurity under the 'Security issue' heading.

PR #2235, opened 2020-01-10
Configure script issues when packaging for Buildroot

frankvanbever:
I encountered these issues while packaging libmodsecurity for Buildroot.

Buildroot can be compiled for uClinux targetting ARM Cortex-M4 processors. The build for this platform fails due to the configure script not recognizing the platform `arm-buildroot-uclinux-uclibcgnueabi`. 
Another minor issue is that the configure script returns 0 even though no Makefile was generated. Returning a non-zero value causes the build to fail in the configure step instead of the make step.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -194,7 +194,7 @@ case $host in
     ;;
        *)
     echo "Unknown CANONICAL_HOST $host"
-    exit
+    exit 1
     ;;
 esac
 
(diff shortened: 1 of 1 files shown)


zimmerle:
Hi @frankvanbever,

Thank you for the pull request. I'll have a look at it.

frankvanbever:
Thanks @zimmerle 

For reference libmodsecurity and nginx-modsecurity have been submitted to Buildroot here:
http://patchwork.ozlabs.org/patch/1221103/
http://patchwork.ozlabs.org/patch/1221102/

zimmerle:
Merged! Thanks!

PR #2253, opened 2020-01-31
modsecurity.pc.in: add Libs.private

### LINKED ISSUE #1918: Incomplete pkg-config for static linking
Hello,

I tried to build the `v3/master` branch today to attempt a static linking of `libmodsecurity` in an autotools project. While this works fine, ultimately when I try to run the executable I get this:

> dlerror: undefined symbol: xmlFree

After a quick look at your `modsecurity.pc` file I realized that the private dependencies are missing. You should at least advertise `Libs.private` for linking and maybe `Requires.private` if some of your dependencies' API leak into your own headers (eg. a type from another library).

I'm only reporting this and not submitting a patch because you have optional dependencies so I'm not sure how to best piece it together.

More info on static linking and private dependencies:

https://people.freedesktop.org/~dbn/pkg-config-guide.html#writing

cc @gquintard

ffontaine:
Add Libs.private to save static dependencies so applications linking
statically with modsecurity will be able to retrieve them through
pkg-config

Fix #1918

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/modsecurity.pc.in
+++ b/modsecurity.pc.in
@@ -8,3 +8,4 @@ Description: ModSecurity API
 Version: @MSC_VERSION_WITH_PATCHLEVEL@
 Cflags: -I@includedir@
 Libs: -L@libdir@ -lmodsecurity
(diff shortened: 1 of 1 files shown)


zimmerle:
Merged! Thanks!

PR #2263, opened 2020-02-06
Fix #610: fix memory leaks caused by unfreed compiled regex data

### LINKED ISSUE #610: Memory leak if builded modsecurity with --enable-pcre-study
In  apache2/msc_pcre.c line 68-74, we will get regex->pe from pcre_study() if we compile with --enable-pcre-study.

 #ifdef WITH_PCRE_STUDY
        #ifdef WITH_PCRE_JIT
                pe = pcre_study(regex->re, PCRE_STUDY_JIT_COMPILE, &errptr);
        #else
                pe = pcre_study(regex->re, 0, &errptr);
        #endif
 #endif

However we will use pcre_free() or free() to free it in msc_pcre_cleanup(), the right choice here is pcre_free_study().
        if (regex->pe != NULL) {
# if defined(VERSION_NGINX)

```
        pcre_free(regex->pe);
```
# else

```
        free(regex->pe);
```
# endif

```
        regex->pe = NULL;
    }
```

This will lead to memory leak, the memory that leaked was allocated use  SLJIT_MALLOC in pcre_study().

This can be a big problem if we use Apache graceful restart.

asterite3:
This pull request aims to fix 2 memory leaks in ModSecurity v2:
1. PCRE JIT data produced by `pcre_study` not freed by  `msc_pcre_cleanup` because `free()`/`pcre_free`  is used to deallocate `pcre_extra` structure. The proposed fix is to use  `pcre_free_study` instead of `free`/`pcre_free` in `msc_pcre_cleanup`. This change was already proposed in #610.
2. Regex data not freed (`msc_pcre_cleanup` not called) for `@rx` rules which regexes contain variable substitutions.  OWASP CRS contains such rules: 920420 and 920480. The proposed  fix is to allocate memory for regexes in `msr->mp` pool instead of `rule->ruleset->mp` in `msre_op_rx_execute`. `msre_op_rx_execute` is called for each request (unlike `msre_op_rx_param_init`) and creates volatile regexes used for that one request only (which is done when regex contains variable substitution and therefore can't be compiled once and reused for each request), that regex does not seem to be stored anywhere in ruleset data.

These two fixes are related: without the second one, for rules with regex depending on a variable, function  `msc_pcre_cleanup` is never called, so the first fix does not help. The other way around, when PCRE JIT is used, the second fix is of little use for such rules, because only the lesser part of memory is freed by `msc_pcre_cleanup`. As noted in #610, first leak also occurs when apache with `mod_security2` is reloaded (and so all the rules are freed and re-allocated again, but PCRE study data from the first config load remains unfreed for all regexes).

It should be noted that first leak is rather hard to trace as it can't be detected by most automated tools including Valgrind because PCRE uses it's own allocator for allocating memory for JIT, this allocator is not based on `malloc` and allocates memory by directly calling `mmap`, and it does not have annotations for Valgrind (`VALGRIND_MALLOCLIKE_BLOCK`/`VALGRIND_MEMPOOL_ALLOC`).

#### Reproducing the leaks

[This test program](https://gist.github.com/asterite3/4458835ec64b50ee980b16ac6b75e27d)  (using ModSecurity 2 built as a standalone module, code is mostly based on `standalone/main.cpp`) with [this minimal config](https://gist.github.com/asterite3/5dc749430f13fe6aa9a29d226cf74448) can be used to demonstrate the leaks. ([This config with more rules can be used instead, with it memory leaks more](https://gist.github.com/asterite3/73185d0f3fa7686a871a5da54e304f0c)). These test configs include rules 920420 and 920480 from OWASP CRS. To test it with PCRE JIT, ModSecurity build was configured with `./configure --with-curl  --with-yajl --enable-standalone-module --disable-mlogc --enable-pcre-jit --without-lua`. When run without the proposed fixes, it consumes a lot of RAM. Note that applying only one of the fixes does not solve the problem - using only the first fix does not change the situation at all (as `msc_pcre_cleanup` is not called), using only the second one reduces memory leak only slightly (as most of the memory allocated belongs to PCRE JIT data).

If PCRE  JIT is disabled (`--enable-pcre-jit` configure flag is omitted), memory leaks slower, but leak still occurs. In my tests, after 200000 requests test program reached RSS 145M when ModSecurity was compiled without PCRE JIT (with PCRE JIT RSS reached nearly 869M).

The other way to reproduce the leak is to use the same config with ModSecurity 2 running as Apache module and continuously sending requests to it - without the fixes memory usage of Apache workers can be seen to grow constantly, with the fixes memory usage of workers in my tests stayed nead 8 megabytes.

#### Fixes

The first leak is fixed by calling `pcre_free_study` on `pcre_extra` structure pointer instead of `free`/`pcre_free` in `msc_pcre_cleanup`. Function `msc_pregcomp_ex` that creates `msc_regex_t` can allocate `pcre_extra`  "manually" with `malloc`/`pcre_malloc` instead of with `pcre_study`. That's why the logic is added that check that `pcre_extra` indeed came from `pcre_study` by checking if `PCRE_EXTRA_EXECUTABLE_JIT` is set in `flags` of `pcre_study` (and `pcre_free_study` is only called in that case). `msc_pregcomp_ex` does not set that flag itself, and in fact `pcre_free_study` works the same way: it checks that flag, if it is not set, it calls `pcre_free` on `pcre_extra` pointer (`pcre_free` defaults for `free`, but can be set to other function by library user).
Please review this solution and tell if it's not the best one. Alternative options are, for example, to call `pcre_free_study` unconditionally or add an extra field to `msc_regex_t` that will indicate whether `pe` field was obtained by `pcre_study`.

The second leak is fixed by "attaching" allocated data to `msr->mp` instead of to `rule->ruleset->mp` . This seems safe because `msc_regex_t` created in `msre_op_rx_execute` seems to be stored nowhere and for subsequent requests a new `msc_regex_t` will be created with calls to `msc_pregcomp_ex`. Anyway, the review here would be appreciated too.

#### Similar leaks in other operators

It seems the same problem exists with `validateHash` https://github.com/SpiderLabs/ModSecurity/blob/12cefbd70f2aab802e1bff53c50786f3b8b89359/apache2/re_operators.c#L787 and similar case is with `gsbLookup` https://github.com/SpiderLabs/ModSecurity/blob/12cefbd70f2aab802e1bff53c50786f3b8b89359/apache2/re_operators.c#L1687 and below. 
I was not able to find the examples of usage of these operators in OWASP CRS or figure out how they work enough to make my own examples - to make working examples reproducing the leaks there.
Still, I can provide analogous fixes for them in this or separate pull request.

Fixes #610

### DIFF
--- a/apache2/msc_pcre.c
+++ b/apache2/msc_pcre.c
@@ -21,11 +21,7 @@
 static apr_status_t msc_pcre_cleanup(msc_regex_t *regex) {
     if (regex != NULL) {
         if (regex->pe != NULL) {
-#if defined(VERSION_NGINX)
-            pcre_free(regex->pe);
-#else
-            free(regex->pe);
-#endif
+            pcre_free_study(regex->pe);
             regex->pe = NULL;
         }
         if (regex->re != NULL) {
@@ -67,19 +63,15 @@ void *msc_pregcomp_ex(apr_pool_t *pool, const char *pattern, int options,
 
     #ifdef WITH_PCRE_STUDY
         #ifdef WITH_PCRE_JIT
-                pe = pcre_study(regex->re, PCRE_STUDY_JIT_COMPILE, &errptr);
+                pe = pcre_study(regex->re, PCRE_STUDY_EXTRA_NEEDED|PCRE_STUDY_JIT_COMPILE, &errptr);
         #else
-                pe = pcre_study(regex->re, 0, &errptr);
+                pe = pcre_study(regex->re, PCRE_STUDY_EXTRA_NEEDED, &errptr);
         #endif
     #endif
 
     /* Setup the pcre_extra record if pcre_study did not already do it */
     if (pe == NULL) {
-#if defined(VERSION_NGINX)
         pe = pcre_malloc(sizeof(pcre_extra));
-#else
-        pe = malloc(sizeof(pcre_extra));
-#endif
         if (pe == NULL) {
             return NULL;
         }
--- a/apache2/re_operators.c
+++ b/apache2/re_operators.c
@@ -1018,7 +1018,7 @@ static int msre_op_rx_execute(modsec_rec *msr, msre_rule *rule, msre_var *var, c
                 msr_log(msr, 6, "Escaping pattern [%s]",pattern);
             }
 
-            regex = msc_pregcomp_ex(rule->ruleset->mp, pattern, PCRE_DOTALL | PCRE_DOLLAR_ENDONLY, &errptr, &erroffset, msc_pcre_match_limit, msc_pcre_match_limit_recursion);
+            regex = msc_pregcomp_ex(msr->mp, pattern, PCRE_DOTALL | PCRE_DOLLAR_ENDONLY, &errptr, &erroffset, msc_pcre_match_limit, msc_pcre_match_limit_recursion);
             if (regex == NULL) {
                 *error_msg = apr_psprintf(rule->ruleset->mp, "Error compiling pattern (offset %d): %s",
                         erroffset, errptr);


apache2/msc_pcre.c
```diff
@@ -21,11 +21,15 @@
 static apr_status_t msc_pcre_cleanup(msc_regex_t *regex) {
     if (regex != NULL) {
         if (regex->pe != NULL) {
+            if (((pcre_extra *)regex->pe)->flags & PCRE_EXTRA_EXECUTABLE_JIT) {
```
WGH-: This's the code from ModSecurity v3. Is it correct? If yes, maybe it's better to just borrow it from there? It seems more straightforward.

https://github.com/SpiderLabs/ModSecurity/blob/357c14000350234beb017ea08f01dc7fed27ed4a/src/utils/regex.cc#L41-L66

apache2/msc_pcre.c
```diff
@@ -21,11 +21,15 @@
...
+                pcre_free_study(regex->pe);
+            } else {
 #if defined(VERSION_NGINX)
-            pcre_free(regex->pe);
+                pcre_free(regex->pe);
```
WGH-: Accidental whitespace change?

apache2/msc_pcre.c
asterite3: This is indeed more straightforward and effective (branching will happen at compile time instead of run time).
I did not do it this way initially because it (at least theoretically) can be so that `WITH_PCRE_STUDY` and `WITH_PCRE_JIT` **are** defined, and `pcre_study` **was** called, but returned `NULL` (man page for `pcre_study` says `If the function returns NULL, either it could not find any additional information, or there was an error. You can tell the difference by looking at the error value. It is NULL in first case.`). In this case `msc_pregcomp_ex` will allocate `pcre_extra` with `malloc`/`pcre_malloc`, it does so because it uses `pcre_extra` to set matching limits.
The relevant code is here

https://github.com/SpiderLabs/ModSecurity/blob/aa18ec7/apache2/msc_pcre.c#L68-L87

To simplify/sum it up: I am trying to avoid calling `pcre_free_study` on data not returned by `pcre_study` (but by `malloc`/`pcre_malloc`).

WGH-: But are you aware that you basically replicated the same (simple) check that's inside `pcre_free_study` itself? A function call to libpcre library is going to happen either way (`pcre_free_study` or `pcre_free`), so I don't think there's any performance gain.

https://vcs.pcre.org/pcre/code/trunk/pcre_study.c?revision=1639&view=markup

(extra ifdefs omitted for clarity)
```c
void pcre_free_study(pcre_extra *extra)
{
if (extra == NULL)
  return;
#ifdef SUPPORT_JIT
if ((extra->flags & PCRE_EXTRA_EXECUTABLE_JIT) != 0 &&
     extra->executable_jit != NULL)
  PRIV(jit_free)(extra->executable_jit);
#endif
PUBL(free)(extra);
}
```

asterite3: > But are you aware that you basically replicated the same (simple) check that's inside `pcre_free_study` itself

Yes, I've even written it in commit message and issue description myself:

> and in fact `pcre_free_study` works the same way: it checks that flag, if it is not set, it calls `pcre_free` on `pcre_extra` pointer 

> so I don't think there's any performance gain

I worry mostly about safety, not performance here (including trying to avoid calling deallocating function from one allocator on pointer produced by other allocator. PCRE allows to replace `pcre_malloc` and `pcre_free` with custom allocation functions, which is used, [for example, by nginx](https://github.com/nginx/nginx/blob/4bf4650f2f10f7bbacfe7a33da744f18951d416d/src/core/ngx_regex.c#L74)).

And I don't know how correct it is to rely on current implementation of `pcre_free_study`, to expect it to stay the same. 

Even safer option here might be also to not rely on `PCRE_EXTRA_EXECUTABLE_JIT` flag, but add a new field to `msc_regex_t`.
What do you think @zimmerle ?

WGH-: There're three cases:

1) The code is not using JIT, `VERSION_NGINX` is not defined. The `pcre_extra` struct is allocated with plain `malloc`, and has to be freed with plain `free` (technically, you can mix and match allocator functions if you don't touch the global variables `pcre_malloc` and `pcre_free`, but that obviously smells and is fragile).
2) The code is not using JIT, `VERSION_NGINX` is defined. Memory is allocated with `pcre_malloc`, and both `pcre_free` and `pcre_free_study` can be used.
3) The code is using JIT. In this case, the memory _has_ to be freed by `pcre_free_study`. Using `pcre_free` would cause memory leak, as JIT-compiled code woudn't be freed. 

The complexity comes from distinguishing cases 1 and 3. If we simply used `pcre_malloc` all the time, we could've just used `pcre_free_study`. 

The `ifdef` was introduced in 154e80750. I think that `ifdef` complicates things unnecessarily, and we can drop it and simply use `pcre_malloc`/`pcre_free`. After all, these two variables are just function pointers initialized to `malloc` and `free` correspondingly.

asterite3: I researched it a bit more and found this doc https://www.pcre.org/original/doc/html/pcreapi.html#SEC13 (which turned out to be a web version of `man pcreapi`). After reading it and after more discussion with @WGH- I got more convinced that it is worth changing the code a bit more to make the fix a bit cleaner and make some refactoring.

The doc says that it is recommended to replace usages of `pcre_free()` to deallocate `pcre_extra` with `pcre_free_study()`:
> When you are finished with a pattern, you can free the memory used for the study data by calling pcre_free_study(). This function was added to the API for release 8.20. For earlier versions, the memory could be freed with pcre_free(), just like the pattern itself. This will still work in cases where JIT optimization is not used, but it is advisable to change to the new function when convenient.

So I agree with @WGH-  that using regular `malloc` and `free` for `pcre_extra` seem to be of no use, and that it is correct and more clean to use `pcre_malloc` instead, and in this case `pcre_free_study` can be used to deallocate it unconditionally.

The doc also mentions there is a flag `PCRE_STUDY_EXTRA_NEEDED` that can be passed to `pcre_study()` to ask it to allocate `pcre_extra` anyway, even if studying `did not find any additional information` - which may be useful for us since `pcre_extra` is needed anyway to set match limits on it.

I created and alternative version of this branch with described approach,in fact only the second commit is changed, it can be found here https://github.com/seclab-msu/ModSecurity/commit/5859417502d3058599c0c84261db5a8fe886c7c1
That branch passes tests (run by `tests/run-unit-tests.pl`) as well as original branch of this PR, I also tested running modsec as Apache module and also built it as Nginx module and made some testing by running these web servers with modsecurity module and sending some requests to them - did not find any problems.
I am ready to replace branch of this pull request with that one if it's better @zimmerle

asterite3: I replaced the branch of this PR with aforementioned alternative version of the fix

marcstern:

marcstern:
Running in prod on 60+ servers for more than 1 year.
It definitely solves a technical problem that could be critical in some environments.
Why isn't it accepted?

marcstern (approved):
Running in prod on 60+ servers for more than 1 year.
It definitely solves a technical problem that could be critical in some environments.

martinhsv:
This PR has become out-of-date.

The change suggested for re_operators.c was made in Jan. 2021 via #2049 . This fix is available in v2.9.4.

That change addressed the more severe of the two issues described. (More severe, in my opinion, simply because macro-expansion patterns are likely to exist on many more installations than installations using pcre jit.)

Do you want to amend this PR (or replace it) to address only the separate JIT issue?

marcstern:
@martinhsv: you mean that the fix in re_operators.c is already included now, so you cannot merge this PR anymore. So you need a PR with only msc_pcre.c. Is that correct?

martinhsv:
Basically, yes.

(Although I'll note that a big part of my suggestion here is for purposes of clarity. Whether the new/revised PR gets merged in the near future or remains open for a more extended period, it's useful for it to be clear what change it represents -- rather than having it muddled with some other change that has been done previously.)

marcstern:
Incorporated into https://github.com/owasp-modsecurity/ModSecurity/pull/3154

Issue #2296, opened 2020-04-20
Selection of TX variables via RegEx should not be case-sensitive


michaelgranzow-avi:
**Description**

Behaviour of ModSec 2.9 and 3.x differs with respect to selection of elements in the 'TX' collection with regular expressions. In 2.9, the RegEx matches without taking case into account, in 3.x the match is case-sensitive.

**Logs and dumps**

Output of the modsec regression test tool with the first test case in the file below as input:
ModSecurity 3.0.4 - tests
(options are not available -- missing GetOpt)

  \# File Name                                         Test Name                                                             Passed?
\--- ---------                                         ---------                                                             -------
  1 tx-regex-var.json                                 Testing regex selection from TX - case                                failed!

Test failed. From: tx-regex-var.json.
Test name: Testing regex selection from TX - case.
Reason:
HTTP code mismatch. expecting: 437 got: 200

Debug log:
[1587380363] [] [4] Initializing transaction
[1587380363] [] [4] Transaction context created.
[1587380363] [] [4] Starting phase CONNECTION. (SecRules 0)
[1587380363] [] [9] This phase consists of 0 rule(s).
[1587380363] [] [4] Starting phase URI. (SecRules 0 + 1/2)
[1587380363] [/] [4] Starting phase REQUEST_HEADERS.  (SecRules 1)
[1587380363] [/] [9] This phase consists of 1 rule(s).
[1587380363] [/] [4] (Rule: 0) Executing unconditional rule...
[1587380363] [/] [4] Running [independent] (non-disruptive) action: setvar
[1587380363] [/] [8] Saving variable: TX:restricted_headers with value: /name1/name2/
[1587380363] [/] [9] Appending request body: 0 bytes. Limit set to: 0.000000
[1587380363] [/] [4] Starting phase REQUEST_BODY. (SecRules 2)
[1587380363] [/] [9] This phase consists of 1 rule(s).
[1587380363] [/] [4] (Rule: 500065) Executing operator "Rx" with param "^.*$" against REQUEST_HEADERS_NAMES.
[1587380363] [/] [9]  T (0) t:lowercase: "host"
[1587380363] [/] [9] Target value: "host" (Variable: REQUEST_HEADERS_NAMES:Host)
[1587380363] [/] [7] Added regex subexpression TX.0: host
[1587380363] [/] [9] Matched vars updated.
[1587380363] [/] [4] Running [independent] (non-disruptive) action: setvar
[...]

PR #2297, opened 2020-04-22
V3/collection re fix

airween:
This PR is the cleaned version of #2107.

I've updated my local branch and forked a new one from current v3/master. Contains only one necessary commit, changed the [variable name](https://github.com/SpiderLabs/ModSecurity/pull/2107#discussion_r412851155) for better readability. 

Edit: this PR fixes the #2296.


martinhsv:
A couple of observations:
1) What if someone wanted to write a rule where they did indeed want the TX:/[regex-content]/ to be case sensitive?  Wouldn't these code changes make that impossible?
2) Would your use case be satisfied by instead modifying the regex itself.  For example in your first test use something like 'TX:/^(?i)HEADER_NAME_/ ' ?

airween:
Hi @martinhsv,

>     1. What if someone wanted to write a rule where they did indeed want the TX:/[regex-content]/ to be case sensitive?  Wouldn't these code changes make that impossible?

If someone wants to do that, how can he/she does it in case of mod_security2? The expected mode is case insensitive (but the documentation doesn't contain it).

I think if you want to implement this feature, it would be more correct to introduce some new syntax (and make the relevant code).

>     2. Would your use case be satisfied by instead modifying the regex itself.  For example in your first test use something like 'TX:/^(?i)HEADER_NAME_/ ' ?

Yes, that works as well - but you can see there are two issues about this problem (#2296, #1808) and one PR (#1810), so I think the users accustomed this form.

martinhsv:
Hi @airween ,

First off, apologies for the delay.

I previously expressed some reservations about the change.  I have put some more thought into this and am persuaded that this is the right thing to do.

A couple of reasons:
- Given that exact names in the VARIABLES portion of a rule (e.g. ARGS:myid) are treated as case insensitive, making the regex equivalent (ARGS:/myid/) behave likewise is good for consistency
- this change probably doesn't reduce user flexibility as much as I had feared.  For example, in PCRE if a rule writer specifically wants to make the test case sensitive, one can prefix the main portion of the pattern with (?i-).  (Not to mention that the cases where one has a compelling reason to do this are likely rare.)

(We also wanted to check that some futures features, such as supporting PCRE alternatives, would be unlikely to be hampered by this change.)

@zimmerle and I discussed this earlier today, and I don't think we have any further concerns.

This should get merged within a few days.

airween:
Hi @martinhsv,

> This should get merged within a few days.

Thanks, it's good to read :).

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
+[
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"OWASP CRS id:920450",
```
zimmerle: I would prefer to have a title with a more meaningful message. If the test "**OWASP CRS id:920450**" fails, what does it means to the user?

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+      "SecRuleEngine On",
+      "SecRule &TX:restricted_headers \"@eq 0\" \"id:901165,phase:1,pass,nolog,setvar:'tx.restricted_headers=/proxy/ /lock-token/ /content-range/ /translate/ /if/'\"",
+      "SecRule REQUEST_HEADERS_NAMES \"@rx ^.*$\" \"id:920450,phase:2,capture,block,deny,t:lowercase,setvar:'tx.header_name_%{tx.0}=/%{tx.0}/',chain",
+      "SecRule TX:/^HEADER_NAME_/ \"@within %{tx.restricted_headers}\" \"\""
+    ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"OWASP CRS id:920450",
```
zimmerle: Same here.

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
+[
```
zimmerle: The file name for issues follow the pattern - issue-number.json.

We do have some **rule-**, but we better keep the issue related to it, so the user will able to link the test with the correspondent background information on it.

The issue number is #2296 opened by @michaelgranzow-avi

Makefile.am
```diff
@@ -185,6 +185,7 @@ TESTS+=test/test-cases/regression/request-body-parser-xml-validade-dtd.json
 TESTS+=test/test-cases/regression/rule-920120.json
 TESTS+=test/test-cases/regression/rule-920200.json
 TESTS+=test/test-cases/regression/rule-920274.json
+TESTS+=test/test-cases/regression/rule-920450.json
```
zimmerle: I would expect a different name for that file. Explained below.

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+  {
+    "enabled":1,
```
zimmerle: Since this is a pull request that was well described in a issue i miss the **github_issue** tag.

As listed -
https://github.com/SpiderLabs/ModSecurity/blob/785958f9b5089b918c7d054cbcc2fe4a3c7b3788/test/test-cases/regression/issue-960.json#L7-L9

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "http_code":200
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule &TX:restricted_headers \"@eq 0\" \"id:901165,phase:1,pass,nolog,setvar:'tx.restricted_headers=/proxy/ /lock-token/ /content-range/ /translate/ /if/'\"",
```
zimmerle: There is a value of having complex rules on the **regression tests** as it increases surface area for testing features. However, I am afraid that together with the complex example, we need to have a very simple use case scenario. If/When this test fails, it was a consequence of a problem on the selection of variables using regular expression or elsewhere (e.g. setvar)?

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+      "ip":"200.249.12.31",
+      "port":123
+    },
+    "server":{
+      "ip":"200.249.12.31",
+      "port":80
+    },
+    "request":{
+      "headers":{
+        "TRANSLATE": "test"
```
zimmerle: I would prefer to use real request headers, to mimic a real world example.

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "http_code":403
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule &TX:restricted_headers \"@eq 0\" \"id:901165,phase:1,pass,nolog,setvar:'tx.restricted_headers=/proxy/ /lock-token/ /content-range/ /translate/ /if/'\"",
```
zimmerle: Same rule complexity issue explained below.

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
+      "headers":{
+        "Date":"Mon, 13 Jul 2015 20:02:41 GMT",
+        "Last-Modified":"Sun, 26 Oct 2014 22:33:37 GMT",
+        "Content-Type":"text/html"
+      },
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
```
zimmerle: Specially because the rules are complex, this test case could fail silent... as a combination of two (or more) failures. Better to also considers the **debug** and **error** log (when possible).

test/test-cases/regression/rule-920450.json
```diff
@@ -0,0 +1,87 @@
...
```
zimmerle: Same comment above. Use error and debug logs.

zimmerle (commented):
I have read the code and made some comments.

I am going the make the modifications by myself as this is going back and forth from a while. We want to have a fix for it.

Vide: #2297, #2296, and #2107.

@airween let me know if you have any doubts or concerns on my comments.

test/test-cases/regression/rule-920450.json
airween: Changed in [ceb0e3](https://github.com/SpiderLabs/ModSecurity/pull/2297/commits/ceb0e381d0338442b6d4761c9847630202611673#diff-e641572b2bd6ff4e44759e3dea1d59cfe9475a364258bdfac6c25628efea2bd1R5)

test/test-cases/regression/rule-920450.json
airween: I changed it in my last commit ([ceb0e3](https://github.com/SpiderLabs/ModSecurity/pull/2297/commits/ceb0e381d0338442b6d4761c9847630202611673)).

Just a small note: this issue (#2296) was reported by @michaelgranzow-avi, but I sent the first PR in June of 2019 (#2107) where I described this issue.

Makefile.am
airween: Fixed.

test/test-cases/regression/rule-920450.json
airween: It's fixed [too](https://github.com/SpiderLabs/ModSecurity/pull/2297/commits/ceb0e381d0338442b6d4761c9847630202611673#diff-e641572b2bd6ff4e44759e3dea1d59cfe9475a364258bdfac6c25628efea2bd1R49).

test/test-cases/regression/rule-920450.json
airween: Thanks for info. The last commit contains both 'url' and 'github_issue' keys, in both test cases.

Note: it would be good to make some wiki page about your expectations for contributors. I'm sure these keywords are part of the regression test structure, but after 17 commits I've met with them first.

test/test-cases/regression/rule-920450.json
airween: I changed the rules in the new regression test file. I based @michaelgranzow-avi test cases, but there are only two cases, with same context: only the TX variable name format differ.

test/test-cases/regression/rule-920450.json
airween: Changed.

test/test-cases/regression/rule-920450.json
airween: Fixed.

test/test-cases/regression/rule-920450.json
airween: I changed the test cases: both of them block the request, and gives the error_log, which used in "expected" block.

test/test-cases/regression/rule-920450.json
airween: Fixed too.

airween:
> @airween let me know if you have any doubts or concerns on my comments.

See my answers and the new commit: [ceb0e38](https://github.com/SpiderLabs/ModSecurity/pull/2297/commits/ceb0e381d0338442b6d4761c9847630202611673).

zimmerle:
Closing this as consequence of 910a1877a9ec673f36c139e6a3e941812b8e986e. Thanks everybody involved on this issue resolution.

PR #2304, opened 2020-05-01
Support reopening audit logs

### LINKED ISSUE #1968: Add interface in libModSecurity for reopening log files
This is related with feature request https://github.com/SpiderLabs/ModSecurity-nginx/issues/121

> Modsecurity should reopen audit log on these two signals for proper logrotate operation.

As noted at https://github.com/SpiderLabs/ModSecurity-nginx/issues/121#issuecomment-420619429, we could leverage a similar approach as described at https://forum.nginx.org/read.php?29,247488,247500#msg-247500 (i.e. use standard nginx API to open some stub-file with ngx_conf_open_file(), add required handler, and use it for detecting USR1 and HUP signals from master process)

But it seems like libModSecurity currently does not have a nice interface to initiate audit/debug log files reopening by connector's request.

There's a PoC on how we could accomplish that on the connector at https://github.com/SpiderLabs/ModSecurity-nginx/issues/121#issuecomment-442416602 as a starting point.

brandonpayton:
This PR adds support for reopening log files by:

* Exposing an `msc_rules_reopen_audit_log(rules, error)` function to trigger audit log reopen for a given rules set
* Adding a `reopen(error)` method to `AuditLog`
* Adding a `reopen(error)` method to the audit log `Writer` interface
* Adding a `reopen(filename, error)` method to `SharedFiles`

⚠️ Currently, I do not believe `SharedFiles::reopen()` to be thread-safe, and thread-safety must be ensured by the caller. Even if the SharedFiles general lock could be used, we may not want to lock for all SharedFile writes. Please let me know if you'd like to resolve the issue as part of this PR or a later PR.

This is based on work to support ModSecurity audit log rotation within [Automattic](https://automattic.com/). It is used in a ModSecurity-nginx patch, and the PR for that is [here](https://github.com/SpiderLabs/ModSecurity-nginx/pull/198).

resolves #1968

### DIFF
--- a/headers/modsecurity/audit_log.h
+++ b/headers/modsecurity/audit_log.h
@@ -168,6 +168,7 @@ class AuditLog {
     bool setType(AuditLogType audit_type);
 
     bool init(std::string *error);
+    bool reopen(std::string *error);
     virtual bool close();
 
     bool saveIfRelevant(Transaction *transaction);
--- a/headers/modsecurity/rules_set.h
+++ b/headers/modsecurity/rules_set.h
@@ -104,6 +104,7 @@ int msc_rules_add_remote(RulesSet *rules, const char *key, const char *uri,
 int msc_rules_add_file(RulesSet *rules, const char *file, const char **error);
 int msc_rules_add(RulesSet *rules, const char *plain_rules, const char **error);
 int msc_rules_cleanup(RulesSet *rules);
+int msc_rules_reopen_audit_log(RulesSet *rules, const char **error);
 
 #ifdef __cplusplus
 }
--- a/src/audit_log/audit_log.cc
+++ b/src/audit_log/audit_log.cc
@@ -372,6 +372,13 @@ bool AuditLog::merge(AuditLog *from, std::string *error) {
     return init(error);
 }
 
+bool AuditLog::reopen(std::string *error) {
+    if (m_writer != NULL) {
+        return m_writer->reopen(error);
+    }
+    return true;
+}
+
 
 }  // namespace audit_log
 }  // namespace modsecurity
--- a/src/audit_log/writer/https.h
+++ b/src/audit_log/writer/https.h
@@ -42,6 +42,9 @@ class Https : public Writer {
     bool init(std::string *error) override;
     bool write(Transaction *transaction, int parts,
         std::string *error) override;
+    bool reopen(std::string *error) override {
+        return true;
+    }
 };
 
 }  // namespace writer
--- a/src/audit_log/writer/parallel.cc
+++ b/src/audit_log/writer/parallel.cc
@@ -191,6 +191,38 @@ bool Parallel::write(Transaction *transaction, int parts, std::string *error) {
     return true;
 }
 
+
+bool Parallel::reopen(std::string *error) {
+    bool        success1 = true;
+    bool        success2 = true;
+    std::string error1;
+    std::string error2;
+
+    if (!m_audit->m_path1.empty()) {
+        success1 = utils::SharedFiles::getInstance().reopen(m_audit->m_path1, &error1);
+    }
+    if (!m_audit->m_path2.empty()) {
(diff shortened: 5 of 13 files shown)


brandonpayton:
This is failing cppcheck. Will take a look and fix.

brandonpayton:
Actually, the warning appears to be unrelated to this patch:

> warning: [src/utils/shared_files.h,88](https://github.com/SpiderLabs/ModSecurity/blob/7e0bc2691727b8c75f74638cdc4d1c45a689a7b6/src/utils/shared_files.h#L79-L87),performance,useInitializationList,When an object of a class is created, the constructors of all member variables are called consecutively in the order the variables are declared, even if you don't explicitly write them to the initialization list. You could avoid assigning 'm_memKeyStructure' a value by passing the value to the constructor in the initialization list.

brandonpayton:
Sorry, I was incorrect. The previous warning is the only warning I'm seeing on my dev machine.

There is a relevant warning from TravisCI running `make check-static`:
> warning: src/utils/shared_files.h,60,style,functionConst,The member function 'modsecurity::utils::SharedFiles::reopen' can be made a const function. Making this function 'const' should not cause compiler errors. Even though the function can be made const function technically it may not make sense conceptually. Think about your design and the task of the function first - is it a function that must not change object internal state?

brandonpayton:
I added the above warning to the cppcheck suppressions list because `SharedFiles::reopen()` does change SharedFiles internal state.

brandonpayton:
Regarding thread-safety:

I was reading the nginx source this evening and noticed it redirects stderr using a POSIX function, `dup2`. `dup2` takes an old and a new descriptor and atomically retargets the new to refer to the old. As evidence, both [glibc docs](https://www.gnu.org/software/libc/manual/html_node/Duplicating-Descriptors.html) and [POSIX docs](https://pubs.opengroup.org/onlinepubs/9699919799/functions/dup.html) (requires login) state that dup2 should be atomic.

This looks like a good way to address the thread-safety issue with the current SharedFiles::reopen() in this PR.

MS [also provides](https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/posix-dup-dup2?view=vs-2019) an implementation of dup2 for POSIX compliance, and given that the POSIX docs require atomicity, it seems like this should be thread-safe on Windows as well.

JokerQyou:
What's the current status of this PR?

mike-mckenna:
Is this pull request getting attention?  The code in this pull request will support the rotation of the audit log in Nginx.  More specifically, the code in this pull request is required before ModSecurity-nginx pull request https://github.com/SpiderLabs/ModSecurity-nginx/pull/198 can proceed.

liudongmiao:
There should be a reopen for debug log.

I'd like to use a flush on `modsecurity::utils::SharedFiles` directly.

tomsommer:
Any chance of getting this fixed? Sort of annoying not being able to reliably rotate logs

brandonpayton:
As @liudongmiao mentioned [here](https://github.com/SpiderLabs/ModSecurity/pull/2304#issuecomment-1280130773), this does not reopen the debug log. It would be good to generalize this PR to reopen logs in general.

I don't have time to work on this immediately but hope to get back to it in the next month or so. If anyone else wants to work on this before then, please feel free.

ghost:
> As @liudongmiao mentioned [here](https://github.com/SpiderLabs/ModSecurity/pull/2304#issuecomment-1280130773), this does not reopen the debug log. It would be good to generalize this PR to reopen logs in general.
> 
> I don't have time to work on this immediately but hope to get back to it in the next month or so. If anyone else wants to work on this before then, please feel free.

Any news on this?

airween:
Hi @Orgoth,

thanks for bringing this up - I try to take a look at this soon.

akama-aka:
Just tried to apply the patch but it failed. I inserted the code my self lets see if it works. Thank you for your help airween ^^

Well it didn't work

PR #2368, opened 2020-07-24
Remove unnecessary copying in transformations

WGH-:
In C++11, string data is always null-terminated[1], and can be freely modified[2].

[1] https://stackoverflow.com/questions/6077189/will-stdstring-always-be-null-terminated-in-c11
[2] https://stackoverflow.com/questions/38875623/is-it-permitted-to-modify-the-internal-stdstring-buffer-returned-by-operator

### DIFF
--- a/src/actions/transformations/utf8_to_unicode.cc
+++ b/src/actions/transformations/utf8_to_unicode.cc
@@ -36,21 +36,10 @@ namespace transformations {
 std::string Utf8ToUnicode::evaluate(const std::string &value,
     Transaction *transaction) {
     std::string ret;
-    unsigned char *input;
     int changed = 0;
     char *out;
 
-    input = reinterpret_cast<unsigned char *>
-        (malloc(sizeof(char) * value.length()+1));
-
-    if (input == NULL) {
-        return "";
-    }
-
-    memcpy(input, value.c_str(), value.length()+1);
-
-    out = inplace(input, value.size() + 1, &changed);
-    free(input);
+    out = inplace(reinterpret_cast<const unsigned char *>(&value[0]), value.size() + 1, &changed);
     if (out != NULL) {
         ret.assign(reinterpret_cast<char *>(out),
             strlen(reinterpret_cast<char *>(out)));
@@ -61,7 +50,7 @@ std::string Utf8ToUnicode::evaluate(const std::string &value,
 }
 
 
-char *Utf8ToUnicode::inplace(unsigned char *input,
+char *Utf8ToUnicode::inplace(const unsigned char *input,
     uint64_t input_len, int *changed) {
     unsigned int count = 0;
     char *data;
@@ -89,7 +78,7 @@ char *Utf8ToUnicode::inplace(unsigned char *input,
         int unicode_len = 0;
         unsigned int d = 0;
         unsigned char c;
-        unsigned char *utf = (unsigned char *)&input[i];
+        const unsigned char *utf = &input[i];
 
         c = *utf;
 
--- a/src/actions/transformations/css_decode.cc
+++ b/src/actions/transformations/css_decode.cc
@@ -37,16 +37,11 @@ namespace transformations {
 std::string CssDecode::evaluate(const std::string &value,
     Transaction *transaction) {
 
-    char *tmp = reinterpret_cast<char *>(
-        malloc(sizeof(char) * value.size() + 1));
-    memcpy(tmp, value.c_str(), value.size() + 1);
-    tmp[value.size()] = '\0';
+    std::string ret = value;
 
-    CssDecode::css_decode_inplace(reinterpret_cast<unsigned char *>(tmp),
-        value.size());
+    auto size = CssDecode::css_decode_inplace(reinterpret_cast<unsigned char *>(&ret[0]), ret.size());
+    ret.resize(size);
 
-    std::string ret(tmp, 0, value.size());
-    free(tmp);
     return ret;
 }
 
--- a/src/actions/transformations/escape_seq_decode.cc
+++ b/src/actions/transformations/escape_seq_decode.cc
@@ -142,17 +142,10 @@ int EscapeSeqDecode::ansi_c_sequences_decode_inplace(unsigned char *input,
 
 std::string EscapeSeqDecode::evaluate(const std::string &value,
     Transaction *transaction) {
+    std::string ret = value;
 
-    unsigned char *tmp = (unsigned char *) malloc(sizeof(char)
-        * value.size() + 1);
-    memcpy(tmp, value.c_str(), value.size() + 1);
-    tmp[value.size()] = '\0';
-
-    int size = ansi_c_sequences_decode_inplace(tmp, value.size());
-
-    std::string ret("");
-    ret.assign(reinterpret_cast<char *>(tmp), size);
-    free(tmp);
+    auto size = ansi_c_sequences_decode_inplace(reinterpret_cast<unsigned char *>(&ret[0]), ret.size());
+    ret.resize(size);
 
     return ret;
 }
--- a/src/actions/transformations/hex_decode.cc
+++ b/src/actions/transformations/hex_decode.cc
@@ -34,23 +34,10 @@ namespace transformations {
 
 std::string HexDecode::evaluate(const std::string &value,
     Transaction *transaction) {
-    std::string ret;
-    unsigned char *input;
-    int size = 0;
+    std::string ret = value;
 
-    input = reinterpret_cast<unsigned char *>
-        (malloc(sizeof(char) * value.length()+1));
-
-    if (input == NULL) {
-        return "";
-    }
-
-    memcpy(input, value.c_str(), value.length()+1);
-
-    size = inplace(input, value.length());
-
-    ret.assign(reinterpret_cast<char *>(input), size);
-    free(input);
+    auto size = inplace(reinterpret_cast<unsigned char *>(&ret[0]), ret.length());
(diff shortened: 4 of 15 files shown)


zimmerle:
Thanks @WGH-. Interesting proposal.

Indeed, as mentioned in [[2]](https://stackoverflow.com/questions/38875623/is-it-permitted-to-modify-the-internal-stdstring-buffer-returned-by-operator) and as we are using c++11, it is possible to access the string buffer by array idx, and the buffer is stored contiguously.

As of our [development branch](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/), we have added the concept of _ModSecString_. It is not yet 100% ready, but the idea is to have an allocator to use a stack-based memory pool for each request, avoiding memory fragmentation. That **stack allocator** should be aware that space should be allocated contiguously.

I am going to review it better and get back to you soon.


Example of ModSecString -
https://github.com/SpiderLabs/ModSecurity/blob/5824470768e73d438a5db03b7a6eb24d7c795097/src/actions/transformations/css_decode.cc#L32-L35

WGH-:
> As of our development branch, we have added the concept of ModSecString. It is not yet 100% ready, but the idea is to have an allocator to use a stack-based memory pool for each request, avoiding memory fragmentation. That stack allocator should be aware that space should be allocated contiguously.

Basically, you want to bring back region-based memory management, which old ModSecurity used with help of `apr_pool_t` and friends?

zimmerle:
> > As of our development branch, we have added the concept of ModSecString. It is not yet 100% ready, but the idea is to have an allocator to use a stack-based memory pool for each request, avoiding memory fragmentation. That stack allocator should be aware that space should be allocated contiguously.
> 
> Basically, you want to bring back region-based memory management, which old ModSecurity used with help of `apr_pool_t` and friends?

Not really, the **apr_pool** was used everywhere. For some pieces of the code where the strings are pre-computed as the request are processed (ARGS and QUERY_STRING), it makes sense to avoid memory fragmentation. The conjunction of **std::string_view** and **stacked** (_not dynamic_) memory usage could be very useful performance-wise. That is different from having dynamically allocated memory pools. It is not about having a generic memory pool but use the resources more wisely.

WGH-:
You're suggesting to use the call stack for variables associated with particular transaction?

WGH-:
This optimization has very small performance improvement: increase from 117-118 requests per second to 121-122 RPS. It might start to matter more once other parts are optimized, but right now it's mostly code cleanup.

zimmerle:
> You're suggesting to use the call stack for variables associated with particular transaction?

Let me illustrate that with an example - 

Lets assume that we have a request at the URI: ```/modsec-is-cool/?foo=bar&github=fine&a=b```
 [[test case on gits](https://gist.github.com/zimmerle/fc84d0ddf77b02ef7f892a44d601d1f6)]

By the time that we achieve a given phase ([request life cycle](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#processing-phases)) the variables will be populated internally in modsec. To exemplifies, let's pick: [ARGS](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#ARGS), [ARGS_GET](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#ARGS_GET), [QUERY_STRING](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#QUERY_STRING), [REQUEST_URI](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#REQUEST_URI), [REQUEST_URI_RAW](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#REQUEST_URI_RAW) they will be populated as such:

#### ARGS
ARGS:foo :: bar
ARGS:github :: **fine**
ARGS:a :: b


#### ARGS_GET
ARGS:foo :: bar
ARGS:github :: **fine**
ARGS:a :: b


#### QUERY_STRING
QUERY_STRING :: foo=bar&github=**fine**&a=b


#### REQUEST_URI
REQUEST_URI :: /modsec-is-cool/?foo=bar&github=**fine**&a=b


#### REQUEST_URI_RAW
REQUEST_URI_RAW :: /modsec-is-cool/?foo=bar&github=**fine**&a=b


Notice that the content of variables (and keys), repeats. Using the example of the word **fine**, it is allocated for each of those variables. In this example, **fine** could be allocated for **QUERY_STRING**, and mapped using ([string_view](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/headers/modsecurity/string_view.hpp)) for the other variables. Reducing memory usage and avoid extra malloc calls, consequently reducing memory fragmentation.

In that described use case, the string_view for each variable will be a part of the Transaction class, that could be stacked allocated as - although dynamically - the parameters do not vary on size: offset (void *) and length (uint?).

The above is just an example to illustrate a rationally; there are some corner cases.

There is usage for a memory pool, especially in terms of disposable memory used in serialized operations such as transformations. First, let's see if it is possible to avoid unnecessary allocations, as you are suggesting in this very own patch.

zimmerle:
> This optimization has very small performance improvement: increase from 117-118 requests per second to 121-122 RPS. It might start to matter more once other parts are optimized, but right now it's mostly code cleanup.

At a certain point, the improvements sound to be numerically inefficient. However, placing it in scale, they are not bad. The benefit here is in the order of magnitude of 3% ~ 4%. Not bad. In a busy server, I am sure that it does make a difference.

More to come, as the code became cleaner, it easy to spot an improve what needs to be improved :)

zimmerle:
Not worked with that yet as I am fixing the **xmlns** action on the _v3.1 branch_. The faulty test case is very annoying.

zimmerle:
> Not worked with that yet as I am fixing the **xmlns** action on the _v3.1 branch_. The faulty test case is very annoying.

**xmlns** is now fixed. Going to fix the _cppcheck warnings_ on **v3/master**. Afterward, I will proceed with the rebase for the **v3.1-experimental** branch. And finally, merge this.

The issue regarding the **xmlns** was a missing attribution for the rule variable. Almost all variables work without the knowledge of the rule that it belongs to, except for two variables: [RULE](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#RULE), and [XML](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#xml).
The variable [RULE](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#RULE) needs to access the metadata (id, etc...) about the **Rule** while the [XML](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#xml) demands to enumerates all [xmlns](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#xmlns) actions in order to correct address the XML namespace.

During the **Rule** instance copy, the variable was being copied correctly. But, the reference to the **Rule** (inside the variable instance) was remained unchanged, point towards the old instance. In the case of **xmlns** the old **Rule** was deleted and the reference was being pointed to an invalid location hence the crash.

### Relevant pieces of code

##### Base class common to XML and RULE
https://github.com/SpiderLabs/ModSecurity/blob/2ddb4ab97229578ae9073627c11ea616639e53ff/src/variables/rule_variable.h#L31-L54

##### Populating all Variables with base type RuleVariable during a Rule instance copy
https://github.com/SpiderLabs/ModSecurity/blob/2ddb4ab97229578ae9073627c11ea616639e53ff/src/rule_with_operator.h#L51-L62

##### XML Variable
https://github.com/SpiderLabs/ModSecurity/blob/2ddb4ab97229578ae9073627c11ea616639e53ff/src/variables/xml.h#L66-L84

WGH-:
I think most old-school pointers can be safely replaced with `std::unique_ptr`. It has zero runtime overhead, and it would make it much easier to reason about lifetimes, and would prevent such mistakes.

zimmerle:
> I think most old-school pointers can be safely replaced with `std::unique_ptr`. It has zero runtime overhead, and it would make it much easier to reason about lifetimes, and would prevent such mistakes.

In this particular case, we are using a shared_ptr. The idea is that Rules could belong to different hosts and so its variables. The variables that do not demand specific processing on execution time, are in practice anchors to offsets on the Transaction object. Safe to share among different Rules.

src/actions/transformations/utf8_to_unicode.cc
```diff
@@ -61,7 +50,7 @@ std::string Utf8ToUnicode::evaluate(const std::string &value,
 }
 
 
-char *Utf8ToUnicode::inplace(unsigned char *input,
+char *Utf8ToUnicode::inplace(const unsigned char *input,
```
zimmerle, line 53: Why not use the in-place method here?

zimmerle: I see, apart from the name the function is not really in place.

zimmerle:
I have **rebased** and **applied** this PR at the top of [v3.1-experimental](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental). Commit: 0cc0b929c36374084e2e607673ab88e29c848fb0

Some changes were placed as a consequence of the rebase.

zimmerle:
It is now - 5226c42f7c84bf138305d596b9d07f88c57fad46

zimmerle:
Merged! Thanks!

PR #2369, opened 2020-07-24
Make all "rule id" variables of type RuleId

WGH-:
Previously, ModSecurity inconsistently used int, int64_t and double for rule id variables in different places.

### DIFF
--- a/headers/modsecurity/rules_exceptions.h
+++ b/headers/modsecurity/rules_exceptions.h
@@ -28,6 +28,8 @@
 #include <memory>
 #endif
 
+#include "modsecurity/modsecurity.h"
+
 #ifndef HEADERS_MODSECURITY_RULES_EXCEPTIONS_H_
 #define HEADERS_MODSECURITY_RULES_EXCEPTIONS_H_
 
@@ -51,9 +53,9 @@ class RulesExceptions {
     ~RulesExceptions();
 
     bool load(const std::string &data, std::string *error);
-    bool addRange(int a, int b);
-    bool addNumber(int a);
-    bool contains(int a);
+    bool addRange(RuleId a, RuleId b);
+    bool addNumber(RuleId a);
+    bool contains(RuleId a);
     bool merge(RulesExceptions *from);
 
     bool loadRemoveRuleByMsg(const std::string &msg, std::string *error);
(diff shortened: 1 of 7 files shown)


zimmerle:
Hi, @WGH-

Thank you for the patch. Apparently we have a similar patch on our development branch -
 - b8aae970158252800862b566216c0b4d54220f34

The dev branch is available here:  [v3/dev/3.1-experimental](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental)

WGH-:
Although similar in scope, that patch merely replaces `int64_t` with `RuleId` typedef, and it doesn't address `int` and `double` stuff.

zimmerle:
I am not saying that one replaces the other. I am saying that both adds a benefit. It would be interesting to have your patch using the RuleId alias. Further, we can adjust the RuleId container size as needed. How do you fill about that?

WGH-:
I can rebase the patch on top of v3/dev/3.1-experimental, if that's what you're saying.

By the way, there's a type called `RuleId` here: https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/actions/rule_id.h. Wouldn't it create conflict with the newly introduced `RuleId` type alias?

> Further, we can adjust the RuleId container size as needed. How do you fill about that?

I'm not sure which RuleId container you're referring to.

zimmerle:
There are: 
 - modsecurity::actions::RuleId
 - modsecurity::RuleId

Those are defined here:

https://github.com/SpiderLabs/ModSecurity/blob/5824470768e73d438a5db03b7a6eb24d7c795097/src/actions/rule_id.h#L32-L35

and

https://github.com/SpiderLabs/ModSecurity/blob/5824470768e73d438a5db03b7a6eb24d7c795097/headers/modsecurity/modsecurity.h#L98

Those are two different things, thanks for the [namespace](https://en.cppreference.com/w/cpp/language/namespace)

I meant a rebase, making sure that the type **RuleId** is used to refer to a _Rule Id_. Not **int_64**, not **double**. Only **RuleId**.

WGH-:
Fair. I totally forgot about namespaces. I'll do a rebase on top of v3/dev/3.1-experimental then.

zimmerle:
@WGH- thank you! That is an important change.

WGH-:
Done. Some tests don't pass, but they don't pass in base v3/dev/3.1-experimental either.

zimmerle:
> Done. Some tests don't pass, but they don't pass in base v3/dev/3.1-experimental either.

Thanks! Yep. You are likely to be referring to the XML tests. Have to fix that and rebase the tree accordingly. It is on my todo list, before making it w3/master.

zimmerle:
Thank you @WGH-  this is now merged into 3.1-experimental.

PR #2370, opened 2020-07-24
Use 'equal_range' instead of full scan for rule exceptions

WGH-:


zimmerle:
As of [3.1-experimental](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental) the majority of the [UpdateActionBy?](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#SecRuleUpdateActionById) are pre-computed. Some parts of the change in this PR no longer exist in [3.1-experimental](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental). I have made some modifications taking the approach suggested in the patch. Published here: b32182940d558db803c063575552621bc061380b

zimmerle:
Merged! Thanks!

PR #2375, opened 2020-07-28
Optimize returning of transformation results

WGH-:
There's no point in using `shared_ptr`s here, because the strings are never reused (except for the rare case of multiMatch).
Likewise, `std::list` has considerable overhead which goes wasted, because there's no random insertions or iterators needed to be preserved after modifications.

In my test, this leads to increase from 117 requests per second to 127 RPS.

### DIFF
--- a/headers/modsecurity/rule.h
+++ b/headers/modsecurity/rule.h
@@ -52,9 +52,8 @@ namespace operators {
 class Operator;
 }
 
-using TransformationResult = std::pair<std::shared_ptr<std::string>,
-    std::shared_ptr<std::string>>;
-using TransformationResults = std::list<TransformationResult>;
+using TransformationResult = std::pair<std::string, std::string>;
+using TransformationResults = std::vector<TransformationResult>;
 
 using Transformation = actions::transformations::Transformation;
 using Transformations = std::vector<Transformation *>;
--- a/headers/modsecurity/rule_with_actions.h
+++ b/headers/modsecurity/rule_with_actions.h
@@ -72,10 +72,10 @@ class RuleWithActions : public Rule {
 
     inline void executeTransformation(
         actions::transformations::Transformation *a,
-        std::shared_ptr<std::string> *value,
+        std::string &value,
         Transaction *trans,
-        TransformationResults *ret,
-        std::string *path,
+        TransformationResults &ret,
+        std::string &path,
         int *nth) const;
 
 
--- a/headers/modsecurity/rule_with_operator.h
+++ b/headers/modsecurity/rule_with_operator.h
@@ -56,7 +56,7 @@ class RuleWithOperator : public RuleWithActions {
         variables::Variables *eclusion, Transaction *trans);
 
     bool executeOperatorAt(Transaction *trasn, const std::string &key,
-        std::string value, std::shared_ptr<RuleMessage> rm);
+        const std::string &value, std::shared_ptr<RuleMessage> rm);
 
     static void updateMatchedVars(Transaction *trasn, const std::string &key,
         const std::string &value);
--- a/src/rule_with_actions.cc
+++ b/src/rule_with_actions.cc
@@ -301,28 +301,28 @@ void RuleWithActions::executeAction(Transaction *trans,
 
 inline void RuleWithActions::executeTransformation(
     actions::transformations::Transformation *a,
-    std::shared_ptr<std::string> *value,
+    std::string &value,
     Transaction *trans,
-    TransformationResults *ret,
-    std::string *path,
+    TransformationResults &ret,
+    std::string &path,
     int *nth) const {
 
-    std::string *oldValue = (*value).get();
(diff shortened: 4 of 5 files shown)


zimmerle:
Hi, @WGH-

That modification was already contemplated on [our development branch](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental).

As listed -

The type **TransformationResults** declared here:
https://github.com/SpiderLabs/ModSecurity/blob/5824470768e73d438a5db03b7a6eb24d7c795097/src/rule_with_actions.h#L110-L112

Those are further used here:
https://github.com/SpiderLabs/ModSecurity/blob/5824470768e73d438a5db03b7a6eb24d7c795097/src/rule_with_actions.h#L242-L256

For bug fixing is usually ok to use the [v3/master](https://github.com/SpiderLabs/ModSecurity/tree/v3/master). For changes like optimization and internal structure modifications, I would suggest using the [v3.1-experimental](https://github.com/SpiderLabs/ModSecurity/tree/v3/dev/3.1-experimental) branch.
 
I also would like to invite you to discuss structural changes before propose fixes. Sometimes there are [Technical debts](https://en.wikipedia.org/wiki/Technical_debt) that we have to live with for a limited time, as more significant changes will fix a given issue as collateral. You can ping @zimmerle (me :P), @martinhsv, @victorhora, @afoxdavidi, we can gather together and discuss how to tackle the issues. We do regular meetings we can have you in.

In that particular example, the class Rule is being splied into sub-classes: [RuleWithActions](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/rule_with_actions.h), [RuleScript](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/rule_script.h), [RuleUnconditional](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/rule_unconditional.cc), [RuleWithOperator](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/rule_with_operator.h), [RuleMarker](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/src/rule_marker.h), in a near further we are going to have [Rule](https://github.com/SpiderLabs/ModSecurity/blob/v3/dev/3.1-experimental/headers/modsecurity/rule.h) split into Rule and ChainedRule. As Rule class became split, the scope of the variable value natuarraly became reduce, therefore the shared_ptr got removed.

The iterator over the results is still needed as used whenever the [multimatch](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#multiMatch) action is in place. Therefore, it does need to hold the computed values into a structure. It may not be necessary to save those values as operators could be applied as values got calculated, but that is discussing that we can save for later.

As the second part of your patch, you have proposed to replace **std::list** by **std::vector**, that is a fair modification. If I am not mistaken - both are _O(1)_ to insert an element by the end; **std::list** is also _O(1)_ to insert in the beginning. At arbitrary positions **std::list** is _O(1)_ in contrast to **std::vector** that is _O(n)_. As we don't insert elements at random positions, the benefit to use **std::vector** is regarding memory usage as the **std::list** is a double linked list; therefore, uses more resources. Although it can make a significant difference for big lists, we are talking about records of less than ten objects ```x = 
 (10*sizeof(*void))``` [educated guess -- don't know for sure]. Yet, it is a benefit.

Since the objective is to save ```x``` why not save twice (```2*x```) by replacing **std::list** with **std::array**?  --  if I am not mistaken **std::array** does not allocate dynamic memory at all. If the problem in using **std::array** is the fact that it is not _dynamic_, **std::doque** may suit the bill, as it does not allocate memory, yet, it is _dynamic_ and uses fewer resources than **std::list** or **std::vector**.

I am going to close this pull request, but you are more than welcomed to discuss the questions that I have raised. Looking forward to receiving anther pull request using **std::array**. Thanks @WGH-.

WGH-:
The problem with `std::list` is not just memory overhead. Its storage not contiguous, and it causes extra memory allocations and has poor locality. In fact, for small list sizes, for many workloads it may exhibit poorer performance than `std::vector`, even though complexity-wise it's better. The problem is O-notation considers only asymptotic complexity, and ignores constant complexity overhead and microarchitectural effects (like memory cache).

As of `std::array`, how can we use `std::array`, if the number of transformations is not limited?

zimmerle:
If pre-compute the transformation during the rules load time is not an option, **std::deque** may be a better indication. What do you think?

WGH-:
`std::array` size is a compile-time parameter. You can't allocate an array of arbitrary size on the stack with it.

zimmerle:
> `std::array` size is a compile-time parameter. You can't allocate an array of arbitrary size on the stack with it.

as-is no, need coding. What about the std::deque?

WGH-:
`std::deque` is slightly slower here as well (126 RPS).

Issue #2336, opened 2020-06-11
Can not get data in group when using regular expression.


anhkhoa14592:
**Describe the bug**
I tried to extract the value from PHPSESSID with regular pattern (This pattern from Web Application Defender's Cookbook: Battling Hackers and Protecting Users):
`(?i:(j?sessionid|(php)?sessid|(asp|jserv|jw)?session[-_]?(id)?|cf(id|token)|sid)=([^\s]+)\;\s?)`

But I can not get the value from group 6 (TX:6). I tried in others Text Editor and everything is fine but I don't know these pattern does not work. Maybe I miss somethings?

**Logs and dumps**

**SecRule**

    SecRule RESPONSE_HEADERS:/Set-Cookie2?/ "(?i:(j?sessionid|(php)?sessid|(asp|jserv|jw)?session[-_]?(id)?|cf(id|token)|sid)\=([^\s]+)\;\s?)" "chain,phase:3,id:'981062',t:none,pass,log,capture,setsid:%{tx.6},setvar:session.sessionid=%{tx.6},setvar:session.valid=1,msg:'%{session.sessionid}, tx.0:%{tx.0},tx.6:%{tx.6}'"
    SecRule REMOTE_ADDR "^(\d{1,3}\.\d{1,3}\.\d{1,3}\.)"  "chain,capture,setvar:session.ip_block=%{tx.1}"
    SecRule REQUEST_HEADERS:User-Agent ".*" "t:none,t:sha1,t:hexEncode,setvar:session.ua=%{matched_var}"

**Output of:**
**Response**
HTTP/1.1 200
Server: nginx/1.18.0
Date: Thu, 11 Jun 2020 11:43:06 GMT
Content-Type: text/html; charset=UTF-8
Connection: keep-alive
Cache-Control: no-store, no-cache, must-revalidate
Pragma: no-cache
X-Powered-By: PHP/7.4.6
Set-Cookie: PHPSESSID=ea101040fa9365d3ad6e921d9e1e04da; path=/
Expires: Thu, 19 Nov 1981 08:52:00 GMT

**AuditLog**
[...]

PR #2012, opened 2019-01-30
WIP: Add RE2 regexp engine support

WGH-:
See #1996.

Every commit has detailed description.

I've implemented the fallback approach so far: if RE2 support is enabled, every regular expression is compiled with RE2, and if that fails, regular expression is compiled with libpcre instead. Please see #1996 for more detailed discussion why this approach is necessary.

There's also lots of refactoring of regexp related code.

Open questions:
- [ ] Perhaps, since RE2 support is considered experimental, configure script should not even attempt to build it unless explicit `--with-re` is supplied?
- [ ] There's no easy way to know which regular expression uses which engine. It might feel like  kind of implicit magic, which isn't really good.
- [x] There's no error handling when regular expression is invalid. This isn't a regression introduced by this PR, but now might be the time to do something about it.
- [ ] PR needs code review, obviously.
- [ ] Changelog entry
- [ ] Documentation
- [ ] Maybe something else I missed.

PR #2393, opened 2020-09-05
Fix Regex::searchAll behaviour wrt empty capturing groups

WGH-:
Previously, searchAll would stop search when it encountered an empty
matching group in any position. This means that, for example,
regular expression "(a)(b?)(c)" would match string "ac", but the
resulting group list would be ["ac", "a"].

After this change, the resulting list for the aforementioned regular
expression becomes ["ac", "a", "", "c"] like it should've been.

Additionally, this also changes behaviour for multiple matches. For
example, when "aaa00bbb" is matched by "[a-z]*", previously only "aaa"
would be returned. Now the matching list is ["aaa", "", "", "bbb", ""].

The old behaviour was confusing and almost certainly a bug. The new
behaviour is the same as in Python's re.findall.

For reference, though, Go does it somewhat differently: empty matches
at the end of non-empty matches are ignored, so in Go above example is
["aaa", "", "bbb"] instead.

This is the root cause of issue #2336 which has been already fixed by
replacing searchAll call there with a new function.

Cherry-picked from #2012

### DIFF
--- a/src/utils/regex.cc
+++ b/src/utils/regex.cc
@@ -84,11 +84,12 @@ std::list<SMatch> Regex::searchAll(const std::string& s) const {
             std::string match = std::string(tmpString, start, len);
             offset = start + len;
             retList.push_front(SMatch(match, start));
+        }
 
-            if (len == 0) {
-                rc = 0;
-                break;
-            }
+        offset = ovector[1]; // end
+        if (offset == ovector[0]) { // start == end (size == 0)
+            // skip zero-length match (otherwise, the loop won't terminate)
+            offset++;
         }
     } while (rc > 0);
 


zimmerle:
@martinhsv  you may want to have a look at this one as well.

martinhsv:
You are quite right that the underlying problem with searchAll, that was one of the initial triggers for creating the new function now used by the rx operator, still exists in searchAll for that other functionality (other than the rx operator) that still use it.

However, I believe there is a correctness issue with the particular solution in this pull request.

Consider a scenario where the regex is "^(|abc|def)" and the input is "abc".

In this case there should be two successful full matches.  The first match should be an empty string.  The second match should be "abc".

With the code in this pull request, the first match of an empty string will occur successfully, but then the code will advance the offset, and on the second iteration of the loop, "abc" can no longer be produced as a match.

(Note:  There is a new searchGlobal function is another recent pull request (to support a new rxGlobal operator).  I had a notion to implement that and then, after some stabilization time, look to migrate the remaining users of searchAll to the new function. 
 You may wish to wait for that process to play out.)

WGH-:
Interesting, I never thought about that. It's also interesting to note that Python used to have "my" behaviour (`re.findall(r"^(|abc|def)", "abc") == [""]`) until Python 3.7, where it was changed to allow non-empty matches right after empty match (`re.findall(r"^(|abc|def)", "abc") == ["", "abc"]`). https://github.com/python/cpython/commit/70d56fb52582d9d3f7c00860d6e90570c6259371

zimmerle:
Hi @WGH- I am closing this and #2394 in favor of #2396. Please let me know if I am missing something.

PR #2394, opened 2020-09-05
Add Regex::searchAllMatches method, remove searchAll

WGH-:
`searchAll` interface was pretty strange, as it returned a reversed flattened list of all groups of all matches. For example, given the regexp `"([a-z])([0-9])"`, and the string `"a1b2"`, it would return the list `["2", "b", "b2", "1", "a", "a1"]`.

`searchAllMatches` introduces better interface. In the aforementioned scenario it would return `[["a1", "a", "1"], ["b2", "b", "2"]]`. Also, unlike `searchAll`, it doesn't make string copies itself, and returns just offsets, like `searchOneMatch` does.

### DIFF
--- a/src/modsecurity.cc
+++ b/src/modsecurity.cc
@@ -229,9 +229,9 @@ int ModSecurity::processContentOffset(const char *content, size_t len,
     const unsigned char *buf;
     size_t jsonSize;
 
-    std::list<Utils::SMatch> vars = variables.searchAll(matchString);
-    std::list<Utils::SMatch> ops = operators.searchAll(matchString);
-    std::list<Utils::SMatch> trans = transformations.searchAll(matchString);
+    auto vars = variables.searchAllMatches(matchString);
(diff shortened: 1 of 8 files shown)


zimmerle:
Thanks @WGH-. @martinhsv : ping. That is somewhat in the same piece of code that you have worked with.

martinhsv:
I haven't reviewed this in detail, but it looks like it at least has the same correctness problem that I outlined in https://github.com/SpiderLabs/ModSecurity/pull/2393 (that it always immediately advances the offset after finding an empty string as a full match).

Note my other remarks in that pull request as well.  To avoid too much implementation of overlapping functionality, it might be simplest to allow the rxGlobal + searchGlobal to proceed (assuming it passes muster), and then somewhat later look to migrate remaining users of searchAll to use searchGlobal.

zimmerle:
Hi @WGH- I am closing this and #2393 in favor of #2396. Please let me know if I am missing something.

PR #2433, opened 2020-10-24
Update AUTHORS

harsha3002:


zimmerle:
Hi @harsha3002,

Thank you for your input. Can you describe a briefly why such modification is important to you?

Issue #2423, opened 2020-10-16
Incorrect log message from modsec v3 for CRS rule 920450


michaelgranzow-avi:
**Describe the bug**

In chained rules, if actions 'msg' or 'logdata' contain variables like %{MATCHED_VAR}, and are used in a rule that isn't the last in the chain, the wrong MATCHED_VAR can end up in the logs.

**Logs and dumps**

The following regression test reproduces the bug:

```.json
[
  {
    "enabled":1,
    "version_min":300000,
    "title":"Test match variable",
    "expected":{
        "http_code": 437,
        "error_log": "msg \"Illegal header \\[restricted\\]\""
    },
    "client":{
      "ip":"200.249.12.31",
      "port":123
    },
    "request":{
      "headers":{
        "Host":"localhost",
        "Restricted":"attack",
        "Other":"Value"
      },
      "uri":"/",
      "method":"GET"
    },
    "server":{
      "ip":"200.249.12.31",
      "port":80
    },
    "rules":[
      "SecRuleEngine On",
      "SecRule REQUEST_HEADERS_NAMES \"^.*$\" \"phase:2,setvar:'tx.header_name_%{TX.0}=/%{TX.0}/',deny,t:lowercase,capture,id:500065,msg:'Illegal header [%{MATCHED_VAR}]',logdata:'Restricted header detected: %{MATCHED_VAR}',chain\"",
      "SecRule TX:/^header_name_/ \"@within /name1/restricted/name3/\" \"status:437\""
    ]
  }
]
```
Full logs from regression test:

```.txt
Debug log:
[1602861515] [] [4] Initializing transaction
[1602861515] [] [4] Transaction context created.
[1602861515] [] [4] Starting phase CONNECTION. (SecRules 0)
[1602861515] [] [9] This phase consists of 0 rule(s).
[1602861515] [] [4] Starting phase URI. (SecRules 0 + 1/2)
[1602861515] [/] [4] Starting phase REQUEST_HEADERS.  (SecRules 1)
[1602861515] [/] [9] This phase consists of 0 rule(s).
[1602861515] [/] [9] Appending request body: 0 bytes. Limit set to: 0.000000
[1602861515] [/] [4] Starting phase REQUEST_BODY. (SecRules 2)
[1602861515] [/] [9] This phase consists of 1 rule(s).
[1602861515] [/] [4] (Rule: 500065) Executing operator "Rx" with param "^.*$" against REQUEST_HEADERS_NAMES.
[...]

PR #2435, opened 2020-10-28
Issue-2423: Meta-actions like 'msg' should be applied at end of chain

michaelgranzow-avi:
Proposed fix for issue #2423

### DIFF
--- a/src/rule_with_actions.cc
+++ b/src/rule_with_actions.cc
@@ -215,17 +215,6 @@ void RuleWithActions::executeActionsIndependentOfChainedRuleResult(Transaction *
         }
     }
 
-    if (m_severity) {
-        m_severity->evaluate(this, trans, ruleMessage);
-    }
-
-    if (m_logData) {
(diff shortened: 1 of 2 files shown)


zimmerle:
Thanks @michaelgranzow-avi. Made few changes in your patch -

Added more tests cases -
https://github.com/SpiderLabs/ModSecurity/blob/1b7aa42c77f3edbf87911161bbe84324e2609519/test/test-cases/regression/issue-2423-msg-in-chain.json#L1-L30


Added the test cases to the **Makefile.am** to allow the execution during the _make check_-
https://github.com/SpiderLabs/ModSecurity/blob/1b7aa42c77f3edbf87911161bbe84324e2609519/Makefile.am#L160-L162


Added an entry on the CHANGELOGs with your name on it -
https://github.com/SpiderLabs/ModSecurity/blob/1b7aa42c77f3edbf87911161bbe84324e2609519/CHANGES#L4-L5

PR #2436, opened 2020-10-28
Issue-2423: Meta-actions like 'msg' should be applied at end of chain

michaelgranzow-avi:
Proposed fix for issue #2423. Slightly modified change compared to current v3/master. The 'severity' action doesn't have an 'evaluate' method in this branch, so that part is not needed.


zimmerle:
Did something different here. I have **rebased** the tree in a fashion that considers your contribution since the day 1 for v3.1. Not easy to do that, but that keeps the git-tree clean. We can keep the history of the changes...

CHANGELOGs -
https://github.com/SpiderLabs/ModSecurity/blob/abf59f4b84992195e9ba58c378d8ab4441f73b6b/CHANGES#L4-L35


Thanks @michaelgranzow-avi

PR #2440, opened 2020-11-02
Fix: Only delete Multipart tmp files after rules have run

### LINKED ISSUE #2427: Temporary file is deleted before @inspectFile is triggered if SecUploadKeepFiles set to Off and SecTmpSaveUploadedFiles set to On
**Bug description**

If: 
- SecUploadKeepFiles is disabled
- SecTmpSaveUploadedFiles is enabled
- a multipart file is uploaded,

Then the file is deleted immediately and no rules could observe this file - this behaviour looks like a bug for me (because of the 2nd item in additional context and the fact that ModSecurity v2.9 behaved as expected in such situation).

**Log, configs, rules**

ModSecurity configuration file:
```
SecRequestBodyAccess On
SecTmpSaveUploadedFiles On
SecResponseBodyLimitAction ProcessPartial
SecTmpDir /var/<dir>/tmp_modsec
SecUploadDir /var/<dir>/tmp_modsec
```

Related debug log:
```
[1602836810] [/accept.php] [9] Multipart: Content-Disposition filename: eicar.com.
[1602836810] [/accept.php] [4] Multipart: Created temporary file 1 (mode o0): /var/<dir>/tmp_modsec/20201016-082650-1602836810-file-xrFoHi
[1602836810] [/accept.php] [9] Multipart: Added file part to the list: name "file" file name "eicar.com" (offset 105, length 68)
[1602836810] [/accept.php] [4] Multipart: Cleanup started (keep files set to Not set)
[1602836810] [/accept.php] [4] Multipart: file deleted successfully (part) "/var/<dir>/tmp_modsec/20201016-082650-1602836810-file-xrFoHi"
....
[1602836810] [/accept.php] [4] (Rule: 33331) Executing operator "InspectFile" with param "<python script>" against FILES_TMPNAMES.
[1602836810] [/accept.php] [9] Target value: "/var/<dir>/tmp_modsec/20201016-082650-1602836810-file-xrFoHi" (Variable: FILES_TMPNAMES:/var/<dir>/tmp_modsec/20201016-082650-1602836810-file-xrFoHi)
[1602836810] [/accept.php] [4] Rule returned 0.
[1602836810] [/accept.php] [9] Matched vars cleaned.
```

Rule:
```
SecRule FILES_TMPNAMES "@inspectFile <python script>" \
    "id:33331,\
    log,\
    auditlog,\
    deny,\
    severity:2,\
    phase:2,\
    t:none,\
    msg:'...',\
    ctl:RuleEngine=On"
```

Note 1: `<python script>` is some python script with a shebang line in it.
Note 2: `<dir>` is a valid (rwx), existing directory in /var.

**To Reproduce**
Prerequisites:
- SecRule FILES_TMPNAMES "@inspectFile <whatever script>",
- SecTmpSaveUploadedFiles set to On,
- SecUploadKeepFiles set to Off.

Test:
- upload a multipart file,
- <whatever script> should receive its path, but the file on this path is deleted - and this is a bug, I believe.

**Expected behavior**

The uploaded file should be deleted _after_ all rules have checked it, _not before_.

**Server:**
 - ModSecurity v3.0.4 (master, effectively) with nginx-connector v1.0.1 (master, effectively),
 - nginx-1.18.0,
 - Linux, CentOS 7.

**Additional context**

1. If we add `SecUploadKeepFiles On` - everything works as expected (but the file is kept in SecTmpDir after processing, which we want to avoid).
2. This behavior seems to be triggered by a combination of:
- [branch with multipart](https://github.com/SpiderLabs/ModSecurity/blob/8da787a3902245d4d5b0c3a2732c5bb3cae6df4e/src/transaction.cc#L854)
- [actual rules call](https://github.com/SpiderLabs/ModSecurity/blob/8da787a3902245d4d5b0c3a2732c5bb3cae6df4e/src/transaction.cc#L938)
- [multipart destructor](https://github.com/SpiderLabs/ModSecurity/blob/b66224853b4e9d30e0a44d16b29d5ed3842a6b11/src/request_body_processor/multipart.cc#L80)

so the destructor is called when the branch finishes (which performs a cleanup), but the place where rules are actually evaluated is outside of this branch, so the file gets deleted.

martinhsv:
Closes #2427 

A new class encapsulates the fd and filename of a multipart-part file on disk, along with the core open/close/delete functionality.

A reference to this object is held by the transaction, so that the delete of the file on disk (which is contained in the new class's destructor) will not occur until the transaction destructor runs.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Fix: Only delete Multipart tmp files after rules have run
+    [Issue #2427 - @martinhsv]
   - Fixed MatchedVar on chained rules
     [Issue #2423, #2435, #2436 - @michaelgranzow-avi]
   - Add support for new operator rxGlobal
--- a/Makefile.am
+++ b/Makefile.am
@@ -159,6 +159,7 @@ TESTS+=test/test-cases/regression/issue-2000.json
 TESTS+=test/test-cases/regression/issue-2111.json
 TESTS+=test/test-cases/regression/issue-2196.json
 TESTS+=test/test-cases/regression/issue-2423-msg-in-chain.json
+TESTS+=test/test-cases/regression/issue-2427.json
 TESTS+=test/test-cases/regression/issue-394.json
 TESTS+=test/test-cases/regression/issue-849.json
 TESTS+=test/test-cases/regression/issue-960.json
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -111,6 +111,7 @@ enum AllowType : int;
 namespace RequestBodyProcessor {
 class XML;
 class JSON;
+class MultipartPartTmpFile;
 }
 namespace operators {
 class Operator;
@@ -611,6 +612,8 @@ class Transaction : public TransactionAnchoredVariables, public TransactionSecMa
     std::string m_variableTimeSec;
     std::string m_variableTimeWDay;
     std::string m_variableTimeYear;
+
+    std::vector<std::shared_ptr<RequestBodyProcessor::MultipartPartTmpFile>> m_multipartPartTmpFiles;
 
  private:
     /**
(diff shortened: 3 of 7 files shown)


zimmerle:
Missing a explicit constructor. Set and sent to v3/dev/pull_2440

zimmerle:
merged! thanks

PR #2443, opened 2020-11-04
Fix: crash caused by use-after-free

martinhsv:
This fixes the use-after-free issue identified by @WGH- , where two different unique_ptr objects were the owner of the same resource-managed object.


zimmerle:
Incorporated at "**actions: Removes Rule parameter from runtime execute**" to avoid fix-of-a-fix-of-a-fix commit.

PR #2446, opened 2020-11-08
Fix memory leak of RuleMessages objects

martinhsv:
This is a fix for the first (of three) memory leaks mentioned in https://github.com/SpiderLabs/ModSecurity/issues/2376 on Nov. 6.


src/transaction.cc
```diff
@@ -85,13 +85,13 @@ void TransactionRuleMessageManagement::logMatchLastRuleOnTheChain(const RuleWith
 }
 
 void TransactionRuleMessageManagement::messageNew() {
-    m_rulesMessages.push_back(new RuleMessage(m_transaction));
+    m_rulesMessages.push_back(std::shared_ptr<RuleMessage>( new RuleMessage(m_transaction)));
```
WGH-: Maybe we should eventually use `std::make_shared<T>(...)` everywhere? It uses a single allocation for both reference-counting control block and the data, whereas `std::shared_ptr<T>(new T(...))` uses a separate allocation for the control block.

WGH-:
Besides, in this particular code no new `shared_ptr` references seem to be ever created, so I think it would make sense to either make functions return `shared_ptr<T>` instead of `T*`, or just use `unique_ptr`.

martinhsv:
Regarding your last point, ordinarily I would have used unique_ptr for the container objects.  However, in this case with RuleMessage an incomplete type in transaction.h, that would create a compilation issue.

That could be resolved -- most obviously by including the full definition of RuleMessage in transaction.h.  That seemed undesirable, however, given the structural separation of API items within /headers/modsecurity and those with /src.

I did also take into account that efficiency issues decline significantly with code like this (less frequently run -- only for rule messages, as opposed to every rule evaluate for example).

martinhsv:
BTW, I agree with you about preferring make_shared.  While I wasn't so concerned about performance in this particular instance, make_shared enhances readability.  (Moreover code does tend to get cut-and-pasted, so having more 'better' examples may help in future cases where performance is more of a concern.)

Thanks for the observation; I'll actually update this pull request to that effect.

zimmerle:
I was thinking to have the connector doing a cleanup of this resource after the use. But I guess the shared_pointer seems to be a clever way to deal with it. Under the QA. Thanks ;)

PR #2449, opened 2020-11-11
Add configure option to parameterize Modsecurity/libcurl timeout for …

uhliarik:
…retrieving rules.

- There are scenarios where the internet connection is not available for the httpd,
for example during the test of Disaster Recovery or other failures. In these cases,
it would be great to have a possibility to start the httpd in the same (more or less)
timing of when there is an internet connection. Moreover, being the currently timeout
a default, it is not possible to control this value.

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -2362,7 +2362,7 @@ static const char *cmd_remote_timeout(cmd_parms *cmd, void *_dcfg, const char *p
     if (dcfg == NULL) return NULL;
 
     timeout = strtol(p1, NULL, 10);
-    if ((timeout == LONG_MAX)||(timeout == LONG_MIN)||(timeout <= 0)) {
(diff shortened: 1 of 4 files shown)


martinhsv:
I'm somewhat skeptical of the usefulness of this feature.

While contributions/suggestions are always welcome (and encouraged), not all such contributions will ultimately be deemed to be of overall benefit to the project.

FIrst of all, it seems a rather esoteric use case is being targetted. And then, even in that use case, the absence of this feature results in, at most, a pretty marginal negative impact.

In general, I'm wary of adding additional, low-value configuration items for niche scenarios. (I'm also guided by feedback from the community -- and this PR has not seen other individuals echoing support for it.)

Counterarguments are welcome. But I would tentatively plan to put this in the 'intend to close without merging' category.

martinhsv:
Closing without merge for reasons stated

Issue #2459, opened 2020-11-23
v3.1-experimental branch -- audit log messages missing when DetectionOnly


martinhsv:
Within current v3/dev/3.1-experimental ,

Audit log messages are produced as expected when 'SecRuleEngine On' but not when DetectionOnly

martinhsv:
There is an intersecting issue where the 'nolog' action is not being correctly processed.

Because there isn't a separate, user-visible use case in the code as it stands today, I'll address this additional issue in the same pull request.

PR #2461, opened 2020-11-24
Fix: audit log output when DetectionOnly; nolog action flag

martinhsv:
This pull request:
- corrects the use case where messages were not written to the audit log when DetectionOnly (the main issue identified in #2459 )
- corrects the RuleWithActionsProperties copy constructor handling of the nolog flag


zimmerle:
merged! thanks!

PR #2467, opened 2020-12-04
Adding DragonFlyBSD support.

devnexen:


zimmerle:
Thank you for the patch.

Reference for the reader
   * DragonFly BSD home - https://www.dragonflybsd.org/.
   * Further info on DragonFly - https://en.wikipedia.org/wiki/DragonFly_BSD

@devnexen out of curiosity, is that any specific reason to add DragonFly on **_whoAmI_**? 

https://github.com/SpiderLabs/ModSecurity/blob/afefda53c69bb17e2ec16d24dd6fbc2c8fd7d063/src/modsecurity.cc#L112-L152

:dragon_face:

devnexen:
Well DragonFlyBSD stands on its own thus deserves to be distinct from FreeBSD.

zimmerle:
merged! Thank you @devnexen ;)

PR #2471, opened 2020-12-07
Fix memory leak in ValidateDTD m_dtd

martinhsv:
This pull request addresses the memory leak listed as 3) in https://github.com/SpiderLabs/ModSecurity/issues/2469 .

The function call in this line of code allocates heap memory.  The second and subsequent calls to this function resulted in the the m_dtd pointer simply being assigned a new value, which meant that the heap memory previously pointed-to was leaked on a per-transaction basis.
```
m_dtd = xmlParseDTD(NULL, (const xmlChar *)m_resource.c_str());
```

With this new code, before the new parse-and-assign, we'll check if m_dtd is not null, and if so do the proper cleanup.

(I have addressed the leak in this pull request, but a larger issue we may want to consider for later implementation:  Do we really want to be calling xmlParseDTD for every transaction?  Granted, if we only want to do this once rather than per-transaction, we would need to consider the use case where the resource content changes over time -- particularly if it is an internet resource.  We could decide we don't care, or perhaps adopt a cacheing-and-periodic-update strategy)

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
  
 
+  - Fix memory leak in ValidateDTD's m_dtd
+    [#2469 - @martinhsv]
   - Using a custom VariableMatch* implementation
     [#2428 - @zimmerle]
   - Avoids to cleanup GeoIp on ModSecurity destructor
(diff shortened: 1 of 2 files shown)


zimmerle:
As mentioned on #2473 I am afraid that the leak of `const` utilization on the **libxml** may be an impediment to actually have those optimized on _run-time_ (evaluate) as oppose to _configuration time_ (init). The plan is to have evaluate as _const_ in the operator context. Lets have a chat on #2473 and we get back to this one.

martinhsv:
Closing this in favour of https://github.com/SpiderLabs/ModSecurity/pull/2478

Issue #2469, opened 2020-12-04
v3.1-experimental branch -- multiple memory leaks


martinhsv:
Within current v3/dev/3.1-experimental, I believe I have identified some additional memory leaks that have not yet been addressed:

1) set_env.cc:41
2) modsec_build.cc:30
3) validate_dtd.cc:52 (this one may be a false positive -- needs more investigation)

zimmerle:
## set_env 

I believe you are referring to:

https://github.com/SpiderLabs/ModSecurity/blob/afefda53c69bb17e2ec16d24dd6fbc2c8fd7d063/src/actions/set_env.cc#L40

Is it a consequence of **strdup**? that was used to set the variable. It is gone when the process exit.



## modsec_bulid 

Duplicate as the memory leak reported on #2376. Fixed.

Here the variable is generated:
https://github.com/SpiderLabs/ModSecurity/blob/afefda53c69bb17e2ec16d24dd6fbc2c8fd7d063/src/variables/modsec_build.cc#L31

Here is the constructor for VariableValue:
https://github.com/SpiderLabs/ModSecurity/blob/5f38e37f73f151db49e3f34981adc96d03633889/headers/modsecurity/variable_value.h#L98-L108




## validate_dtd 

https://github.com/SpiderLabs/ModSecurity/blob/afefda53c69bb17e2ec16d24dd6fbc2c8fd7d063/src/operators/validate_dtd.cc#L51-L53

Is it the right place?

martinhsv:
set_env.cc:  Perhaps I'm misunderstanding you but waiting for the process to terminate doesn't help here.  Because the memory allocation is occurring in the execute function, this is a per-transaction leak.

martinhsv:
modsec_build.cc:  Yes, the heap memory allocation is happening in that particular VariableValue constructor.  And it is due to the line with that push_back in modsec_build.cc.

But, like the previous case, because this is occurring within the evaluate function, this is a per-process leak.

martinhsv:
validate_dtd:

No, it's at validate_dtd.cc:52 (line number from 3.1-experimental):
```
m_dtd = xmlParseDTD(NULL, (const xmlChar *)m_resource.c_str());
```

I haven't fully confirmed this one, but I believe this is the same sort of issue; this line of code is in the evaluate function which means it runs every transaction and overwrites the the previous value of the pointer m_dtd. Which means that the memory that had been pointed to is leaked on a per-transaction basis.

[...]

PR #2473, opened 2020-12-08
Fix memory leak in ValidateSchema

martinhsv:
This pull request addresses the memory leak listed as 3) in #2469 .

This is analogous to https://github.com/SpiderLabs/ModSecurity/pull/2471 .

Prior to this change, second and subsequent calls to this object's evaluate function would cause a memory leak, when m_validCtx was assigned a new pointer value before the previously-pointed-to memory had been free'd.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,15 @@
+name: Manual Inputs
+on: 
+  workflow_dispatch:
+    inputs:
+      version:
+        description: Bump Version
+        default: v1.0.0
+        required: true
+jobs:
+  bump-version:
+    name: Bump Package Version
+    runs-on: ubuntu-latest
+    steps:
+      - name: Example variable usage
+        run: git tag ${{ github.event.inputs.version }}
--- a/CHANGES
+++ b/CHANGES
@@ -2,6 +2,8 @@ v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
  
 
+  - Fix memory leak in ValidateSchema
+    [#2469 - @martinhsv]
   - Using a custom VariableMatch* implementation
     [#2428 - @zimmerle]
   - Avoids to cleanup GeoIp on ModSecurity destructor
--- a/src/operators/validate_schema.cc
+++ b/src/operators/validate_schema.cc
@@ -82,6 +82,10 @@ bool ValidateSchema::evaluate(Transaction *transaction,
         return true;
     }
 
+    if (m_validCtx != NULL) {
+        xmlSchemaFreeValidCtxt(m_validCtx);
+        m_validCtx = NULL;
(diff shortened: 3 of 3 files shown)


zimmerle:
This issue is an exciting one. Let me highlight some pieces of code -

### Looking at the code

The variable instantiation happens here -
https://github.com/SpiderLabs/ModSecurity/blob/6464b9a1c570161cae7cffc93807c2a340967a2b/src/operators/validate_schema.cc#L85-L93

In the same function, we have some cleanups -
https://github.com/SpiderLabs/ModSecurity/blob/6464b9a1c570161cae7cffc93807c2a340967a2b/src/operators/validate_schema.cc#L131-L132

The variable **m_validCtx** is not being held on the cleanups. That is odd; I guess it was forgotten, and it is working like that for a while.

###  My 2 cents on the fix

I am afraid I would prefer to have a different approach on the fix -
  
1. Having ```xmlSchemaFreeValidCtxt(m_validCtx); ``` altogether with the rest of the cleanups. About line 131-132.
2. Changing the scope of the variables listed in (A) to be contiguous in the function. There is no need for those variables to be class members.

As suggested in the patch, the cleanup happens on the subsequent request, not on the current one. I am not saying that the library is thread-safe at this level, but I think it hurts the best practices having this as is. The problem preceding @martinhsv' patch: the scope is too broad.

I understand that who was implementing this operator a while ago - myself? :facepalm: - was wondering about having those initializations optimized on the operator initialization. I guess, because **m_validCtx** is not const in _xmlSchemaValidateDoc_ [[1]](http://www.xmlsoft.org/html/libxml-xmlschemas.html#xmlSchemaValidateDoc) it is impractical to have it in long term. Eventually, ```evaluate(...)``` will become _const_ in the operator context.

I understand the overhead of setting up on the **evaluation** as oppose to instantiate on the initialization. Nevertheless, not having the parameter as _const_ on the library will demand a **cast** to 'constify' it. Consequently, not gentle and could lead to a race condition very annoying for debugging :disappointed_relieved: in a not so popular functionality.

Further study can be done to identify whatever those are treated a _const_ or not by libxml. Moreover, the rationale is valid for all parameters.


#### (A) Variables that I think we should reduce the scope
https://github.com/SpiderLabs/ModSecurity/blob/6464b9a1c570161cae7cffc93807c2a340967a2b/src/operators/validate_schema.h#L136-L138

Well notice @martinhsv!! Thanks!

martinhsv:
I agree that this pull request is problematic.  Also agree that, if we still want to call xmlSchemaNewValidCtxt() on each call of evaluate(), that it makes sense to get rid of m_validCtx as a member variable and use a stack variable within the evaluate() function instead.

martinhsv:
Closing in favour or #2479

PR #2478, opened 2020-12-15
Fix memory leak of ValidateDTD's dtd object

martinhsv:
This pull request addresses the memory leak listed as 3) in #2469 .

With this pull request, xmlParseDTD() continues to be called for each call of evaluate().

However, the dtd pointer is now no longer a member variable of ValidateDTD.  It is held indirectly by a resource-managing object that is itself a stack variable.  The resource-managing object then ensures that xmlFreeDtd() when the stack variable goes out of scope and is destructed.

### DIFF
--- a/src/operators/validate_dtd.h
+++ b/src/operators/validate_dtd.h
@@ -33,18 +33,30 @@
 namespace modsecurity {
 namespace operators {
 
+class XmlDtdPtrManager {
+ public:
+    explicit XmlDtdPtrManager(xmlDtdPtr dtd)
+        : m_dtd(dtd) { }
+    ~XmlDtdPtrManager() {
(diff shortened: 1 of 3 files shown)


src/operators/validate_dtd.h
```diff
@@ -33,18 +33,30 @@
 namespace modsecurity {
 namespace operators {
 
-class ValidateDTD : public Operator {
+class XmlDtdPtrManager {
  public:
-    /** @ingroup ModSecurity_Operator */
```
zimmerle, line 38: This marking is further used by doxygen, in order to place this class under the operators sub class.

zimmerle (commented):
All perfect! Thank you! Merged!

PR #2479, opened 2020-12-16
Fix memory leaks in ValidateSchema

martinhsv:
This pull request addresses the 4th one raised in #2469 .

The main change here is to move the pointers to the three xml api objects to the evaluate function as function-scope stack variables.  At each stage in the function where a return may occur, all so-far-relevant xmlSchemaFree functions are called to clean up allocated memory.

Some additional memory leak scenarios have been addressed. For examples (transaction->m_xml->m_data.doc == NULL) and (transaction->m_xml->m_data.well_formed != 1).  These two in particular have been dealt with by performing the checks at the beginning of the function.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
  
+  - Fix memory leaks in ValidateSchema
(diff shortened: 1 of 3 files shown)


zimmerle:
Merged! thanks!

PR #2485, opened 2020-12-24
Make the `configure` step more reliable

dtoubelis:
It appears that in cross-compile environments the location of the
"current" directory cannot be assumed. This fix makes it explicit.


zimmerle:
merged! thanks!

PR #2499, opened 2021-01-09
Security.md

ghost:


zimmerle:
Hi @krebsyba, thank you.

We have opted to make a **security section** as part of our Readme file, as you can check here -

https://github.com/SpiderLabs/ModSecurity#security-issue

We think that it is important to give such a section great visibility, therefore we made it as part of the README.

PR #2505, opened 2021-01-22
Fix memory leak in rx operator when pattern includes macro

martinhsv:
This corrects an issue only very recently introduced in v3/master (just a few days ago).


zimmerle:
Hi @martinhsv,

This was fixed here - 9b40a045bba26808d444e7c6ade5f6b2a7f4af15

Issue #2514, opened 2021-02-16
bad request leads to two responses


pgajdos:
**Bug Description**

After specific wrong request I get second response from Apache httpd server:

```
$ printf "GET / HTTP/1.1\r\nHost: localhost\r\nTransfer-Encoding: chunked\r\n\r\n\r\n" | nc localhost 60080
HTTP/1.1 400 Bad Request
Date: Thu, 11 Feb 2021 12:48:06 GMT
Server: Apache/2.4.46 (Linux/SUSE)
Connection: close
Content-Type: text/html; charset=iso-8859-1

<html>400</html>
<html>It works!</html>
$
```
Where
```
<html>It works!</html>
```
is content of index.html in docroot.

**Logs and dumps**

Not necessary.

**To Reproduce**

Minimal httpd config:

```
ServerName test
User user
Group users
Listen 60080
PidFile /tmp/apache-rex/poc/pid
ErrorLog /tmp/apache-rex/poc/error_log
LoadModule auth_basic_module /usr/lib64/apache2-prefork/mod_auth_basic.so
LoadModule dir_module /usr/lib64/apache2-prefork/mod_dir.so
LoadModule authz_host_module /usr/lib64/apache2-prefork/mod_authz_host.so
LoadModule authz_core_module /usr/lib64/apache2-prefork/mod_authz_core.so
LoadModule unique_id_module /usr/lib64/apache2-prefork/mod_unique_id.so
LoadModule security2_module /usr/lib64/apache2/mod_security2.so
LogLevel trace8
DocumentRoot /srv/www/htdocs
DirectoryIndex index.html

ErrorDocument 400 "<html>400</html>\n"

SecRuleEngine DetectionOnly
SecRequestBodyAccess on
SecAuditLogType Serial
SecAuditLog /tmp/apache-rex/poc/modsec_audit.log
SecTmpDir /tmp/
SecDataDir /tmp/

### example configuration

<Directory "/tmp/apache-rex/poc/htdocs">
  Require local
</Directory>
```

**Expected behavior**

```
$ printf "GET / HTTP/1.1\r\nHost: localhost\r\nTransfer-Encoding: chunked\r\n\r\n\r\n" | nc localhost 60080
HTTP/1.1 400 Bad Request
Date: Thu, 11 Feb 2021 12:48:06 GMT
Server: Apache/2.4.46 (Linux/SUSE)
Connection: close
Content-Type: text/html; charset=iso-8859-1

<html>400</html>
[...]

PR #2518, opened 2021-02-22
fix for #2514, thanks to @ylavic

pgajdos:
See discussion in the issue.


marcstern:
For what it worth, we're running this patch from begin March in several (QA) WAF without any problem

marcstern:
Running in 60+ prod servers since June 2021

martinhsv:
I think this PR would benefit from a description about the current code path, why it's a problem, and how this PR fixes it.

I'm not doubting that those involved have come to correct conclusions, but more information would help me to assess the PR for merging. (Not to mention that it might assist other members of the community following this issue both now and later.)

PR #2535, opened 2021-03-15
Test for proving Loading rule is not thread safe

iosetek:
When multiple threads are trying to load rule at the same
time even when they refer to different rule set a string
representing rule is malformed. It's believed that
this is caused by seclang-parser which contains plenty
of global variables.

To run the test simply enter 'test' directory and run
"./unit_tests"

### DIFF
--- a/test/unit/unit.cc
+++ b/test/unit/unit.cc
@@ -13,12 +13,14 @@
  *
  */
 
-#include <string.h>
-#include <cstring>
 
-#include <iostream>
+#include <chrono>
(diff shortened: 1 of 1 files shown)


zimmerle:
I think this pull request was built to illustrate #2536. Please follow the discussion on #2536.

JohannesAarsnes (commented):
Need help to catch the IP of user who hax files.. also how to change adminrights(script?) to prevent users to hijack account?

PR #1213, opened 2016-09-20
Optional JIT support in PCRE

defanator:
May be useful for everyone who's attempting to build libmodsecurity on systems with pcre < 8.20 (such as RHEL/CentOS/OEL 6.x).

PR #2562, opened 2021-04-28
Fix pcre's JIT support detection was not working

nullpo-head:
Pcre's JIT support detection introduced by #1213 was not working due to a bug caused by `AC_TRY_COMPILE`.
This PR fixes it and enables to opt out JIT support correctly.

`AC_TRY_COMPILE` macro detects whether a given C program compiles, and the original code used it to detect JIT support by `pcre_config(PCRE_CONFIG_JIT,`. However, `pcre_config(PCRE_CONFIG_JIT,` is a runtime feature to detect JIT support, and it compiles anyway, regardless of whether JIT support exists or not. A correct fix is to check whether it can be linked with some jit-specific API such as `pcre_jit_exec`.

### DIFF
--- a/build/pcre.m4
+++ b/build/pcre.m4
@@ -69,14 +69,12 @@ if test -n "${PCRE_VERSION}"; then
     AC_MSG_CHECKING(for PCRE JIT)
     save_CFLAGS=$CFLAGS
     save_LDFLAGS=$LDFLAGS
+    save_LIBS=$LIBS
     CFLAGS="${PCRE_CFLAGS} ${CFLAGS}"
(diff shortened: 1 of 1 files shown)


zimmerle:
Kudos @nullpo-head 👍 👍 . Thank you for reporting. I am going to analyze it a little bit further tomorrow.

zimmerle:
Having it on QA here - https://github.com/SpiderLabs/ModSecurity/pull/2562/commits/e23aa2ff67174953027bb1d784629fd5f08f4930

nullpo-head:
Thanks so much!

zimmerle:
Merged! on v3/master.

PR #2574, opened 2021-05-31
New recommended rule to enable JSON parser for subtypes with +json su…

### LINKED ISSUE #2564: Recommended rules: Expand 200001 / JSON body processor rule
OWASP ModSec CRS project here. We have a pull request to override the very basic Content-Type definition in the
ModSec recommended rule 200001 (-> `REQUEST_HEADERS:Content-Type "application/json"`) with a much more comprehensive list of JSON CTs: [CRS PR 2049](https://github.com/coreruleset/coreruleset/pull/2049).

This is what IANA lists for JSON:

```
curl -s 'https://www.iana.org/assignments/media-types/application.csv' | grep json | awk 'BEGIN{FS=","}{print $2}'
```

I would rather keep things separate, since I think this belongs into ModSec recommended rules.

But that depends on your interest to update 200001. We're happy to provide a PR, but please give us a signal first.

martinhsv:
This pull request includes one way to address the underlying ask in #2564 .

Without assuming that this change (or anything similar) will necessarily proceed within the ModSecurity project and its modsecurity.conf-recommended file, this pull request is made available at least for assessment and comment.

Rather than adjust the existing rule 200001, I have opted to create a new rule 200006 to cover the additional use cases (the type is 'application' with the subtype not 'json', but has a +json subtype).  This separation of rules might be advantageous if we are at all worried that some ModSecurity users might experience undesired effects from this change.  Having a separate rule for the additional functionality makes it simpler to disable or comment-out the new rule and restore former behaviour.

A second nuance/detail of note may be that I chose to specify the literal '+' plus character within a character class.  This was done to avoid any inconsistencies (especially between v2.9 and v3) with backslash-escaping.

If are hesitant to take even this level of risk with rules that activate the JSON parser, yet another option could be to include the new rule in the file, but leave it commented-out.  This would not be entirely unprecedented; we already have several configuration items in the file present but commented-out.  This would at least save users interested in the functionality from having to think about and compose a new or modified regex themselves.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - New rule to enable JSON parser for subtypes with +json suffix
+    [@dune73, @martinhsv]
   - Having ARGS_NAMES, variables proxied
     [@zimmerle, @martinhsv, @KaNikita]
   - Use explicit path for cross-compile environments.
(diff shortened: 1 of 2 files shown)


martinhsv:
Closing in favour of #2586 (an equivalent rule structure but, per the last paragraph above, the rule is provided commented-out).

PR #2578, opened 2021-06-14
fix memory leak in the example

bsulmanas:
`error` string is created via `strdup` and it must be passed to `free()` to avoid a memory leak.

Not really important in this particular case, but if would be consistent with `msc_rules_cleanup(rules)` and `msc_cleanup(modsec)`

### DIFF
--- a/examples/simple_example_using_c/test.c
+++ b/examples/simple_example_using_c/test.c
@@ -70,6 +70,7 @@ int main (int argc, char **argv)
 end:
     msc_rules_cleanup(rules);
     msc_cleanup(modsec);
+    free(error);
 
     return 0;
 }


examples/simple_example_using_c/test.c
```diff
@@ -70,6 +70,7 @@ int main (int argc, char **argv)
 end:
     msc_rules_cleanup(rules);
     msc_cleanup(modsec);
+    free(error);
```
zimmerle, line 73: Hi @bsulmanas,

The variable *error* is indeed filled if there is an error. Otherwise, it may point towards NULL. Considering the former case and the suggested patch, it will lead to a double-free-ish mistake. The variable error can be sefely freed on lines 45 and 54, where its usage is known.

zimmerle (changes requested):
Hi @bsulmanas,

Please see the review inline.

examples/simple_example_using_c/test.c
masterl: Yeah, zimmerle's suggestion is a good solution.

To keep it at the 'end' label as your code is right now, you would need to wrap it inside a `if ( error != NULL )` because reaching the end label doesn't mean there was an error, and if there was no error the code would try to do a `free( NULL );`...

martinhsv: Actually, I'm fairly sure that the PR is safe as it stands. Calling free on a NULL pointer has been considered safe from very early days.

See, for example 7.20.3.2 from https://www.open-std.org/JTC1/SC22/wg14/www/docs/n1124.pdf

"If ptr is a null pointer, no action occurs"

eduar-hte:
This has been inadvertently addressed in the context of #3209, see commit dab9bb6.

airween:
Thanks @eduar-hte for pointing this out. I'm closing this issue.

PR #2580, opened 2021-06-18
fix some memory leaks for parsing & cleaning up rules

Abce:


877509395:
great for "delete transformations"

src/rule_with_actions.cc
```diff
@@ -80,6 +80,10 @@ RuleWithActions::RuleWithActions(
     m_containsStaticBlockAction(false),
     m_isChained(false) {
 
+    if (transformations != NULL) {
```
liudongmiao, line 83: why this?

liudongmiao: Just understand. It's the same like `delete actions` do.

poszu:
@zimmerle What is holding you back from merging this? I also have a problem with these memory leaks, it would be great to merge it. ;)

liudongmiao:
@poszu From my test in #2710, even patched this, there are many memory leak.
Which g++ version do you use?

poszu:
> @poszu From my test in #2710, even patched this, there are many memory leak. Which g++ version do you use?

Yes, I am aware of other leaks too. I'm using clang 10. ASAN reports plenty of leaks on the v3.0.6.

src/parser/driver.cc
```diff
@@ -129,9 +129,9 @@ int Driver::parse(const std::string &f, const std::string &ref) {
     m_lastRule = nullptr;
     loc.push_back(new yy::location());
     if (ref.empty()) {
-        loc.back()->begin.filename = loc.back()->end.filename = new std::string("<<reference missing or not informed>>");
+        loc.back()->begin.filename = loc.back()->end.filename = std::shared_ptr<const std::string>(new std::string("<<reference missing or not informed>>"));
```
poszu, line 132: Consider using `std::make_shared` to avoid using `new`
```suggestion
        loc.back()->begin.filename = loc.back()->end.filename = std::make_shared<const std::string>("<<reference missing or not informed>>");
```

poszu, line 132: Consider using `std::make_shared` to avoid using `new`
```suggestion
        loc.back()->begin.filename = loc.back()->end.filename = std::make_shared<const std::string>("<<reference missing or not informed>>");
```

martinhsv:
I don't believe the portions of this PR related to filename are correct.

The problem is that location.hh is a generated file. The manual change to that file suggested in this PR could be made and a build done. But thereafter, any time bison rebuilt the generated files, those manual changes would be lost.

I'll go ahead and commit the other two fixes via a separate PR:
- Rule class was missing a virtual destructor
- delete of transformations vector

The parser-filename issue can be addressed later via a separate issue or PR.

martinhsv:
The two issues have been merged via #2801 .

Thanks @Abce

PR #2586, opened 2021-06-30
Add commented-out sample rule to engage JSON Processor for more subtypes

### LINKED ISSUE #2564: Recommended rules: Expand 200001 / JSON body processor rule

martinhsv:
This pull request is a replacement for #2574 in addressing #2564 

The rule provided here is the same as in the former pull request, but it is provided (like some other configuration items in the file) commented-out.

This is provided as a convenience for users so that they might be able to avoid having to think about how to compose a rule themselves.  Only removing the two comment characters is necessary to activate the new rule.

### DIFF
--- a/modsecurity.conf-recommended
+++ b/modsecurity.conf-recommended
@@ -29,6 +29,13 @@ SecRule REQUEST_HEADERS:Content-Type "(?:application(?:/soap\+|/)|text/)xml" \
 SecRule REQUEST_HEADERS:Content-Type "application/json" \
      "id:'200001',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
 
+# Sample rule to enable JSON request body parser for more subtypes.
+# Uncomment or adapt this rule if you want to engage the JSON
+# Processor for "+json" subtypes
+#
+#SecRule REQUEST_HEADERS:Content-Type "^application/.+[+]json$" \
+#     "id:'200006',phase:1,t:none,t:lowercase,pass,nolog,ctl:requestBodyProcessor=JSON"
+
 # Maximum request body size we will accept for buffering. If you support
 # file uploads then the value given on the first line has to be as large
 # as the largest file you are willing to accept. The second value refers


martinhsv:
Unless there are objections, I'll plan to merge this within around a day.

dune73:
I overlooked #2574 and only returned to this topic today.

I second your comments about the `+` symbol. That makes a lot of sense.

But what is your reasoning for commenting out this update? Or in other words, where is the sense in accepting JSON payloads without handling them with the right body processor. That way they will just be blobs with a lot of curly brackets in REQUEST_BODY - or completely garbled variables with funny names.

Adding it as a complementary JSON rule forces you to add a new rule id, you picked 200006. I do not readily see a way around this, but with 200000 and 200001 and 200006 being very different in nature from 200002, 3, 4 and 5, the set of recommended rules becomes somewhat un-orderly. This is probably not very obvious, but in my classes, people struggle to grasp the function of the various recommended rules.

zimmerle:
As the name suggests [modsecurity.conf-recommended](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/modsecurity.conf-recommended) is a recommendation to the user;  The recommendations oftentimes can be used as a guide for users to customize ModSecurity to fit their end-application. I understand this pull request's value as it educates the user of the possibilities. Therefore, I think it is safe to proceed with the merge.

I understand there may be a second discussion about whether it will be interested in supporting that as _default_ in our recommendation file (uncommented); keep this pull request open for further discussion.

@martinhsv are you interested in doing this change for v2/master as well? If so, please link this pull request.

dune73:
I see. 

I agree having the commented out new rule is better than not having it. But I would argue that the new rule should replace the existing 200001.

martinhsv:
Hi @dune73 ,

Regarding your observation about the rule numbering:  I agree that it is a downside that the sample, commented-out rule is 006 and hence discontinuous with the 'normal' JSON-Processor-enabling rule.  Renumbering rules 002-005, however, seemed like a very bad idea.  I did also locate the new item within the file just after 200001 in an attempt to minimize potential confusion.

To your most recent comment, and to build slightly on what @zimmerle noted ...

The modsecurity.conf-recommended rules are not intended to be comprehensive.  Indeed the comments above rule 200001 already put the onus on the user to modify as required:
```
# Initiate JSON Processor in case of JSON content-type; change accordingly
# if your application does not use 'application/json'
```

Given that +json subtype suffixes are sufficiently non-rare, I was nevertheless persuaded that we should consider doing _something_ here to be of at least some minor assistance to users.

Why not simply amend 200001 to cover all cases?

One reason is that there is some non-zero risk that not all of those +json subtypes will actually be json.  As noted previously, there are over a hundred such subtypes and it is impractical to assess them all individually.  While one would think they all would be, based on the name, we already had an example from your original posting in #2564 that defied the initial expectation (application/json-seq).  So what if one or more are json-ish but not a true, valid json object?  I was wary of inadvertently sowing more confusion than we would be alleviating.

Given all of the above, at a recent internal discussion, I wound up suggesting the commented-out, sample rule in this pull requests as something of a compromise (possibly merely as an interim step towards having that functionality enabled in a non-commented-out rule).

Given that you seem to agree that the current pull request represents at least some incremental improvement, I hope you can take it in that spirit.  As this gets noticed by more people over the coming weeks and months, feedback from the community will, as always, be highly valued.

dune73:
Thank you for your extensive explanations, @martinhsv. I like your considerate approach to this problem.

So you have real evidence that somebody submits a payload marked as JSON but it's not really JSON. I see the problem, but without knowing any numbers I'd probably conclude this was block-worthy offense. But I admit that this can have tough effects in the context of the recommended-rules since you block immediately (vs. anomaly scoring).

I've never read the recommended rules as a proposal. I do indeed see them as comprehensive even when there are shortcomings (and I tell my students to remove 200004 since it causes too many FPs in my experience). This difference points to a bigger problem. I think it takes a lot of operational experience to come up with something that can be seen as more comprehensive than the recommended rules. I would definitely not expect the average WAF engineer to create something secure in this regard. So my experience is that people take the recommended rules as is (minus 200004 if they follow my tutorials).

We may have to talk this through, since we need a common ground in this regard. We've had discussions within CRS about replacing the recommended rules with our own set of base configuration rules. I always thought that was an unnecessary complication of the setup, so I would very much prefer to stick to your rules as long as they are comprehensive _enough_.

So yes, I can live with a commented out rule and it's better than the status quo. But it also tilts the balance somewhat in favor of replacing the recommended rules for CRS.

Asking for more people to chime in here is a good plan. Thank you.

lamjay2704 (commented):
[]()

Issue #2131, opened 2019-07-10
Linking issue on FreeBSD


granalberto:
**Describe the bug**

When compiling from source on FreeBSD 11.2-RELEASE, the linker is called with -llibmaxminddb and it makes it fail. The correct option would be -lmaxminddb.

**Logs and dumps**

/usr/bin/ld: cannot find -llibmaxminddb

zimmerle:
Hi @granalberto 

Thanks for the report. I will have a look. Just installed a FreeBSD machine here to run a couple of tests.

zimmerle:
Build bots are running on v3/dev/issue2131. A small change but may broken compilation for others, so let's see what the buildbots are going to tell us.

airween:
I can't reproduce this issue.

```
./configure --without-lmdb --enable-parser-generation --enable-mutex-on-pm --prefix=/usr
...
ModSecurity - v3.0.3-95-gd5b93c10 for FreeBSD
 
 Mandatory dependencies
   + libInjection                                  ....v3.0.3-95-gd5b93c10
   + SecLang tests                                 ....d5b93c10
 
 Optional dependencies
   + GeoIP/MaxMind                                 ....found 
      * (MaxMind) v1.3.2
         -lmaxminddb , -DWITH_MAXMIND -I/usr/local/include 
   + LibCURL                                       ....found v7.65.1 
      -L/usr/local/lib -lcurl, -I/usr/local/include -DWITH_CURL_SSLVERSION_TLSv1_2 -DWITH_CURL
   + YAJL                                          ....found v2.1.0
      -lyajl , -DWITH_YAJL -I/usr/local/include 
   + LMDB                                          ....disabled
   + LibXML2                                       ....found v2.9.9
      -L/usr/local/lib -lxml2 -lz -L/usr/lib -llzma -L/usr/lib -lm, -I/usr/local/include/libxml2 -I/usr/include -DWITH_LIBXML2
   + SSDEEP                                        ....not found
   + LUA                                           ....found 
      -llua-5.3 -lm  -L/usr/local/lib , -DWITH_LUA -I/usr/local/include/lua53 
 
 Other Options
   + Test Utilities                                ....enabled
   + SecDebugLog                                   ....enabled
   + afl fuzzer                                    ....disabled
   + library examples                              ....enabled
   + Building parser                               ....enabled
   + Treating pm operations as critical section    ....enabled
 ```
additional info's:
```
## --------- ##
## Platform. ##
[...]

PR #2596, opened 2021-07-25
build/libmaxmind.m4: fix build with host-pkgconf

ffontaine:
Build with maxminddb is broken since version 3.0.5 and https://github.com/SpiderLabs/ModSecurity/commit/785958f9b5089b918c7d054cbcc2fe4a3c7b3788 because `libmaxminddb` has been removed from `MAXMIND_POSSIBLE_LIB_NAMES`

So, as suggested by Arnout in #2131, don't use `MAXMIND_POSSIBLE_LIB_NAMES` for pkg-config, because it was never called
anything other than `libmaxminddb`

Signed-off-by: Fabrice Fontaine <fontaine.fabrice@gmail.com>

### DIFF
--- a/build/libmaxmind.m4
+++ b/build/libmaxmind.m4
@@ -60,12 +60,10 @@ else
             # Nothing about MaxMind was informed, using the pkg-config to figure things out.
             if test -n "${PKG_CONFIG}"; then
                 MAXMIND_PKG_NAME=""
-                for x in ${MAXMIND_POSSIBLE_LIB_NAMES}; do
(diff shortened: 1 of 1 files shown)


ffontaine:
Do you have some comments on this PR?

ffontaine:
No feedback?

martinhsv:
I suppose this looks reasonable.

In a worst case we can revisit (or revert) if anyone reports it resulting in problems.

Thanks @ffontaine.  And also @arnout for making the original suggestion here: https://github.com/SpiderLabs/ModSecurity/issues/2131#issuecomment-886224005

Issue #2006, opened 2019-01-20
Extra whitespace in some configuration directives causing error


porjo:
**Describe the bug**

Putting more than one space or tab between *some* configuration directives and their value causes an error.

For example:
```
SecAuditLog /usr/local/apache24/logs/modsec_audit.log
```
If I put 2 or more spaces between `SecAuditLog` and `/usr/local...` (or any number of tabs) I get this error:

```
nginx: [emerg] "modsecurity_rules_file" directive Rules error. File: /usr/local/apache24/conf/modsecurity_rules.conf. Line: 34. Column: 55. Invalid input:  SecAuditLog  /usr/local/apache24/logs/modsec_audit.log in /usr/local/nginx-1.10.1/conf/nginx.conf:48
```

Other configuration directives don't have this problem. For example I can put many spaces between `SecAuditLogFormat` and `JSON`

**Expected behavior**

Extra whitespace (within reason) should be permitted between directive and value as this can assist to make the configuration file easier to read. Also, behaviour should be consistent for all directives.

**Server (please complete the following information):**
 - ModSecurity version (and connector): ModSecurity v3.0.3 with nginx-connector v1.0.0
 - WebServer: nginx-1.10.1
 - OS (and distro): Linux, centos7

**Rule Set (please complete the following information):**
 - Running any public or commercial rule set? SpiderLabs commercial rules
 - What is the version number? 2018-08-11

**Additional context**

I've also tested with Apache connector (cloned from tip 2019-01-17)

zimmerle:
Hi @porjo,

Thanks for the report. Fixed at: dc78c0e180d09aa1b2f0ffad1a8d6967e69f984f

porjo:
Thanks for the quick fix! :smile:

PR #2600, opened 2021-08-08
fix include handle

877509395:
"const char *file = strchr(yytext, ' ') + 1"  supposes only one space after Include, but the grammar "[ \t]+" mean multiple space and TAB including one space;

### DIFF
--- a/src/parser/seclang-scanner.ll
+++ b/src/parser/seclang-scanner.ll
@@ -3,7 +3,6 @@
 #include <climits>
 #include <cstdlib>
 #include <string>
-#include <regex.h>
 
 #include "src/parser/driver.h"
 #include "src/parser/seclang-parser.hh"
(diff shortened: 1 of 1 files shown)


martinhsv:
Just to flesh out a little background (and correct me if you think I'm mistaken) ...

In v3.0.3, several configuration directives allowed for exactly one space character after the directive name.

v3.0.4 included changes (https://github.com/SpiderLabs/ModSecurity/issues/2006 ) to broaden the whitespace that is allowed for many of these directives, including the two referenced in this pull request.

The last two changes in that commit ( https://github.com/SpiderLabs/ModSecurity/commit/dc78c0e180d09aa1b2f0ffad1a8d6967e69f984f ) changed the whitespace definition from '[ ]' to '[ \t]+' ...

... and you believe that fix was incomplete.

---------------------------------------------------

That background stated, can you describe what use case fails to work as expected?

I just tried changing my main.conf file to include additional spaces between 'Include' and the file path and it worked fine.  (Perhaps I've misunderstood something.)

martinhsv:
Just to flesh out a little background (and correct me if you think I'm mistaken) ...

In v3.0.3, several configuration directives allowed for exactly one space character after the directive name.

v3.0.4 included changes ( #2006  ) to broaden the whitespace that is allowed for many of these directives, including the two referenced in this pull request.

The last two changes in that commit ( https://github.com/SpiderLabs/ModSecurity/commit/dc78c0e180d09aa1b2f0ffad1a8d6967e69f984f ) changed the whitespace definition from '[ ]' to '[ \t]+' ...

... but that that portion of change appears to be incomplete.

That seems to me to be a correct conclusion.  For one thing, using tabs instead of spaces after the 'Include' keyword is a problem use case.

Turning to this proposed fix:

One thing I'm a little wary of, is including regex.h here.  We haven't previously used that within the parser subcomponent, and, given that Flex has its own regular expression engine, it might be confusing in future if we start mixing that in.

877509395:
I will  remove regex.h;

877509395:
Yes, You are right
If no #2006 included, use one space is correct; 
but after #2006 included, [ \t]+ should be supported correctly for  {CONFIG_INCLUDE}[ \t]+["]{CONFIG_VALUE_PATH}[" and {CONFIG_INCLUDE}[ \t]+{CONFIG_VALUE_PATH}.
So after v3.0.4, it should be fixed, using tabs/spaces after the 'Include' keyword should not be a problem of use case, the code and the lex do not match.

877509395:
ignore this PR.

PR #2602, opened 2021-08-09
Fix #2601 misuses of LMDB API

### LINKED ISSUE #2601: LMDB support is broken
**lmdb.cc misuses the LMDB API in multiple ways**

The code calls mdb_dbi_open/mdb_dbi_close in multiple locations without any concurrency protection.
The API doc clearly states that mdb_dbi_open must not be called from multiple transactions at once.
The doc also clearly states that mdb_dbi_close is not protected at all and must never be called from
more than one thread at once. The doc also states that programs *SHOULD* only call mdb_dbi_open
once, when the environment is created, and should only call mdb_dbi_close when the env is being
destroyed.

http://www.lmdb.tech/doc/group__mdb.html#ga52dd98d0c542378370cd6b712ff961b5
http://www.lmdb.tech/doc/group__mdb.html#ga52dd98d0c542378370cd6b712ff961b5

The code calls mdb_txn_abort on invalid transactions after mdb_txn|_commit fails. The doc clearly
states that after mdb_txn_commt has been called, the txn handle is freed. As such, passing the handle
the txn_abort references freed memory and triggers undefined behavior.

http://www.lmdb.tech/doc/group__mdb.html#ga846fbd6f46105617ac9f4d76476f6597

**Logs and dumps**

A crash caused by these bugs has been reported here https://bugs.openldap.org/show_bug.cgi?id=9626

hyc:
Only open DBI once, doesn't need closing.
Never reuse a txn handle after commit.
Use MDB_RDONLY for txns that aren't doing any writes

### DIFF
--- a/src/collection/backend/lmdb.cc
+++ b/src/collection/backend/lmdb.cc
@@ -37,9 +37,13 @@ namespace backend {
 
 LMDB::LMDB(std::string name) :
     Collection(name), m_env(NULL) {
+    MDB_txn *txn;
     mdb_env_create(&m_env);
     mdb_env_open(m_env, "./modsec-shared-collections",
         MDB_WRITEMAP | MDB_NOSUBDIR, 0664);
+    mdb_txn_begin(m_env, NULL, 0, &txn);
+    mdb_dbi_open(txn, NULL, MDB_CREATE | MDB_DUPSORT, &m_dbi);
+    mdb_txn_commit(txn);
 }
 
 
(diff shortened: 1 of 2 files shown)


martinhsv:
Hello @hyc ,

I finally had a chance to look at this in detail and one thing I wondered about:

There is a new member variable 'MDB_dbi m_dbi;' in the LMDB class -- which makes sense.

MDB_dbi is typedef'd to 'unsigned int', which means m_dbi is undefined initially.

The LMDB constructor will initialize it with a successful call to mdb_dbi_open.

But what if it fails? The return code from mdb_dbi_open isn't checked and we proceed as if it had been successful.

I had a look at both the LMDB documentation and (a brief) skim of the code and it doesn't look like a failure condition within mdb_dbi_open will result in the MDB_dbi out-parameter being updated to a special value or anything (that might have been checked for in subsequent calls to other functions).

Thoughts?

hyc:
The default dbi, which is what your code uses, always exists and is always valid, so the mdb_dbi_open call will never fail.

PR #2604, opened 2021-08-14
fix include

877509395:
Fix lex "include [ \t]filename" handle. To support more spaces/tabs after include key word. Before this change, code supports only one space after include. but lex support more.

### DIFF
--- a/src/parser/seclang-scanner.ll
+++ b/src/parser/seclang-scanner.ll
@@ -3,6 +3,7 @@
 #include <climits>
 #include <cstdlib>
 #include <string>
+#include <string.h>
 
 #include "src/parser/driver.h"
 #include "src/parser/seclang-parser.hh"
@@ -1237,7 +1238,12 @@ EQUALS_MINUS                            (?i:=\-)
 
 {CONFIG_INCLUDE}[ \t]+{CONFIG_VALUE_PATH} {
     std::string err;
-    const char *file = strchr(yytext, ' ') + 1;
+
+    char *tmpStr = strdup ( yytext+strlen("include")); 
+    char *fileNameStart  = strtok ( tmpStr, " \t");
+    const char *file = yytext +strlen("include") + (fileNameStart-tmpStr);
+    free(tmpStr);
+
     std::string fi = modsecurity::utils::find_resource(file, *driver.loc.back()->end.filename, &err);
     if (fi.empty() == true) {
         BEGIN(INITIAL);
@@ -1264,9 +1270,14 @@ EQUALS_MINUS                            (?i:=\-)
 
 {CONFIG_INCLUDE}[ \t]+["]{CONFIG_VALUE_PATH}["] {
     std::string err;
-    const char *file = strchr(yytext, ' ') + 1;
-    char *f = strdup(file + 1);
-    f[strlen(f)-1] = '\0';
+
+    char *tmpStr = strdup ( yytext+strlen("include")); 
+    char *fileNameStart  = strtok ( tmpStr, " \t");
+    const char *file = yytext +strlen("include") + (fileNameStart-tmpStr);
+    free(tmpStr);
+
+    char *f = strdup(file);
+
     std::string fi = modsecurity::utils::find_resource(f, *driver.loc.back()->end.filename, &err);
     if (fi.empty() == true) {
         BEGIN(INITIAL);


martinhsv:
I like the simplification of leaving out regex.h.

But I don't think you need to make a copy.  For example, for the first of the two sections, wouldn't this suffice?:
```
char *tmpStr = yytext + strlen("include");
const char *file = strtok ( tmpStr, " \t");
```

877509395:
strtok(str) will set the next byte of last byte of token to '\0' if token found, so it may change the str. If str contains multiple [ \t], and we need enumerate  all possible tokens(call strtok multiple times), multiple bytes would be set to 0. below is a example.

</br>#include <string.h>
int main () {
   &nbsp;&nbsp;char str[80] = "This is - www.runoob.com - website";
   &nbsp;&nbsp;const char s[2] = "-";
   &nbsp;&nbsp;int i=0;
   &nbsp;&nbsp;char *token;
   &nbsp;&nbsp;int length;
   &nbsp;&nbsp;length = strlen(str);
   &nbsp;&nbsp;token = strtok(str, s); 
   &nbsp;&nbsp;while( token != NULL ) { 
   &nbsp;&nbsp;&nbsp;&nbsp;i++;
   &nbsp;&nbsp;&nbsp;&nbsp;printf( "token%d        =%s\n", i,token );
   &nbsp;&nbsp;&nbsp;&nbsp;printf( "token%d address=%p\n", i, token );  
   &nbsp;&nbsp;&nbsp;&nbsp;token = strtok(NULL, s); 
   &nbsp;&nbsp;&nbsp;&nbsp;printf("\n");
   &nbsp;&nbsp;}   
   &nbsp;&nbsp;printf( "str          =%s\n",str);
   &nbsp;&nbsp;printf( "str          =%p\n", str  );  
   &nbsp;&nbsp;return(0);
}

the output is:
token1        =This is 
token1 address=0x7ffe0e856810

token2        = www.runoob.com 
token2 address=0x7ffe0e856819

token3        = website
token3 address=0x7ffe0e85682a

str          =This is 
str          =0x7ffe0e856810
~                
From the output, str is changed to "This  is", the next byte of "This is"  changed to 0. every token address is between str and str+length;  

For our case, strtok only is called only once, and based on the below grammar, it is NOT possible there is a [ \t]+ in "CONFIG_VALUE_PATH", so strtok will not change CONFIG_VALUE_PATH LUCKILY;
"CONFIG_VALUE_PATH                       [0-9A-Za-z_\/\.\-\*\:]+"
{CONFIG_INCLUDE}[ \t]+{CONFIG_VALUE_PATH}
{CONFIG_INCLUDE}[ \t]+["]{CONFIG_VALUE_PATH}["]

so below change can suffice.
  char *tmpStr = yytext + strlen("include");
  const char *file = strtok ( tmpStr, " \t");

But I think it is a good practice to first strdup and then free for strtok： 
1. strtok may do harm to  input.
2. In case for some reason in the future, CONFIG_VALUE_PATH can contain space/tab. 
3. hidden logic(can not contain space/tab in CONFIG_VALUE_PATH) of the code need to dig out

martinhsv:
Ok, good point.  I agree that it's unwise to use strtok here without a copy.

Is it worth considering using strspn instead?  It's a little more narrowly focussed on what is needed here.

```
char *tmpStr = yytext + strlen("include");
const char file = tmpStr + strspn ( tmpStr, " \t");
```

877509395:
Perfect, strspn exactly matches this context. I think I should go through libc.pdf to know every function.
</br>
remove tmpStr, merge two lines into  one line:

const char *file = yytext + strlen("include")+ strspn ( yytext + strlen("include"), " \t");

Is it OK to remove tmpStr?

877509395:
#include <stdio.h></br>
#include <string.h>
int main(int argc, char**argv)
{
&nbsp;&nbsp;const char *yytext = "include           lj lj ljsdfj";
&nbsp;&nbsp;const char *file = yytext + strlen("include")+ strspn ( yytext + strlen("include"), " \t");
&nbsp;&nbsp;printf("%s", file);
}
the output is:
[root@localhost strcspn]# ./a.out 
lj lj ljsdfj

877509395:
BTW,
Same issue also seems exists in many other line. for example,
{CONFIG_DIR_AUDIT_LOG2}[ \t]+{CONFIG_VALUE_PATH}                        { return p::make_CONFIG_DIR_AUDIT_LOG2(strchr(yytext, ' ') + 1, *driver.loc.back());
I have not check it carefully now.

877509395:
misoperation close this PR and delete lexhandle

martinhsv:
Regarding collapsing the code change to a single line, I might lean towards leaving it as two lines, because:
- allocating an extra pointer on the stack doesn't really create extra cost or risk and
- it may make the code a little bit easier to read and
- it avoids duplicating code `strlen("include")` ... including having a hard-coded string specified twice; it's generally a good coding practice to avoid such duplication

My preference on that isn't particularly strong, however,

877509395:
reasonable

877509395:
will submit PR tomorrow

PR #2612, opened 2021-09-01
fix [ \t]+ processing for include xxxx.conf

877509395:


martinhsv:
This was a PR replacement for #2604 (which in turn was a replacement for #2600 ).

This change is being merged via #2800 , so that the generated file is included as well.

Thanks @877509395 , for the contribution.

PR #2619, opened 2021-09-10
Get client public ip address in json audit log

rahulgargmit:
When Modsecurity running behind a proxy that send X-Forwarded-For header(for example load balancer), current remoteip module is broken. it's not able to get client public ip.
To get client public ip in v3, i removed remoteip module and added a new variable "client_public_ip" in audit log json, which have value sent by 'X-Forwarded-For' header.

Got the idea to remove remoteip module from here
https://github.com/SpiderLabs/ModSecurity/issues/756#issuecomment-612549616

My setup is
 - Modsecurity in a docker behind load balancer
 - Removed remoteip module and it's recommended config from apache
 - JSON logs with HTTPS type
     SecAuditLogType HTTPS
     SecAuditLog "http://127.0.0.1:8042/post/waflog"
     SecAuditLogFormat JSON

### DIFF
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -430,6 +430,11 @@ class Transaction : public TransactionAnchoredVariables, public TransactionSecMa
      * Holds the HTTP version: 1.2, 2.0, 3.0 and so on....
      */
     std::string m_httpVersion;
+    
+    /**
+     * Holds the Public ip of user, as per X-Forwarded-For header
+     */
+    std::string m_remotePublicIp;
 
     /**
      * Holds the server IP Address
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -1659,7 +1659,7 @@ std::string Transaction::toJSON(int parts) {
     yajl_gen_map_open(g);
     /* Part: A (header mandatory) */
     LOGFY_ADD("client_ip", this->m_clientIpAddress->c_str());
-    LOGFY_ADD("client_public_ip", m_remotePublicIp.c_str());
+    LOGFY_ADD("x_forwarded_for", m_remotePublicIp.c_str());
     LOGFY_ADD("time_stamp", ts.c_str());
     LOGFY_ADD("server_id", uniqueId.c_str());
     LOGFY_ADD_NUM("client_port", m_clientPort);
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -574,6 +574,18 @@ int Transaction::processRequestHeaders() {
         ms_dbg(4, "Rule engine disabled, returning...");
         return true;
     }
+    
+    std::vector<const VariableValue *> *l2 = \
+	new std::vector<const VariableValue *>();
+    m_variableRequestHeaders.resolve("X-Forwarded-For", l2);
+
+    if (l2->size() >= 1) {
+        m_remotePublicIp = std::string(l2->at(0)->getValue());
+        ms_dbg(4, "PUBLIC_IP header found " + m_remotePublicIp);
+    }
+    else {
+        ms_dbg(4, "PUBLIC_IP header not found");
+    }
 
     this->m_rules->evaluate(modsecurity::RequestHeadersPhase, this);
 
@@ -1467,6 +1479,7 @@ std::string Transaction::toOldAuditLogFormatIndex(const std::string &filename,
        m_variableRequestHeaders.resolveFirst("Host").get())
         << " ";
     ss << utils::string::dash_if_empty(this->m_clientIpAddress->c_str()) << " ";
+    ss << utils::string::dash_if_empty(m_remotePublicIp.c_str()) << " ";
     /** TODO: Check variable */
     variables::RemoteUser *r = new variables::RemoteUser("REMOTE_USER");
     std::vector<const VariableValue *> l;
@@ -1526,6 +1539,7 @@ std::string Transaction::toOldAuditLogFormat(int parts,
     audit_log << tstr;
     audit_log << " " << m_id->c_str();
     audit_log << " " << this->m_clientIpAddress->c_str();
+    audit_log << " " << m_remotePublicIp.c_str();
     audit_log << " " << this->m_clientPort;
     audit_log << " " << m_serverIpAddress->c_str();
     audit_log << " " << this->m_serverPort;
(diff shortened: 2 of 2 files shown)


877509395:
if modsecurity running behind a proxy, REMOTE_ADDR is invalid, So all remote_addr related rule is invalid

877509395:
you help me a lot.

martinhsv:
The motivation here is laudable, but I would lean away from adding this sort of logic to the ModSecurity engine.  I'll list some reasons:

1) 'X-Forwarded-For' is only one of the header names that may be used for this type of functionality.  While 'X-Forwarded-For' is the most commonly used in my experience, for a general-purpose tool like ModSecurity to only focus on one of them in populating something called CLIENT_IP has the potential to create confusion.

2) 'X-Forwarded-For' can be spoofed.  Take an ordinary deployment where there is no load balancer or anything else between the untrusted internet and ModSecurity.  A malicious user can simply input anything desired into the 'X-Forwarded-For' header and send the request.  In this case, if ModSecurity reports that maliciously-composed content as something called 'CLIENT_IP', I would regard that as quite misleading.  This would be easy for the ModSecurity user to misunderstand or misinterpret ... and this would be a particularly unfortunate misunderstanding in security software.

3) 'X-Forwarded-For' can have a set of IP addresses.  For example, this is a properly formatted header:
```
X-Forwarded-For: 192.168.66.67, 192.168.66.68, 192.168.66.69
```
The pull request here does not attempt to parse these and will simply record the entire header value as 'CLIENT_IP'.  Parsing could be added to the code, of course, but it increases the complications.  See also the next point.

4) The last of the IP addresses in the string should ordinarily be the original IP address.  However because of the spoofing problem (point 2 above), the load balancer (or whatever is at the edge of trusted network) would need to be configured to first strip off any 'X-Forwarded-For' content that it received **before** populating the header that it passes on.

All told, I think the ModSecurity engine is not really the right place to add this functionality.

On the other hand, populating a transaction variable based on content extracted from headers can already be done using rules -- and this gives the ModSecurity admin to tailor the functionality to the environment being protected.

I'll leave this open for now to give others opportunity to comment.

rahulgargmit:
@martinhsv Purpose is not to overwrite current CLIENT_IP or REMOTE_IP, it's just to add one more parameter in json, which can provide value of X-Forwarded-For header.
We can change parameter name to "x-forwarded-for", instead of "client_public_ip".
There is another parameter in json "cleint_ip", that'll remain same.
It's up to users then, if they wish to use value provided in "x-forwarded-for" or keep using "client_ip".

Its needed because currently there is no way to get client's public ip when modsecurity behind proxy. Please let me know if there is one.i already tried apache remoteip module.

877509395:
Just wonder if client public IP is really needed behind load balance, How should I get it ?




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "SpiderLabs/ModSecurity"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2021年9月15日(星期三) 晚上8:17
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [SpiderLabs/ModSecurity] Get client public ip address in json audit log (#2619)





 
@martinhsv Purpose is not to overwrite current CLIENT_IP or REMOTE_IP, it's just to add one more parameter in json, which can provide value of X-Forwarded-For header.
 We can change parameter name to "x-forwarded-for", instead of "client_public_ip".
 There is another parameter in json "cleint_ip", that'll remain same.
 
—
You are receiving this because you commented.
Reply to this email directly, view it on GitHub, or unsubscribe.
Triage notifications on the go with GitHub Mobile for iOS or Android.

rahulgargmit:
> Just wonder if client public IP is really needed behind load balance, How should I get it ?
> […](#)
> ------------------&nbsp;原始邮件&nbsp;------------------ 发件人: "SpiderLabs/ModSecurity" ***@***.***&gt;; 发送时间:&nbsp;2021年9月15日(星期三) 晚上8:17 ***@***.***&gt;; ***@***.******@***.***&gt;; 主题:&nbsp;Re: [SpiderLabs/ModSecurity] Get client public ip address in json audit log (#2619) @martinhsv Purpose is not to overwrite current CLIENT_IP or REMOTE_IP, it's just to add one more parameter in json, which can provide value of X-Forwarded-For header. We can change parameter name to "x-forwarded-for", instead of "client_public_ip". There is another parameter in json "cleint_ip", that'll remain same. — You are receiving this because you commented. Reply to this email directly, view it on GitHub, or unsubscribe. Triage notifications on the go with GitHub Mobile for iOS or Android.

There can be different reasons why they need clientip, in our scenario, we are sending audit log to another application, which use clientip to get geolocaiton.
I added some code in this PR about how we can get clientip, please see code changes in commit.

macrokernel:
> @martinhsv Purpose is not to overwrite current CLIENT_IP or REMOTE_IP, it's just to add one more parameter in json, which can provide value of X-Forwarded-For header.
> We can change parameter name to "x-forwarded-for", instead of "client_public_ip".

Indeed, it looks like a good idea to rename the parameter to "x_forwarded_for" to make it more self-descriptive.

rahulgargmit:
@macrokernel 
Renamed parameter from "client_public_ip" to "x_forwarded_for"
We can leave parsing of multiple values, it's user responsibility, we are just adding X-Forwarded-For header value to json parameter.

martinhsv:
Perhaps I'm misunderstanding something about the perceived need.  But if all you want is to see the content of the X-Forwarded-For header in the JSON-formatted audit log, it's already present where request headers are shown:

{"transaction":{...,"request":{"method":"GET","http_version":1.1, ... ,"headers":{"Host":"localhost",**"X-Forwarded-For":"192.168.66.66"**,"Content-Length":"0"} ...

rahulgargmit:
@martinhsv 
Yes, you are correct
we never got this header in json because of mod remoteip used
"RemoteIPHeader X-Forwarded-For"

as given here
https://github.com/SpiderLabs/ModSecurity/issues/756#issuecomment-612549616

After removing mod remoteip, i can see X-Forwarded-For header in json.
Thanks

877509395:
Want to configure black/white list behind load balance based on client IP address, 
1. Add  X_FORWARDED_FOR_CLIENT_IP;
2. Add rule something like
   SecRule  X_FORWARDED_FOR_CLIENT_IP "@ipMatchFromFile blacklist.data" "chain,phase:1,deny,log,tag:blacklist,msg:blackmember,status:403,id:2700000"

How about this procedure? any suggestion ?

877509395:
hi,&nbsp;
&nbsp; &nbsp; &nbsp;I had thought that msc_add_n_request_header should be called to handle request header, but NOT. I suspect I had make mistake, but I think I should not, I double checked. I can see header info by tcpdump tools, but this function does not be called. if this function not used, which function to handle request header&nbsp; ? can someone help m.
&nbsp; &nbsp; &nbsp;


extern "C" int msc_add_n_request_header(Transaction *transaction,
&nbsp; &nbsp; const unsigned char *key,
&nbsp; &nbsp; size_t key_len, const unsigned char *value, size_t value_len);
extern "C" int msc_add_response_header(Transaction *transaction,
&nbsp; &nbsp; const unsigned char *key,
&nbsp; &nbsp; const unsigned char *value) ;



thanks in advance.
huiming

PR #2644, opened 2021-11-19
Support configurable limit on depth of JSON parsing

martinhsv:


877509395:
Why this limit? does it cause Practical problems?
memory should only be the limit, depth should not. i think

martinhsv:
Additional information will be posted later today.

877509395:
with malicious  input, the yajl parser can be very deep (big depth),  So cause DOS. right?

877509395:
Where can i found POC for for this Dos?

PR #2649, opened 2021-12-02
Create SECURITY.md

877509395:


877509395:
so DOS is possible.

martinhsv:
Hi @877509395 ,

Per my earlier comments, I do not think it is wise or appropriate to include this here.

There is an email address available for communicating possibly sensitive security concerns.  See https://github.com/SpiderLabs/ModSecurity#security-issue

PR #2661, opened 2021-12-23
Multipart names may include single quote if double-quote enclosed

martinhsv:
This pull request reduces the set of use cases that result in 'Invalid Quoting' being flagged during multipart parsing of the Content-Disposition's 'name' or 'filename'.

The following will no longer result in 'Invalid Quoting':
```
name="ab'cd"
```

The following will continue to result in 'Invalid Quoting':
```
name='abcd'
name=ab'cd
```

This change matches recent updates done in v2, see that pull request for more detail:
https://github.com/SpiderLabs/ModSecurity/pull/2660

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Multipart names/filenames may include single quote if double-quote enclosed
+    [Issue #2352 @martinhsv]
   - Add SecRequestBodyJsonDepthLimit to modsecurity.conf-recommended
     [Issue #2647 @theMiddleBlue, @airween, @877509395 ,@martinhsv]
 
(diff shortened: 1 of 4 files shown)


binhjy:
Hi @martinhsv , this fix will be on which release? Thanks

martinhsv:
It will be included automatically in the next official release in each of v2.9.x and v3.0.x. But it is available immediately for anyone who clones from github and builds ModSecurity themselves.

binhjy:
Hi @martinhsv , can you double check also with this rule SecRule FILES_NAMES|FILES: https://github.com/coreruleset/coreruleset/blob/v3.0.2/rules/REQUEST-920-PROTOCOL-ENFORCEMENT.conf#L100 (pls change the branch)? Will the name="ab'cd" passes this rule?

martinhsv:
Hello @binhjy ,

The coreruleset is managed by a separate group. If you have questions about the specific behaviour of individual rules, you should contact them directly.

PR #2685, opened 2022-02-25
V2/redundancy

marcstern:
This adds a little check before adding tags and logdata to the actions, to ensure this is not already present in the table.
In case of redundant (thus exactly identical) tags, it avoids the tag being logged multiple times. Same for logdata.
This redundancy is common when using macros that adds some tags/logdata at different places.

Remark 1: This avoids duplicates in the ruleset table which is allocated in the ruleset pool that stays for the whole process life. This frees thus some memory.
Therefore, I also added other actions that have no mean to be redundant and cannot be "undone" by another action. This memory optimization is not so minor as the ruleset is duplicated in each sub-location.

Remark 2: We could also free memory for all the actions that can only have one value, like severity, by applying the opposite rule: removing everything but the last action. This could also be achieved by a cleanup process, instead of "inline" like my test. But that's another story ...

### DIFF
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -1029,8 +1029,10 @@ msre_action *msre_create_action(msre_engine *engine, apr_pool_t *mp, const char 
     return action;
 }
 
-// Return 1 if "name=value" is present in table
-static int apr_table_exists_(apr_pool_t* p, const apr_table_t* vartable, const char* expected, const char* name, const char* value) {
+// Return 1 if "name=value" is present in table (for supplied action)
+static int apr_table_action_exists(apr_pool_t* p, const apr_table_t* vartable, const char* action, const char* name, const char* value) {
+    if (strcmp(name, action) != 0) return 0;
+
     const char* vars = apr_table_getm(p, vartable, name);
     if (!vars) return 0;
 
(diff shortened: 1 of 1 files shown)


martinhsv:
I'm not sure I'm convinced that this is a good idea.

The aim is basically to auto-correct an (apparent) error on the part of the rule writer.

As a general principle, I would argue that that's the sort of thing that ModSecurity should steer away from. The power of the tool should be left in the hands of the administrator(s) who are responsible for the rules. Moreover, any tag and logdata duplicate logging is something that is readily apparent from the audit log output; the admin can easily make any needed adjustments themselves when they see that. And we shouldn't assume that it's an error.

Also, I haven't tried it out, but it seems to me that this change would also reduce flexibility available to the admin. For example what if the content is not *really* duplicate because it involves a macro expansion and the expanded value changes between the first and second occurrences? Granted this is a likely a pretty marginal use case. But what this change is intended to help with is also for a pretty marginal use case.

marcstern:
The only auto-correction is to remove **redundant** entries like tags (mainly tags), nothing else. No flexibility reduction; if several different tags are added, they will all be present. Only if twice the same is registered, it's removed.
100% bullet-proof.

Marginal case: not when you use macros, it's very common.

> the admin can easily make any needed adjustments themselves
When using macros it's often impossible because you create several macros adding, each of them, a tag. When combining them, you end up with 3 times the same tag.

martinhsv:
Hi @marcstern ,

Perhaps it would be helpful if you provided a couple of sample rules that you think would be representative of common use cases?

marcstern:
Example with Define:
```
Define xss_tags "tag:security,tag:xss,tag:html"
Define sql_tags "tag:security,tag:sql"
Define dev_tags "tag:security,tag:develop"
SecRule ARGS "@pm /*" "msg:'C-like comment used also in JavaScript & SQL',${xss_tags},${sql_tags},${dev_tags}"
```
The tag 'security' will be added 3 times

martinhsv:
Hi @marcstern ,

That doesn't seem like that compelling of a case to me, since the admin could just set it up instead to only include the tags that are necessary to avoid the duplication.

Moreover, what if the admin specifically wants duplicates? Perhaps she's even using something similar to what you have, and then does post-processing of the logs to take special action if there are two or more occurrences of the same 'security' tag in the same log line?

marcstern:
> the admin could just set it up instead to only include the tags that are necessary to avoid the duplication

The admin that sets up the the Define is not always the same one as the one using it, so "no"

> what if the admin specifically wants duplicates?

You're right (although this use case sounds much weirder than mine 😉), so an option would probably be the best choice

martinhsv:
" ... admin that sets up the the Define is not always the same ... "

Maybe the two admins should talk to each other ;-)

If Admin2, who sets up the rule, cannot convince Admin1, who sets up the Defines, to provide what Admin2 needs, then Admin2 can just write his own text for the rule(s) in question.

In any case, I'm going to go ahead and close this for the reasons discussed.

PR #2714, opened 2022-04-12
Use pkg-config to find libxml2, falling back to xml2-config if necessary

hughmcmaster:
Debian is taking steps to remove `xml2-config` in favour of `pkg-config`. This means ModSecurity will build without libxml2 support by default on Debian, Ubuntu and other distributions tracking Debian packages.

This patch modifies `build/libxml.m4` to check for libxml2 via `pkg-config`, falling back to `xml2-config` if necessary.

### DIFF
--- a/build/libxml.m4
+++ b/build/libxml.m4
@@ -4,20 +4,7 @@ dnl Sets:
 dnl  LIBXML2_CFLAGS
 dnl  LIBXML2_LIBS
 
-LIBXML2_CONFIG=""
-LIBXML2_VERSION=""
-LIBXML2_CFLAGS=""
-LIBXML2_CPPFLAGS=""
-LIBXML2_LDADD=""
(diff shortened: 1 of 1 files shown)


hughmcmaster:
Hi @martinhsv, hoping you can look at this MR soon. Debian goes into freeze in January...

martinhsv:
Hello @hughmcmaster ,

I wasn't aware of Debian's plan re: xml2-config. Thanks for both raising the issue and submitting the PR.

I'll plan to merge this shortly.

In our other maintained branch (v2/master), it looks to me like the same issue applies in the comparable file (find_xml.m4). Would you be interested in submitting a pull request to update  that one as well? (No obligation, of course.)

hughmcmaster:
Thanks, @martinhsv. I'm happy to send a patch for the other branch.

PR #2718, opened 2022-04-19
Fix build errors with multiple Lua libraries

stanhu:
This commit fixes two items:

1. Previously `configure` would search for Lua in known paths, but
this could cause the Makefile to use inconsistent header and library
paths. For example, if luajit were installed in `/usr/local/include`
but lua 5.1 were installed in `/usr/lib/liblua-5.1.so`, the build
would attempt to use the luajit headers but link against the lua 5.1
library.

To fix this, we switch the order of the search:

* First attempt to find an installed LUA library with `pkg-config`. 
* If we cannot find a library that way, fall back to the known-path
scan.

This is actually what is already documented in `LUA_POSSIBLE_PATHS`.

Note that `PKG_CONFIG_PATH` can be specified in `configure` to look in the right path:

```shell
PKG_CONFIG_PATH=/usr/local/pkgconfig ./configure
```

2. Add luajit back into `LUA_POSSIBLE_LIB_NAMES`. This was added in
0ac23a47 but quietly reverted in fe98ce4c. The changes in the first
item also ensure the `CFLAGS` are set properly for luajit.

This should fix the issues raised in #1909.

### DIFF
--- a/build/lua.m4
+++ b/build/lua.m4
@@ -6,7 +6,7 @@ AC_DEFUN([CHECK_LUA],
 [dnl
 
 # Possible names for the lua library/package (pkg-config)
-LUA_POSSIBLE_LIB_NAMES="lua54 lua5.4 lua-5.4 lua53 lua5.3 lua-5.3 lua52 lua5.2 lua-5.2 lua51 lua5.1 lua-5.1 lua"
+LUA_POSSIBLE_LIB_NAMES="lua54 lua5.4 lua-5.4 lua53 lua5.3 lua-5.3 lua52 lua5.2 lua-5.2 lua51 lua5.1 lua-5.1 luajit-5.1 luajit lua"
 
 # Possible extensions for the library
 LUA_POSSIBLE_EXTENSIONS="so la sl dll dylib"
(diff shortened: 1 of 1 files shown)


stanhu:
@victorhora Could you take a look at this?

stanhu:
@martinhsv Would you be able to review this?

martinhsv:
Hello @stanhu ,

The only question that comes to mind initially is about where you have placed the two new items in LUA_POSSIBLE_LIB_NAMES.

I guess I'm wondering aloud if it might be better for the two luajit names to be at the end of the list instead. That might make the order in which things are looked for more straightforward (I.e. all regular lua options first, then luajit)

stanhu:
@martinhsv I moved them to the end. It's been a while since I looked at this, so I don't remember if there was a reason why I put them before the standard `lua` options. But the main issue as I recall was that we weren't taking advantage of `pkg-config` to find the right libraries and headers when there was ambiguity.

stanhu:
From https://github.com/SpiderLabs/ModSecurity/issues/1909#issuecomment-1101861948, I think I intended to the `luajit` modules to precede `lua` since `luajit` could install itself in `/usr/local/lib/lua`.

PR #2719, opened 2022-04-20
Support PCRE2

### LINKED ISSUE #2668: Add support for PCRE2
The legacy PCRE implementation (more recently referred to as PCRE1) is no longer being updated, with v8.45 (from June 2021) expected to be the final version.

Also, nginx has recently (nginx/1.21.5) changed to use PCRE2 by default, and this will cause problems if not using the --without-pcre2 flag when building (see https://github.com/SpiderLabs/ModSecurity-nginx/pull/260, https://github.com/SpiderLabs/ModSecurity-nginx/pull/261 ).

martinhsv:
This pull request provides initial support for PCRE2 on an experimental basis. (Closes #2668 )

To use PCRE2 with nginx and ModSecurity you need to:
- use a version of nginx that supports it (e.g. nginx v1.21.5 or 1.21.6)
- use a version of the connector (ModSecurity-nginx) at or later than https://github.com/SpiderLabs/ModSecurity-nginx/commit/1c32794046c9e0065d4ad58f949fabb538c44222
- include the ```--with-pcre2``` option when running ModSecurity's ```./configure```

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Support PCRE2
+    [Issue #2668 - @martinhsv]
   - Support SecRequestBodyNoFilesLimit
     [Issue #2670 - @airween, @martinhsv]
   - Fix misuses of LMDB API
--- a/build/pcre2.m4
+++ b/build/pcre2.m4
@@ -1,0 +1,183 @@
+dnl Check for PCRE2 Libraries
+dnl CHECK_PCRE2(ACTION-IF-FOUND [, ACTION-IF-NOT-FOUND])
+
+AC_DEFUN([PROG_PCRE2], [
+
+# Possible names for the pcre2 library/package (pkg-config)
+PCRE2_POSSIBLE_LIB_NAMES="pcre2 pcre2-8"
+
+# Possible extensions for the library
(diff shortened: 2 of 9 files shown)


istiak101:
Do I need to pass the PCRE2 path too? Explicit `--with-pcre2=yes` didn't work.

```
```

I don't know it's using build/pcre.m4 instead of build/pcre2.m4

istiak101:
So PCRE1 config binary location is /usr/bin/pcre-config and PCRE2 config binary /usr/bin/pcre2-config.
In PCRE1 the config script is detected correctly but not with PCRE2.

dvershinin:
@istiak101 @martinhsv apparently there's a bug somewhere.

`--with-pcre2` only compiles fine if both pcre1 and pcre2 devel libs installed. And the resulting binary links with both, which is incorrect, of course.

If only pcre2 devel is installed on the system, the error message  during configure is always `configure: error: pcre library is required`.

See filed new bug https://github.com/SpiderLabs/ModSecurity/issues/2750

martinhsv:
Hello @istiak101 ,

As @dvershinin has noted, the dependency on pcre1 has not been removed even if you specifically want to use pcre2. But beyond that, does it work if you just use (although =yes should work too):
```
./configure --with-pcre2
```

istiak101:
Hello @martinhsv,

Yes if install both pcre-devel and pcre2-devel it works.

PR #2721, opened 2022-04-21
Scheme

877509395:
just to show error, not PR.

### DIFF
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -522,7 +522,7 @@ int Transaction::processURI(const char *uri, const char *method,
     if (!m_uri_decoded.empty() && m_uri_decoded.at(0) != '/') {
         bool fullDomain = true;
         size_t scheme = m_uri_decoded.find(":")+1;
-        if (scheme == std::string::npos) {
+        if (scheme == std::string::npos) { //impossible case, for +1 in size_t scheme = m_uri_decoded.find(":")+1
             fullDomain = false;
         }
         // Searching with a pos of -1 is undefined we also shortcut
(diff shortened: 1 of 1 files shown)


martinhsv:
Hi @877509395 ,

Thanks for highlighting that.

The logic is indeed not correct for what was intended, and should get fixed.

Note, however, that it doesn't look like there is any circumstance with the currently-supported v3 deployment (with nginx and the ModSecurity-nginx connector), that this code path will execute.

If you think I'm mistaken and the effect is relatively trivial, feel free to post here. If you think I'm mistaken and the effect is more consequential (possible rule bypass, etc.) please outline the use case to us at the address listed at https://github.com/SpiderLabs/ModSecurity#security-issue .

(Note: For future reference, items like this are better created as 'issues' rather than pull requests.)

877509395:
OK, will open issue later. Will double check the code later.




------------------&nbsp;原始邮件&nbsp;------------------
发件人:                                                                                                                        "SpiderLabs/ModSecurity"                                                                                    ***@***.***&gt;;
发送时间:&nbsp;2022年4月22日(星期五) 晚上11:18
***@***.***&gt;;
***@***.******@***.***&gt;;
主题:&nbsp;Re: [SpiderLabs/ModSecurity] Scheme (PR #2721)





 
Hi @877509395 ,
 
Thanks for highlighting that.
 
The logic is indeed not correct for what was intended, and should get fixed.
 
Note, however, that it doesn't look like there is any circumstance with the currently-supported v3 deployment (with nginx and the ModSecurity-nginx connector), that this code path will execute.
 
If you think I'm mistaken and the effect is relatively trivial, feel free to post here. If you think I'm mistaken and the effect is more consequential (possible rule bypass, etc.) please outline the use case to us at the address listed at https://github.com/SpiderLabs/ModSecurity#security-issue .
 
(Note: For future reference, items like this are better created as 'issues' rather than pull requests.)

PR #2727, opened 2022-04-24
fix memory leak when concurrent log includes REMOTE_USER

liudongmiao:


martinhsv:
A less common configuration + use case, that even then, should only leak at most proportionally to the number of transactions generating an audit log write.

Nevertheless, it is in principle a per-transaction leak, so a good find.

Thanks for the contribution.

PR #2731, opened 2022-04-26
Tolerate other parameters after boundary in multipart C-T

martinhsv:
This pull request supports permissively handling a Content-Type header for a multipart request, where the header incorrectly has additional parameters after the boundary. A debug log message is produced for this use case, but processing can otherwise continue as if the extra parameter(s) had not been present.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Tolerate other parameters after boundary in multipart C-T
+    [Issue #1900 - @martinhsv]
   - Add DebugLog message for bad pattern in rx operator
     [Issue #2723 - @martinhsv]
   - Support PCRE2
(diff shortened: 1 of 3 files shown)


koms313:
Does this works if the charset was before the boundary, not after it?

martinhsv:
@koms313 ,

No. This PR only supports toleration of the charset= parameter if it appears after.

Because inclusion of the charset= parameter was determined to be non-compliant with the relevant RFCs, the goal was not to provide full parsing support as if it were a legitimate inclusion. The scope of the intended change was limited to tolerating a known non-malicious client -- and the only such identified cases involved the charset= parameter occurring after.

koms313:
Thanks @martinhsv. We are currently using Apache HTTP5 client which includes charset in the header without a way to remove it. This is an example of a header
`Content-Type: multipart/form-data; charset=ISO-8859-1; boundary=TkTEHffIt9-zLdw5OeAQd5yv4ytx52OH3H_`
We will look into using another library instead.

fkesheh:
Feign also adds charset before the boundary:
https://github.com/OpenFeign/feign-form/blob/98e51d187f1b7acba3b39d1131371b761d87259b/feign-form/src/main/java/feign/form/MultipartFormContentProcessor.java#L94

PR #2757, opened 2022-06-02
Update README.md with syntax changes

jeremiah:
Syntax changes that increase comprehension especially for native English speakers.


martinhsv:
I'm going to close this without merging.

While README.md could use a rewrite, and there is one true grammar error identified, the specific suggestions made here are either debatable or inadvisable (e.g. I actually think the pre-existing 'rearchitected' captures what happened better than the suggested replacement 'redesigned').

And right now, I just cannot justify any time for a back-and-forth on minor wording changes.

Issue #2755, opened 2022-06-02
v3.0.7 segfault at liblmdb.so.0.0.0


dvershinin:
It appears in v3.0.7 that some changes have been done in regards to LMDB that cause a segfault in the library. 

Environment:

* CentOS 7 or CentOS Stream 8
* nginx v1.22.0 compiled with latest release version of nginx security module and libmodsecurity v3.0.7 (compiled library with `./configure --with-lmdb --with-pcre2`)
* latest release of OWASP CRS

Result: any URLs being accessed trigger:

```
[2766481.294938] nginx[18205]: segfault at 7c ip 00007fad3894fd91 sp 00007ffcabc55b50 error 4 in liblmdb.so.0.0.0[7fad3894c000+14000]
[2766481.300369] Code: ff ff 48 8d 78 40 e8 3e ed ff ff e9 02 ff ff ff 66 0f 1f 84 00 00 00 00 00 41 57 41 56 41 55 41 54 55 53 48 89 fb 48 83 ec 18 <44> 8b 6f 7c 48 8b 6f 20 41 81 e5 00 00 02 00 4c 8b 7d 40 0f 84 86
```

martinhsv:
Hi @dvershinin ,

Are you able to provide a stack trace?

dvershinin:
@martinhsv 

```console
gdb nginx-debug /tmp/cores/core.nginx-debug.626957
GNU gdb (GDB) Red Hat Enterprise Linux 8.2-16.el8
Copyright (C) 2018 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-redhat-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from nginx-debug...Reading symbols from /usr/lib/debug/usr/sbin/nginx-debug-1.22.0-4.el8.x86_64.debug...done.
done.

warning: Can't open file (null) during file-backed mapping note processing

[...]

PR #2761, opened 2022-06-09
Prevent LMDB related segfault for #2755

dvershinin:
Make transactions no-op if the file handle is invalid (due to permissions, or other reasons like lmdb library mismatch).


martinhsv:
While it's a good idea to avoid conditions that result in a segfault, alternatives that result in failing silently are pretty disadvantageous too.

I'll have a closer look. What we might wind up doing is merging something close to what you have here, and then I can add some additional content on top of it.

martinhsv:
I'll go ahead and merge this. Some further improvements, including to make the failure non-silent, can be made later.

Thanks for the contribution @dvershinin

PR #2780, opened 2022-08-03
V2/rule remove target by tag/macro

marcstern:
Allow macro expansion in ctl:ruleRemoveTargetByTag.
This allows, for instance, to detect headers not containing a specific character, then removing these headers from some rules.


martinhsv:
I have a couple of points of hesitation about this:

This usage is outside of how the 'tag' action was originally conceived as working. I.e. as simple static text that is useful for classifying and categorization rules.

That doesn't mean we cannot consider expanding the functionality beyond its originally-intended use cases, but perhaps it would be useful to include some examples of where/how this expanded functionality might be useful?

The main downside of proceeding with this would be the added processing cost/time. Granted it's not large, but every transaction would have to pay this cost. And if only a very small number of ModSecurity deployments are interested in this functionality it should give us a bit of pause -- particularly if the underlying need in the expected use cases could be addressed some other way.

marcstern:
Example (very simplified):

# Store name of all ARGS containing %
SecRule ARGS "%" "phase:1,setvar:tx.percentArgs=%{tx.percentArgs}|%{MATCHED_VAR_NAME},..."
# Exclude all ARGS not containing % for rules tagged with "PERCENT"
SecRule ARGS @unconditionalMatch "phase:1,chain"
SecRule MATCHED_VARS_NAMES "!^(?:%{tx.percentArgs})$" "ctl:'ruleRemoveTargetByTag=PERCENT:%{MATCHED_VAR}'"

marcstern:
Obsoleted by https://github.com/SpiderLabs/ModSecurity/pull/2964

Issue #1135, opened 2016-05-03
module name not recorded in apache error.log


benkr94:
Here is a sample entry from error.log written by this module:

`[Tue May 03 13:27:18.000422 2016] [:notice] [pid 14:tid 139983030101888] ModSecurity for Apache/2.7.7 (http://www.modsecurity.org/) configured.`

I am using the default apache2 ErrorLogFormat. In lines written by other modules, the module name appears in the second set of brackets before the log level (eg, `[pagespeed:warn]` or `[core:error]`). Why does security2 appear to give an empty string here? This occurs for all error log lines for this module.

csanders-git:
Actually looking at the code i'm not quite sure as to why this is. I'm rather unsure of where Apache gleans this information from. 
The code itself is rather straight forward: https://github.com/SpiderLabs/ModSecurity/blob/808ea48263f2123ea7b8d56678b16ea42855cfb3/apache2/mod_security2.c#L743-L744

zimmerle:
The lines that you used as example, is part of the "startup banner" of ModSecurity. During the execution you still get the same behavior? Can you paste an example?

benkr94:
Yes, during execution I get the same behavior. Here's an example:

[...]

PR #2781, opened 2022-08-04
Fix logging for Apache 2.4 (again)

erkia:
There was a fix for Apache 2.4 logging in https://github.com/SpiderLabs/ModSecurity/commit/f813365f, which was partly reverted by https://github.com/SpiderLabs/ModSecurity/commit/da995bb6. 

Unfortunately, this reversion also removed the APLOG_USE_MODULE macro , which most likely wasn't causing any issues but is needed for Apache log messages to be correctly marked as belonging to mod_security2 module (i.e. `%m` token in https://httpd.apache.org/docs/2.4/mod/core.html#errorlogformat).

This pull request restores the APLOG_USE_MODULE, also adds it globally so all `ap_log_error` messages in all files are marked as belonging to mod_security2 module.

### DIFF
--- a/apache2/mod_security2.c
+++ b/apache2/mod_security2.c
@@ -25,12 +25,6 @@
 #include "apr_optional.h"
 #include "mod_log_config.h"
 
-/*
- * #ifdef APLOG_USE_MODULE
- * APLOG_USE_MODULE(security2);
- * #endif
- */
-
 #include "msc_logging.h"
 #include "msc_util.h"
 
--- a/apache2/modsecurity.h
+++ b/apache2/modsecurity.h
@@ -145,6 +145,9 @@ extern DSOLOCAL char *real_server_signature;
 extern DSOLOCAL char *chroot_dir;
 
 extern module AP_MODULE_DECLARE_DATA security2_module;
+#ifdef APLOG_USE_MODULE
+    APLOG_USE_MODULE(security2);
+#endif
 
 extern DSOLOCAL const command_rec module_directives[];
 


martinhsv:
Thanks for the contribution @erkia .

I expect this would resolve #1135 

The only thing I wonder about is whether the name 'security2' is ideal for this purpose. I.e. an apache error.log log line would look like:
```
... [security2:error] [pid ...
```

With a quick perusal, I wasn't able to identify any standard that is commonly adhered to.

I.e. Is the official name of the mod after the 'mod_' the common thing to do? Or would 'mod_security2' be more expected? Or perhaps something less based on the official mod name but more like how the engine is normally termed (like 'ModSecurity' -- as was floated as a possibility in #1135 )

Opinions and comments are welcome from the community at large. I'll leave this unmerged for at least a few more days for that purpose.

(One could argue the precise text isn't that critical, as long as we get something meaningful in there.)

erkia:
> With a quick perusal, I wasn't able to identify any standard that is commonly adhered to.

APLOG_USE_MODULE doesn't allow using any name for logging. It is used to select the module to which the source file belongs to and it is not possible to use anything other than `security2` unless the whole module is renamed.

https://github.com/apache/httpd/blob/669bdbb/include/http_config.h#L456

martinhsv:
Thanks @erkia

dune73:
Now that's a welcome fix! Thank you @erkia! If I knew it was that easy ...

marcstern:
This change breaks custom transformations (unresolved external symbol security2_module)

martinhsv:
Hi @erkia , do you have any thoughts on the issue raised by @marcstern ?

@marcstern: Could you perhaps describe a little bit what you mean by "custom transformations" and how they are structured/implemented?

marcstern:
If you try to compile the extensions in the "ext" directory, they contain
`#include "modsecurity.h"`
If modsecurity.h contains this APLOG_USE_MODULE statement, it tries to link with security2_module and you get that error.

erkia:
I was able to compile extensions in "ext", but I get the idea, what could fail. Alternative is to move `APLOG_USE_MODULE` out of  `modsecurity.h` to each source file which uses ap_log_* functions. I'll make the change.

erkia:
Suggested fix in https://github.com/SpiderLabs/ModSecurity/pull/2832

PR #2785, opened 2022-08-17
Fix: MULTIPART_INVALID_PART connected to wrong internal variable

martinhsv:
Despite this bug, MULTIPART_STRICT_ERROR was still being set correctly in this case. However, the 'msg' action in rule 200003 in modsecurity.conf-recommended would not show 'IP 1' properly.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Fix: MULTIPART_INVALID_PART connected to wrong internal variable
+    [Issue #2785 - @martinhsv]
   - Restore Unique_id to include random portion after timestamp
     [Issue #2752, #2758 - @datkps11, @martinhsv]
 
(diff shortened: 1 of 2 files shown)


elexito:
I keep getting this error - 

Unknown variable: &MULTIPART_PART_HEADERS

Please how do I update my modsecurity version?

Anytime I run sudo yum install mod_security or sudo apt install libapache2-mod-security2, it does not install the latest version which is v3.0.8

martinhsv:
Hello @elexito ,

There are two actively-maintained branches of ModSecurity: v2 (for use with Apache HTTP Server and IIS) and v3 (for use with nginx).

If you are using Apache, the latest version of ModSecurity for that is v2.9.7. MULTIPART_PART_HEADERS is supported in versions beginning with ModSecurity v2.9.6.

To use a more recent version of ModSecurity than what is available via apt, the main option is to build the software yourself. There are a number of online guides for this. For example: https://hostadvice.com/how-to/how-to-setup-modsecurity-for-apache-on-ubuntu-18-04/

elexito:
Thank you very much

PR #2788, opened 2022-08-20
Fix msc_transaction_cleanup function comment typo

lookat23:
Hi, I used the MoSecurity C API to discover the comments and code of the msc_transaction_cleanup function are not matching.


martinhsv:
Thanks @lookat23

PR #2791, opened 2022-08-24
Add JIT support for PCRE2

wfjsw:
Added JIT to PCRE2. 

I didn't thoroughly investigate the lifetime of the `Regex` class. The lifetime of the JIT result is attached to it.

Also it is worth noticing that PCRE2 is not following the `pcre_malloc` and `pcre_free` conventions, though I wonder what it will add if the allocation is scoped to requests.

### DIFF
--- a/src/utils/regex.cc
+++ b/src/utils/regex.cc
@@ -73,11 +73,7 @@ Regex::Regex(const std::string& pattern_, bool ignoreCase)
     PCRE2_SIZE erroroffset = 0;
     m_pc = pcre2_compile(pcre2_pattern, PCRE2_ZERO_TERMINATED,
         pcre2_options, &errornumber, &erroroffset, NULL);
-
     m_pcje = pcre2_jit_compile(m_pc, PCRE2_JIT_COMPLETE);
-    m_pmc = pcre2_match_context_create(NULL);
-    m_pcjs = pcre2_jit_stack_create(32*1024, 512*1024, NULL);
-    pcre2_jit_stack_assign(m_pmc, NULL, m_pcjs);
 #else
     const char *errptr = NULL;
     int erroffset;
@@ -97,8 +93,6 @@ Regex::Regex(const std::string& pattern_, bool ignoreCase)
 Regex::~Regex() {
 #if WITH_PCRE2
     pcre2_code_free(m_pc);
-    pcre2_match_context_free(m_pmc);
-    pcre2_jit_stack_free(m_pcjs);
 #else
     if (m_pc != NULL) {
         pcre_free(m_pc);
@@ -127,10 +121,10 @@ std::list<SMatch> Regex::searchAll(const std::string& s) const {
     do {
         if (m_pcje == 0) {
             rc = pcre2_jit_match(m_pc, pcre2_s, s.length(),
-                            offset, 0, match_data, m_pmc);
+                            offset, 0, match_data, NULL);
         } else {
             rc = pcre2_match(m_pc, pcre2_s, s.length(),
-                            offset, 0, match_data, m_pmc);
+                            offset, 0, match_data, NULL);
         }
         PCRE2_SIZE *ovector = pcre2_get_ovector_pointer(match_data);
 #else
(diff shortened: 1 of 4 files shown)


src/utils/regex.cc
```diff
@@ -73,6 +73,11 @@ Regex::Regex(const std::string& pattern_, bool ignoreCase)
     PCRE2_SIZE erroroffset = 0;
     m_pc = pcre2_compile(pcre2_pattern, PCRE2_ZERO_TERMINATED,
         pcre2_options, &errornumber, &erroroffset, NULL);
+
+    m_pcje = pcre2_jit_compile(m_pc, PCRE2_JIT_COMPLETE);
+    m_pmc = pcre2_match_context_create(NULL);
+    m_pcjs = pcre2_jit_stack_create(32*1024, 512*1024, NULL);
```
airween: Just my 2c question: why does it need the own stack handling?

I do not find the article now, where the author wrote unless there is a very compelling reason, you don't need to use `pcre_jit_stack_create()`.

Few months ago I've started to play with PCRE2 JIT implementation - you can find it [here](https://github.com/digitalwave/msc_retest/tree/pcre2support-jit-without-stack). In performance, there isn't any difference between controlling own stack with huge memory like this, or without own stack. 

I also checked Nginx, it also [uses](https://github.com/nginx/nginx/blob/master/src/core/ngx_regex.c) pcre2 JIT without own stack.

wfjsw: It is actually a copy-pasta from https://www.pcre.org/current/doc/html/pcre2jit.html#SEC10
I don't know if there is any penalty doing so, and I'm fine removing this.

wfjsw: Tried without PCRE2 jit stack and the memory usage dropped by 50%. That's nice.

I do worry about situations like https://www.mail-archive.com/bug-grep@gnu.org/msg05763.html

It depends on the rules, though.

martinhsv:
The main thing I would be concerned about here is the jit stack limit issue.

I don't have an opinion on whether the default of 32KB on the machine stack is large enough for normal (non-attack) use cases. But I can almost guarantee that either an attacker or a researcher will be able to breach that limit without using particularly excessive inputs (I'm thinking of some of the more complex CRS rules ).

The problem is that we are not handling this error scenario. PCRE2_ERROR_JIT_STACKLIMIT can effectively be treated the same as a simple 'no match' result -- which makes for a bypass opportunity.

(This new need overlaps with proposed per-match PCRE limits in #2736 , which I'll have to review soon.)

martinhsv:
The other possible alternative if PCRE2_ERROR_JIT_STACKLIMIT in encountered, is to retry the match using regular (non-JIT) matching.

FireBurn:
I've just tried this PR locally and I get the following error:

```
utils/regex.cc: In member function 'int modsecurity::Utils::Regex::search(const string&, modsecurity::Utils::SMatch*) const':
utils/regex.cc:294:24: error: 'rc' was not declared in this scope
     if (m_pcje != 0 || rc == PCRE2_ERROR_JIT_STACKLIMIT) {
                        ^~
utils/regex.cc:294:24: note: suggested alternative: 'rcmd'
     if (m_pcje != 0 || rc == PCRE2_ERROR_JIT_STACKLIMIT) {
                        ^~
                        rcmd
```

Adding an "int rc;" fixes things

wfjsw:
Ah my fault

martinhsv:
The current version looks ok. Thanks all for the various contributions. I'll plan to merge this.

martinhsv:
Hi @wfjsw ,

cppcheck is complaining that:
```
warning: src/operators/verify_cc.h,39,warning,uninitMemberVar,Member variable 'VerifyCC::m_pcje' is not initialized in the constructor.
```

The tool is correct. Unlike with the Regex class, the normal assignment to m_pcje occurs in the init() function rather than in the constructor.

Do you want to make that amendment to the VerifyCC class? (An initial value of PCRE2_ERROR_JIT_BADOPTION might be appropriate.)

martinhsv:
Unfortunately, it's still failing cppcheck ...
```
2022-12-11T21:07:44.1294490Z warning: src/utils/regex.cc,127,warning,uninitvar,Uninitialized variable: rc
2022-12-11T21:07:44.1295630Z warning: src/utils/regex.cc,175,warning,uninitvar,Uninitialized variable: rc
2022-12-11T21:07:44.1296350Z warning: src/utils/regex.cc,294,warning,uninitvar,Uninitialized variable: ret
2022-12-11T21:07:44.1296950Z warning: src/utils/regex.cc,327,warning,uninitvar,Uninitialized variable: rc
```

Don't worry about it. I can commit as-is and then fix it up.

PR #2792, opened 2022-08-31
fix no "echo -n" in macOS's sh

liudongmiao:
on macOS, the "/bin/sh" has no "-n" support.
There is $as_echo_n in autoconf, just use it.
And, there is AS_ECHO_N macro in autoconf too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -418,15 +418,18 @@ AX_VALGRIND_CHECK
 AC_OUTPUT
 
 
+if test -z "$as_echo_n"; then
+    as_echo_n='echo -n'
+fi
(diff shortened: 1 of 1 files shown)


martinhsv:
Overall, this is a reasonable change to pursue.

However, it seems like maybe AS_ECHO_N (that you have noted as an alternative) would be a bit cleaner and seems to be the more preferred/recommended way to do this these days.

martinhsv:
Hi @liudongmiao , thanks for the submission. Per my previous comment, a substitute PR ( #2894 ) has been used to resolve this. You were credited in the CHANGES file.

PR #2801, opened 2022-09-15
Fix two rule-reload memory leak issues

martinhsv:
This PR commits two of three issues originally submitted in https://github.com/SpiderLabs/ModSecurity/pull/2580.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 v3.x.y - YYYY-MMM-DD (to be released)
 -------------------------------------
 
+  - Fix two rule-reload memory leak issues
+    [Issue #2801 - @Abce, @martinhsv]
   - Correct whitespace handling for Include directive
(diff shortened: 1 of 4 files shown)


liudongmiao:
@martinhsv I suggest to avoid use:

```c++
if (transformations != NULL) {
    delete transformations;
}
```

It's confused to free memory of constructor argument.

In the similiar https://github.com/SpiderLabs/ModSecurity/pull/2728, I use `shared_ptr` to avoid `delete`.

For virtual `~Rule`, it's ok. I have explained other solutions too: https://github.com/SpiderLabs/ModSecurity/pull/2728#issuecomment-1229828008

martinhsv:
@liudongmiao ,

This way of deleting the transactions vector has some disadvantages, but it is that way in this PR because:
- that is how it was coded in the original PR, which this is replacing
- this way is consistent with how 'actions' was already being handled

PR #2806, opened 2022-10-04
Use AS_HELP_STRING instead of obsolete AC_HELP_STRING macro

hughmcmaster:
`AC_HELP_STRING` has been obsolete since at least Autoconf 2.53, which was released in 2003.


martinhsv:
Merged. Thanks @hughmcmaster

PR #2818, opened 2022-10-16
Use pkg-config to find libxml2, falling back to xml2-config if necessary

hughmcmaster:
Debian is taking steps to remove `xml2-config` in favour of `pkg-config`. This means ModSecurity will build without `libxml2` support by default on Debian, Ubuntu and other distributions tracking Debian packages.

This patch modifies `build/libxml.m4` to check for `libxml2` via `pkg-config`, falling back to `xml2-config` if necessary.

Patch for ModSecurity v2, as discussed in #2714.

### DIFF
--- a/build/find_xml.m4
+++ b/build/find_xml.m4
@@ -4,21 +4,7 @@ dnl Sets:
 dnl  LIBXML2_CFLAGS
 dnl  LIBXML2_LIBS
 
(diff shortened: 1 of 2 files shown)


hughmcmaster:
Here you go, @martinhsv.

martinhsv:
Merged. Thanks @hughmcmaster .

PR #2823, opened 2022-10-20
Move log opening to appropriate execution phase

### LINKED ISSUE #2822: ModSecurity can cause hang of apache httpd during graceful restart
**Describe the bug**

When modsec is configured to pipe logs to different process (like rotatelogs for example) it can cause hang of apache httpd during graceful restart.

When SecAuditLog directive is set as for example`SecAuditLog "|/usr/sbin/rotatelogs -l -n 3 /var/log/httpd/modsec_audit.log 50M"`
modsec opens pipe to rotatelogs during its configuration in `cmd_audit_log` function in apache2/apache2_config.c:1210

```
    if (dcfg->auditlog_name[0] == '|') {
        const char *pipe_name = dcfg->auditlog_name + 1;
        piped_log *pipe_log;

        pipe_log = ap_open_piped_log(cmd->pool, pipe_name);  <------------
        if (pipe_log == NULL) {
            return apr_psprintf(cmd->pool, "ModSecurity: Failed to open the audit log pipe: %s",
                    pipe_name);
        }
        dcfg->auditlog_fd = ap_piped_log_write_fd(pipe_log);
```
There is a problem that child processes created by httpd inherit the read side of pipe even when they are not using it and are meant to close them on startup which is ensured by read_handles global variable in server/log.c in apache httpd source.
For illustration here is code handling recording of the read handle:
```
/* remember to close this handle in the child process
 *
 * On Win32 this makes zero sense, because we don't
 * take the parent process's child procs.
 * If the win32 parent instead passed each and every
 * logger write handle from itself down to the child,
 * and the parent manages all aspects of keeping the
 * reliable pipe log children alive, this would still
 * make no sense :)  Cripple it on Win32.
 */
static void close_handle_in_child(apr_pool_t *p, apr_file_t *f)
{
#ifndef WIN32
    read_handle_t *new_handle;

    new_handle = apr_pcalloc(p, sizeof(read_handle_t));
    new_handle->next = read_handles;
    new_handle->handle = f;
    read_handles = new_handle;
#endif
}
```
These handles get cleared when plog memory pool gets cleared which happens after the configuration is loaded and before open logs
hooks are called. That means that child will get no reference to opened read side of the pipe and thus can not close it. This causes that process can wait for the write inside of the pipe indefinitely and will never get EPIPE from the OS informing the process about rotatelogs being already killed. This is why process waiting for write will hang.

**Logs and dumps**

Output of:
 1. DebugLogs (level 9)
 
[modsec_debug.log.tar.gz](https://github.com/SpiderLabs/ModSecurity/files/9829252/modsec_debug.log.tar.gz)

 2. AuditLogs

Empty because reproducing requires to stop rotatelogs process. 

 3. Error logs
 
[error_log.txt](https://github.com/SpiderLabs/ModSecurity/files/9830681/error_log.txt)

4. If there is a crash, the core dump file.

Backtrace seems sufficient to me.
#0  0x00007f560757259f in write () from target:/lib64/libc.so.6
#1  0x00007f560767aa1f in apr_file_write () from target:/lib64/libapr-1.so.0
#2  0x00007f560767ac4a in apr_file_write_full () from target:/lib64/libapr-1.so.0
#3  0x00007f56067a8ea1 in sec_auditlog_write.isra () from target:/etc/httpd/modules/mod_security2.so
#4  0x00007f5606784d21 in sec_audit_logger_native () from target:/etc/httpd/modules/mod_security2.so
#5  0x00007f560677c10c in modsecurity_process_phase () from target:/etc/httpd/modules/mod_security2.so
#6  0x00007f560676f9bd in hook_log_transaction () from target:/etc/httpd/modules/mod_security2.so
#7  0x000055f66b95ba98 in ap_run_log_transaction ()
#8  0x000055f66b96f89b in eor_bucket_cleanup ()
#9  0x00007f560767dade in apr_pool_destroy () from target:/lib64/libapr-1.so.0
#10 0x000055f66b964cc2 in eor_bucket_destroy ()
#11 0x000055f66b96fa89 in ap_core_output_filter ()
#12 0x000055f66b979dab in ap_process_request ()
#13 0x000055f66b97a073 in ap_process_http_connection ()
#14 0x000055f66b945e98 in ap_run_process_connection ()
#15 0x00007f5606ff3292 in child_main () from target:/etc/httpd/modules/mod_mpm_prefork.so
#16 0x00007f5606ff3607 in make_child () from target:/etc/httpd/modules/mod_mpm_prefork.so
#17 0x00007f5606ff3669 in startup_children () from target:/etc/httpd/modules/mod_mpm_prefork.so
#18 0x00007f5606ff3d93 in prefork_run () from target:/etc/httpd/modules/mod_mpm_prefork.so
#19 0x000055f66b9467c8 in ap_run_mpm ()
#20 0x000055f66b9346d5 in main ()

**To Reproduce**
This problem is tricky to reproduce so forgive me for a complex reproducer, repeat these steps:

1. Set SecAuditLog directive inside the modsec configuration to:
`SecAuditLog "|/usr/sbin/rotatelogs -l -n 3 /var/log/httpd/modsec_audit.log 50M"`
2. enable mpm prefork module like this:
`LoadModule mpm_prefork_module modules/mod_mpm_prefork.so`
3. Add mpm prefork module configuration
```
<IfModule prefork.c>
StartServers           1
MinSpareServers        1
MaxSpareServers        1
ServerLimit            1
MaxRequestWorkers      1
MaxRequestsPerChild    0
MaxConnectionsPerChild 0
</IfModule>
```
These steps make it easier to reproduce the issue
4. Create index page for httpd

`$ touch /var/www/html/index.html`

5. Start httpd (i am using systemd since its easier to keep track of processes)

`# systemctl start httpd`

6. Find rotatelogs process

`$ systemctl status httpd`

7. stop the rotatelogs process

`# kill -STOP <rotatelogs_pid>`

8. Perform requests to fill pipes buffer with modsecurity logs (you must have active rules that log something for this request)

`$ while true; do curl http://127.0.0.1 -m 5; [ $? -eq 28 ] && break; done`

9. Try to perform graceful restart

`# systemctl reload httpd`

10. Resume rotatelogs process

`# kill -CONT <rotatelogs_pid> `

11. Find httpd processes with systemctl and check whether one of them is not stuck

`$ systemctl status httpd`

`# for PID in '<httpd_pids>'; do pstack $PID >> ./pstack_logs; done;`

`$ cat ./pstack_logs | grep 'in write ('`

12. If grep find string `in write (` then one of the processes is still waiting for other process to read from previously created pipe.

**Expected behavior**

No hanging process.

**Server (please complete the following information):**
 - ModSecurity version (and connector): 2.9.6
 - WebServer: 2.4.54
 - OS (and distro): Linux Fedora Rawhide

 - Running any public or commercial rule set?
 Yes coreruleset mod_security_crs 3.3.4 from Fedora
 - What is the version number? [e.g. 2018-08-11]
3.3.4-1

**Additional context**
I am creating a PR that fixes this issue.
It is sufficient that the code which opens the auditlog is moved to a open_logs hook instead of being executed in a configuration command.

Please do not hesitate to ask further questions.

TomasKorbar:
When piped logs are opened during parsing of configuration it results in unexpected situations in apache httpd and can cause hang of process which is trying to log into auditlog.

Closes #2822

### DIFF
--- a/apache2/msc_logging.c
+++ b/apache2/msc_logging.c
@@ -2316,3 +2316,55 @@ void sec_audit_logger(modsec_rec *msr) {
     }
     #endif
 }
+
+static int open_audit_log(char *auditlog_name, unsigned char primary, apr_file_t **auditlog_fd,
+                          apr_fileperms_t *auditlog_fileperms, apr_pool_t *p) {
+    if (auditlog_name == NOT_SET_P) {
+        return OK;
+    }
+    if (auditlog_name[0] == '|') {
+        const char *pipe_name = auditlog_name + 1;
+        piped_log *pipe_log;
+
+        pipe_log = ap_open_piped_log(p, pipe_name);
+        if (pipe_log == NULL) {
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, NULL,
+                         "ModSecurity: Failed to open the %saudit log pipe: %s",
+                         primary ? "" : "secondary ", pipe_name);
+            return primary ? DONE : OK;
+        }
+        *auditlog_fd = ap_piped_log_write_fd(pipe_log);
+    }
+    else {
+        const char *file_name = ap_server_root_relative(p, auditlog_name);
+        apr_status_t rc;
+
+        if (*auditlog_fileperms == NOT_SET) {
+            *auditlog_fileperms = CREATEMODE;
+        }
+        rc = apr_file_open(auditlog_fd, file_name,
+                APR_WRITE | APR_APPEND | APR_CREATE | APR_BINARY,
+                *auditlog_fileperms, p);
+
+        if (rc != APR_SUCCESS) {
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, NULL,
+                         "ModSecurity: Failed to open the %saudit log file: %s",
+                         primary ? "" : "secondary ", file_name);
+            return primary ? DONE : OK;
+        }
+    }
+
+    return OK;
+}
+
+int modsec_open_logs(apr_pool_t *pconf, apr_pool_t *p, apr_pool_t *ptemp, server_rec *s_main) {
+    directory_config *dcfg = ap_get_module_config(s_main->lookup_defaults, &security2_module);
+
+    int primary_log_rc = open_audit_log(dcfg->auditlog_name, 1,
+                                        &dcfg->auditlog_fd, &dcfg->auditlog_fileperms, p);
+    if (primary_log_rc != OK) {
+        return primary_log_rc;
+    }
+    return open_audit_log(dcfg->auditlog2_name, 0,
+                          &dcfg->auditlog2_fd, &dcfg->auditlog_fileperms, p);
+}
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -1239,35 +1239,6 @@ static const char *cmd_audit_log(cmd_parms *cmd, void *_dcfg, const char *p1)
     directory_config *dcfg = _dcfg;
 
     dcfg->auditlog_name = (char *)p1;
-
-    if (dcfg->auditlog_name[0] == '|') {
-        const char *pipe_name = dcfg->auditlog_name + 1;
-        piped_log *pipe_log;
-
-        pipe_log = ap_open_piped_log(cmd->pool, pipe_name);
-        if (pipe_log == NULL) {
-            return apr_psprintf(cmd->pool, "ModSecurity: Failed to open the audit log pipe: %s",
-                    pipe_name);
-        }
-        dcfg->auditlog_fd = ap_piped_log_write_fd(pipe_log);
-    }
-    else {
-        const char *file_name = ap_server_root_relative(cmd->pool, dcfg->auditlog_name);
-        apr_status_t rc;
-        
-        if (dcfg->auditlog_fileperms == NOT_SET) {
-            dcfg->auditlog_fileperms = CREATEMODE;
-        }
-        rc = apr_file_open(&dcfg->auditlog_fd, file_name,
-                APR_WRITE | APR_APPEND | APR_CREATE | APR_BINARY,
-                dcfg->auditlog_fileperms, cmd->pool);
-
-        if (rc != APR_SUCCESS) {
-            return apr_psprintf(cmd->pool, "ModSecurity: Failed to open the audit log file: %s",
-                    file_name);
-        }
-    }
-
     return NULL;
 }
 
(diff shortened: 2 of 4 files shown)


martinhsv:
Hello @TomasKorbar ,

There is at least one downside here.

Current functionality means that the attempt to open the audit log file occurs at startup. If it fails -- perhaps because the chosen location is unwriteable -- then Apache will fail to start and the user can see a message like:
```
ModSecurity: Failed to open the audit log file: /var/log/apache/modsec_audit.log
```

(As a side note, with these changes I could not see a failure message anywhere, including in Apache's error.log. Perhaps I didn't look carefully enough.)

In general, it's preferable to have configuration errors identified at startup whenever possible.

TomasKorbar:
Hi @martinhsv 
Log about not accessible file is now present and failure to open log prevents apache from starting.
Let me know if there are any more changes needed.

TomasKorbar:
@martinhsv Hi, could we get this merged please? I can provide further help if needed.

apache2/msc_logging.c
```diff
@@ -2325,3 +2325,77 @@ void sec_audit_logger(modsec_rec *msr) {
...
+    }
+    if (dcfg->auditlog_name[0] == '|') {
+        const char *pipe_name = dcfg->auditlog_name + 1;
+        piped_log *pipe_log;
+
+        pipe_log = ap_open_piped_log(p, pipe_name);
+        if (pipe_log == NULL) {
+            ap_log_error(APLOG_MARK, APLOG_ERR, 0, NULL,
+                         "ModSecurity: Failed to open the audit log pipe: %s", pipe_name);
+            return -2;
```
martinhsv: This seems obscure -- here and at line 2361. What is '-2'? Is there an appropriate apache-defined named constant that could be used instead?

TomasKorbar: Yeah, you are right. i replaced -2 and 0 with appropriate macros from `httpd.h`

TomasKorbar:
@martinhsv should be good now.

eflanagan0:
@marcstern I've verified this builds correctly on a NixOS system after cherry-picking the commit to v2.7.3. I can also start an Apache/2.4.59 server loading the compiled module.

marcstern (changes requested):
Sounds sensible, but I wonder why it was handled differently than error logs at first. Does anybody see a potential problem with this fix?

Otherwise, I think it's the right moment to de-duplicate the code between dcfg->auditlog_name & dcfg->auditlog_name2 file opening. Can you use a static function that centralizes the common code?

Additional question (linked to code duplication):
Why do we have `pipe_name = dcfg->auditlog_name + 1;` and `pipe_name = ap_server_root_relative(p, dcfg->auditlog2_name + 1);`? Is this normal?

eflanagan0:
I realized this PR also will need rebased atop `v2/master` to incorporate the CI changes and kick off those checks.

TomasKorbar:
@marcstern Hi,
I added static function to deduplicate the code. Also the secondary pipe is now not evaluated relatively to server root, as i think that when you want to enter executable path, it is more useful to just specify absolute path, as the executable does not have to be in the same directory as server.

Please tell me about additional changes if necessary.

marcstern:
Using a hook looks indeed cleaner and more consistent with error log. Does anybody see a potential risk with this change?

airween:
> Using a hook looks indeed cleaner and more consistent with error log. Does anybody see a potential risk with this change?

Let me check this a bit later (probably in next few days).

TomasKorbar:
@airween @marcstern would you by any chance have time to look on this again?

airween:
Sorry, I completely forgot this. I just assigned this PR to myself, please let me review this soon.

Really sorry again!

notroj:
Moving this code to the hook and attaching the logs to `plog` looks exactly right to me to fix #2822 and it will make this code work like the logging code in httpd itself (e.g. `mod_log_config)`. Is there anything else required here to get this merged?

airween:
@TomasKorbar, @notroj: meanwhile I found a few small issues in the part of `standalone` code and test (regression test). Let me fix them in next few days, and then I can try this with those modifications. If everything will be okay, I'll merge this PR.

A question: this modification probably does not affect the standalone module, but I want to be sure, so please confirm this.

notroj:
I'm not familiar with the standalone build but it seems to stub out `ap_open_piped_log` so I guess this patch won't make any difference there.

https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/standalone/server.c#L618

uhliarik:
Hi, is there any progress on merging this PR? @airween

airween:
> Hi, is there any progress on merging this PR? @airween

Unfortunately not, I didn't have any time to take this issue.

PR #2831, opened 2022-11-14
Fix FILES_TMP_CONTENT collection key naming mechanism

airween:
It seems that in v3 the `FILES_TMP_CONTENT` collection key names get wrong value: it's the name of the user's file, not the name of the multipart part's name.

The documentation is a bit poor both in [v2](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v2.x)#FILES_TMP_CONTENT) and [v3](https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-(v3.x)#FILES_TMP_CONTENT), none of them mentioned the keys.

I just realized this behavior when I tested a rule, where I wanted to inspect a specified field in multipart request.

For eg. take a look to the v3's relevant [regression test](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/test/test-cases/regression/offset-variable.json#L1767-L1771):

```
        "----------------------------756b6d74fa1a8ee2",
        "Content-Disposition: form-data; name=\"filedata\"; filename=\"small_text_file1.txt\"",
        "Content-Type: text/plain",
        "",
        "This is a very small test file..",
```
The test uses [this rule](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/test/test-cases/regression/offset-variable.json#L1787):
```
        SecRule FILES_TMP_CONTENT:small_text_file1.txt "small" "id:1,phase:3,pass,t:trim,msg:'s'"
```
In ModSecurity v2, this example works with `FILES_TMP_CONTENT:filedata`. Btw I think using the file name specified by the user as keys makes no sense.

### DIFF
--- a/src/request_body_processor/multipart.cc
+++ b/src/request_body_processor/multipart.cc
@@ -1220,7 +1220,7 @@ int Multipart::multipart_complete(std::string *error) {
                 m->m_tmp_file_size.second,
                 m->m_tmp_file_size.first);
 
-            m_transaction->m_variableFilesTmpContent.set(m->m_filename,
+            m_transaction->m_variableFilesTmpContent.set(m->m_name,
                m->m_value, m->m_valueOffset);
 
             file_combined_size = file_combined_size + m->m_tmp_file_size.first;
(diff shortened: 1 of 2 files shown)


martinhsv:
Hi @airween. Thanks for that finding and the PR.

I agree. Although superficially one could argue that either the "name" or the "filename" could be a reasonable key for the file content, in practice the part "name" is almost always going to be the more meaningful from an application protection perspective.

PR #2832, opened 2022-11-15
Move APLOG_USE_MODULE out of modsecurity.h

erkia:


marcstern:
Putting it in apache2.h solves the problem in a cleaner way

erkia:
> Putting it in apache2.h solves the problem in a cleaner way

I kind of agree, but my first try to solve it in a cleaner way failed, so I'm a bit cautious to add it again in a header file...

marcstern:
I tried in msc_logging.h and it works: extensions can compile and all needed files include it

marcstern:
Indeed, apache2.h (as most of other .h files) is inherited as well

martinhsv:
All right, this PR seems like a reasonable way to proceed.

If no one raises any objections, I'll plan to merge this before the end of the day.

martinhsv:
Thanks @marcstern , for raising that detail on such a timely basis.

And thanks @erkia for providing the adjustment.

Issue #2815, opened 2022-10-13
Invalid memory access in parser_conn_limits_operator()


marcstern:
First problem (quite unusual, I admit):
```
config_orig_path = apr_pstrndup(mp, filename, strlen(filename) - strlen(apr_filepath_name_get(filename)));
apr_filepath_merge(&file, config_orig_path, param, APR_FILEPATH_TRUENAME, mp);
```
config_orig_path can be NULL, so 
```
config_orig_path = apr_pstrndup(mp, filename, strlen(filename) - strlen(apr_filepath_name_get(filename)));
if (!config_orig_path) {
   return apr_psprintf(mp, "ModSecurity: failed to duplicate filename in parser_conn_limits_operator");
}
apr_filepath_merge(&file, config_orig_path, param, APR_FILEPATH_TRUENAME, mp);
```

Second problem (I found it in prod, difficult to troubleshoot):
```
char* param = strchr(p2, ' ');
[...]
param++;
```
In case we use the SecConnReadStateLimit diective without operator (only a regex), paparm is NULL => memory fault, crash without any message.
Fix:
```
if (!param && *p2) return apr_psprintf(mp, "ModSecurity: Invalid operator for " \
   "SecConnReadStateLimit: %s, expected operators: @ipMatch, @ipMatchF " \
   "or @ipMatchFromFile with or without !", p2);
param++;
```

martinhsv:
Hi Marc,

Thanks for noting these.

Regarding the first of the two, there are more than a few places in v2 where a (freakishly unlikely) failure to allocate memory from a memory pool is not checked. These could perhaps be best handled en masse at some point.

Regarding the second item, just to expand on it a little:
- the problem input is expressly not supported; although the rx operator is implied in a SecRule directive if an operator is absent, the same is not true for SecConnReadStateLimit. Indeed, the documentation expressly states which three operators are supported and it is only the three (ipMatch, ipMatchF and ipMatchFromFile
- the failure occurs on startup rather than while running

That doesn't mean this second issue shouldn't be fixed. If an admin creates such an erroneous entry, diagnosing the problem is not at all friendly. I'll tentatively plan to include this in the next 2.9.x release.

PR #2834, opened 2022-11-20
Fix: handle error with SecConnReadStateLimit configuration

martinhsv:
Fix NULL-memory use case if an admin has not correctly included both an operator and a parameter (with a space-character separator).

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD mmm YYYY - 2.9.x (to be released)
 -------------------
 
+ * Fix: handle error with SecConnReadStateLimit configuration
+   [Issue #2815, #2834 - @marcstern, @martinhsv]
  * Only check for pcre2 install if required
    [Issue #2833 - @martinhsv]
  * Adjustment of previous fix for log messages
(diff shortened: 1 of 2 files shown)


martinhsv:
Addresses the main issue raised in #2815

marcstern:
The third parameter is optional, thus the test should be:
`if (!param && *p2)`
This will accept
- SecConnReadStateLimit 50 "@ipMatch 127.0.0.1" ...
- SecConnReadStateLimit 50
But will refuse
- SecConnReadStateLimit 50 "127.0.0.1" ...

martinhsv:
Hi @marcstern ,

Perhaps I'm not understanding what you mean.

But within parser_conn_limits_operator() we don't need to check whether p2 is NULL or not because parser_conn_limits_operator() will not be called at all if p2 is NULL.

And I did do some testing prior to the merge and it seemed to work per your summary. What use case do you believe is not working correctly?

marcstern:
Hi @martinhsv,
Indeed, there's a test before the call.
Sorry for the mistake.

martinhsv:
No worries.

Issue #2554, opened 2021-04-16
ipMatchFromFile does not support comments in v3


tomsommer:
`ipMatchFromFile` supported `#`-comments in v2, this is not possible in v3 and just reports a generic error

zimmerle:
Hi @tomsommer that is already fixed on v3/master. Yet to be released. For the reference 42a16c71cfb995bb2cd76dab02f6fe7aa6b596be

tomsommer:
This is still broken in 3.0.7

Blank lines are ignored, but lines that begin with # still throw errors with `ipMatchFromFile`

The fix in #1645 only applied to `pmFromFile`?

@zimmerle

feanorknd:
we need comments in the file ;)

daveram:
Issues like this are what prevent large enterprises and vendors like most of my clients from moving to mod_security v3

Comments in the file are used to denote which ranges are for what, so others don't have to go looking them up separately or note them elsewhere if we had to put in a rule against specific ranges and what they represent

martinhsv:
Hi all, thanks for re-raising this.

Per the observation by @tomsommer , this item appears to have been incorrectly closed since the fix for #1645 was for a different operator.

martinhsv:
If anyone wants to try it out, I have created a pull request for this at #2839 .

I will tentatively plan to merge it within a few days.

tomsommer:
@martinhsv any chance of getting this fix in a release?

martinhsv:
@tomsommer ,

Since it has already been merged to v3/master, it will be in the coming v3.0.x release by default.

tomsommer:
@martinhsv I meant can we get a release :)

martinhsv:
Hello @tomsommer ,

There is a v3.0.x release tentatively planned for the near-ish future (let's call it 5-7 weeks).

In the meantime feel free to build from source. (Or are you in a position where the policy at your installation only allows use of official releases?)

tomsommer:
@martinhsv built from source and this issue is indeed resolved. Found another issue though (#2877) :)

Thanks

PR #2839, opened 2022-12-01
Support comments in ipMatchFromFile file via '#' token

martinhsv:
Lines that have '#' as the first character will be ignored in the same way that zero-length lines are ignored.
Lines that have '#' after legitimate content, e.g.:
```
127.0.0.1#A comment
```
... will have the characters beginning with the '#' ignored.

There is no new whitespace handling in this PR; whitespace remaining after the removal of comments is treated the same as before.


martinhsv:
Closes #2554

PR #2840, opened 2022-12-07
Support for JIT option for PCRE2

martinhsv:
As with legacy PCRE, you need to include ```--enable-pcre-jit``` during the configure step to use this functionality.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD mmm YYYY - 2.9.x (to be released)
 -------------------
 
+ * Support for JIT option for PCRE2
+   [Issue #2840 - @martinhsv]
  * Use uid for user if apr_uid_name_get() fails
    [Issue #2046 - @arminabf, @marcstern]
  * Fix: handle error with SecConnReadStateLimit configuration
--- a/apache2/msc_pcre.c
+++ b/apache2/msc_pcre.c
@@ -84,7 +84,9 @@ void *msc_pregcomp_ex(apr_pool_t *pool, const char *pattern, int options,
         return NULL;
     }
 
-    /* TODO: Add PCRE2 JIT support */
+#ifdef WITH_PCRE_JIT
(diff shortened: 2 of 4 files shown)


S0obi:
Hey @martinhsv, thanks for this PR. Are you advising to use this flag in Modsec installations ? Did you notice any interesting gains in term of performance ? Any downsides ? Thanks in advance.

airween:
> Did you notice any interesting gains in term of performance ?

There is a small project which emulates the behavior of `@rx` operator, both for mod_security2 and libmodsecurity3: [msc_retest](https://github.com/digitalwave/msc_retest/tree/pcre2support-jit-without-stack) (please note, that this link points to a not-merged separated branch).

With this, you can try the performance (actually only in case of libmodsecurity3 - but I'm sure you will be able to compare the performance, and want to use the flag :smiley:). For more information see the attached README, or [this](https://coreruleset.org/20210106/introducing-msc_retest/) article.

martinhsv:
Hi @S0obi ,

I did not attempt to quantify the performance gains here -- much depends on the regular expressions that are being used and the strings that are matched against them. However, most installations would see some benefit (or at worst no performance reduction) in using the jit option.

(One exception to this could be if most of the regular expressions being used for matching are generated at runtime (for example by including "%{}" variables in the pattern). Since the extra up-front processing would not happen only once at startup, there is at least a chance that the any efficiency improvement during the match might be exceeded by the compile costs.)

martinhsv:
I am likely to merge this is soon, as is.

One extra thing to note is that actually making use of the PCRE2 JIT functionality depends not only on setting this option during the ModSecurity configure step, but that PCRE2 itself was built with the JIT option specified (it is not the default).

If you specify ```--with-pcre2 --enable-pcre-jit``` while configuring ModSecurity but your build of PCRE2 does not support JIT, nothing bad will happen. Processing in ModSecurity will fall back to non-JIT matching.

But, ModSecurity's configure step will not currently warn you that what you have asked for is not available. That would be a good enhancement that I will try to take a look at.

Issue #533, opened 2013-10-17
Apache ErrorDocument is not executed when error code has been rewritten in Phase 4 


rcbarnett:
MODSEC-385: (Same behavior which was written in https://www.modsecurity.org/tracker/browse/MODSEC-183.)

Basically the problem is that, although the right status code returns to the client, the appropriate ErrorDocument doesn't appear in a browser. Furthermore, there are no indication in strace that apache even tried to open any ErrorDocument page.

Setup:
Apache config has ErrorDocument 403 /error/403.html

Basic modsecurity configuration, two more line added:
SecRule REQUEST_URI attack "id:'1',phase:1,log,deny,status:403"
SecRule RESPONSE_STATUS "@eq 500" "phase:4,deny,status:403,id:1111"

Response to the first:
http://s18.postimage.org/hmr0y4sax/custom403working.png

Text form:
"http://192.168.110.139/attack

403

This is my custom error message page, every 403 should be end up with this. "

Response to the second:
http://s18.postimage.org/l530uit6x/custom403notworking.png

Text form:
"http://192.168.110.139/test.php

Forbidden

You don't have permission to access /test.php on this server.

Additionally, a 403 Forbidden error was encountered while trying to use an ErrorDocument to handle the request.
Apache/2.2.22 (Ubuntu) Server at 192.168.110.139 Port 80"

Test.php (this triggers an Error 500):
<?php
    require_once('unavailable file');
?>

It seems, in Phase 4 (as a matter of fact Phase 3 produces the same) response rewrite can't end on a custom ErrorDocument page for some reason. Even funnier, if I modify the line for the sake of testing as below:
SecRule RESPONSE_STATUS "@eq 500" "phase:4,deny,status:200,id:1111"

Response:
OK

The server encountered an internal error or misconfiguration and was unable to complete your request.

Please contact the server administrator, webmaster@localhost and inform them of the time the error occurred, and anything you might have done that may have caused the error.

More information about this error may be available in the server error log.

Additionally, a 200 OK error was encountered while trying to use an ErrorDocument to handle the request.

Apache/2.2.22 (Ubuntu) Server at 192.168.110.139 Port 80

If you need any more information please let me know.

bhafichuk-redfall:
I was wondering if there was any insight into a work around for this issue? http://sourceforge.net/p/mod-security/mailman/message/24053417/ seems to explain fairly closely the problem we are having.

[...]

PR #2850, opened 2022-12-22
Clear original response code in send_error_bucket function

### LINKED ISSUE #2849: Apache httpd does not execute ErrorDocument directive if response code is changed during phase 4
**Describe the bug**
When response code is changed during phase 4 and the original response code is other than 200, Apache httpd does not execute ErrorDocument directive and bad response is generated.

**Logs and dumps**

Output of:
 1. DebugLogs (level 9)
 
[modsec_debug.log](https://github.com/SpiderLabs/ModSecurity/files/10278739/modsec_debug.log)

 2. AuditLogs

[modsec_audit.log](https://github.com/SpiderLabs/ModSecurity/files/10278741/modsec_audit.log)

 3. Error logs
 
[error_log.txt](https://github.com/SpiderLabs/ModSecurity/files/10278744/error_log.txt)

**To Reproduce**

1. Add following configuration:
```
    ProxyPass /helloworld http://localhost:9090/unavaible
    ErrorDocument 403 /custompage.html
    SecRule RESPONSE_PROTOCOL "@contains HTTP" "id:'4',phase:4,auditlog,log,deny,status:403,msg:'yay!'"
```
2. Create and place custompage.html page to server root.
3. execute `$ curl -v localhost/helloworld`

**Actual behavior**
```
<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>403 Forbidden</title>
</head><body>
<h1>Forbidden</h1>
<p>You don't have permission to access this resource.</p>
<p>Additionally, a 503 Service Unavailable
error was encountered while trying to use an ErrorDocument to handle the request.</p>
</body></html>
```

**Expected behavior**
Proper rewritten response with custom error document:

**Server (please complete the following information):**
 - ModSecurity version (and connector): mod_security-2.9.6
 - WebServer: httpd-2.4.37
 - OS (and distro): Linux Fedora

 - Running any public or commercial rule set? No

**Additional context**

In my opinion, this is caused by apache httpds assumption that changed status of request which got through filter is a recursive error which occured during handling of previous errors.
See `ap_die_r` function in http module of apache, particularly this part of code:
```
    /*
     * The following takes care of Apache redirects to custom response URLs
     * Note that if we are already dealing with the response to some other
     * error condition, we just report on the original error, and give up on
     * any attempt to handle the other thing "intelligently"...
     */
    if (recursive_error != HTTP_OK) {
        while (r_1st_err->prev && (r_1st_err->prev->status != HTTP_OK))
            r_1st_err = r_1st_err->prev;  /* Get back to original error */

        if (r_1st_err != r) {
            /* The recursive error was caused by an ErrorDocument specifying
             * an internal redirect to a bad URI.  ap_internal_redirect has
             * changed the filter chains to point to the ErrorDocument's
             * request_rec.  Back out those changes so we can safely use the
             * original failing request_rec to send the canned error message.
             *
             * ap_send_error_response gets rid of existing resource filters
             * on the output side, so we can skip those.
             */
            update_r_in_filters(r_1st_err->proto_output_filters, r, r_1st_err);
            update_r_in_filters(r_1st_err->input_filters, r, r_1st_err);
        }

        custom_response = NULL; /* Do NOT retry the custom thing! */
    }
    else {
        int error_index = ap_index_of_response(type);
        custom_response = ap_response_code_string(r, error_index);
        recursive_error = 0;
    }
```
Clearing the original response code with `f->r->status = 200;` in `send_error_bucket` function appears to fix this issue.

This issue is a continuation of #533 and extends findings of Marc Stern in https://github.com/SpiderLabs/ModSecurity/issues/533#issuecomment-1270214868

TomasKorbar:
If this is left intact, then apache thinks that this code was generated during processing of ErrorDocument and does not handle it properly

Fix #2849

### DIFF
--- a/apache2/apache2_util.c
+++ b/apache2/apache2_util.c
@@ -31,6 +31,11 @@ apr_status_t send_error_bucket(modsec_rec *msr, ap_filter_t *f, int status) {
     /* Set the status line explicitly for the error document */
     f->r->status_line = ap_get_status_line(status);
 
+    /* Clear previously set response code to make clear that this is
+     * not a recursive error
+     */
+    f->r->status = 200;
+
     brigade = apr_brigade_create(f->r->pool, f->r->connection->bucket_alloc);
     if (brigade == NULL) return APR_EGENERAL;
 


TomasKorbar:
Could this be merged? I can provide further help if needed.

marcstern:
FYI: We're using it on prod for 2 months on several huge sites

martinhsv:
I haven't had opportunity to do a detailed analysis of this.

It looks simple, but this is the type of change that could potentially have unintended consequences -- perhaps only for some configurations and/or use cases.

Some questions that spring to mind immediately:
- why is the new line of code setting status to a hard-coded 200, when the assignment to status_line (just above the new code) has been set using the status value passed in to the function?
- is 200 really always the correct setting here? What if the original response was some other non-error response code like 301?

TomasKorbar:
> I haven't had opportunity to do a detailed analysis of this.
> 
> It looks simple, but this is the type of change that could potentially have unintended consequences -- perhaps only for some configurations and/or use cases.
> 
> Some questions that spring to mind immediately:
> 
>     * why is the new line of code setting status to a hard-coded 200, when the assignment to status_line (just above the new code) has been set using the status value passed in to the function?

I admit it seemed strange to me as well but fortunately i described the breaking point in #2849, so let me clarify.
There is a code in Apache httpd which assumes that when ap_die function is called and request record contains status code other than 200, then this error code was encountered while trying to access custom response URL. Setting response code to 200 here has in fact no effect on the final response because just after the check for recursive error, original response code gets overwritten by the output of filter.
```
r->status = type;
```

>     * is 200 really always the correct setting here? What if the original response was some other non-error response code like 301?

I think so, because it looks to me that send_error_bucket is used only when mod_security wants to change status code of the response and if the original status code, so it does not matter what the old response code was.

Please correct me if i am wrong.

TomasKorbar:
Hi @martinhsv ,
Just wanted to check in on my pull request - any updates or plans to merge it? Thanks!
Please let me know if there are further actions required from my side.

martinhsv:
Hi @TomasKorbar ,

No imminent plans. As I suggested previously, I'd need to gain higher confidence of the correctness of this change before merging. In the meantime, the PR is available for anyone who wishes to merge in their own environment to make use of it.

chnagy:
Hi, is there any chance this issue could be revisited?

It looks like this has been a known limitation since #533 was opened back in 2013, and the PR hasn’t seen any activity for a few years now. Having proper ErrorDocument handling seems like something a lot of production setups would need.

I understand the concerns about merging the PR without being 100% sure the change is the right approach. At the same time, keeping a custom patch around for an unmerged PR isn’t really a great long-term solution for people running ModSecurity in production.

Would be great to know if there are any plans to take another look at this or if there is a possible Workaround.

airween:
Hi @chnagy,

let me review this PR and if it's necessary I'll align it.

PR #2854, opened 2022-12-30
Escape error.log fields ('data', etc.) value

airween:
This patch escapes the log's `data` field, similar as mod_security2 does.

The aim is to help to process the content for log parsers.


airween:
May be we should add the new function [here](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/operators/operator.cc#L111-L128) too:

```diff
--- a/src/operators/operator.cc
+++ b/src/operators/operator.cc
@@ -112,17 +112,17 @@ std::string Operator::resolveMatchMessage(Transaction *t,
         if (m_couldContainsMacro == false) {
             ret = "Matched \"Operator `" + m_op + "' with parameter `" +
                 utils::string::limitTo(200, m_param) +
-                "' against variable `" + key + "' (Value: `" +
-                utils::string::limitTo(100,
-                    utils::string::toHexIfNeeded(value)) + \
+                "' against variable `" + 
+                utils::string::log_escape_hex(utils::string::limitTo(100, key)) + "' (Value: `" +
+                utils::string::log_escape_hex(utils::string::limitTo(100, value)) +
                 "' )";
         } else {
             std::string p(m_string->evaluate(t));
             ret = "Matched \"Operator `" + m_op + "' with parameter `" +
                 utils::string::limitTo(200, p) +
-                "' against variable `" + key + "' (Value: `" +
-                utils::string::limitTo(100,
-                    utils::string::toHexIfNeeded(value)) +
+                "' against variable `" +
+                utils::string::log_escape_hex(utils::string::limitTo(100, key)) + "' (Value: `" +
+                utils::string::log_escape_hex(utils::string::limitTo(100, value)) +
                 "' )";
         }
     }
```

I didn't find any occurrence where the user input appears.

There is a similar function [toHexIfNeeded()](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/utils/string.cc#L139-L152), but it does not encode `"`.

src/rule_message.cc
```diff
@@ -31,7 +31,7 @@ std::string RuleMessage::_details(const RuleMessage *rm) {
     msg.append(" [id \"" + std::to_string(rm->m_ruleId) + "\"]");
     msg.append(" [rev \"" + rm->m_rev + "\"]");
     msg.append(" [msg \"" + rm->m_message + "\"]");
-    msg.append(" [data \"" + utils::string::limitTo(200, rm->m_data) + "\"]");
+    msg.append(" [data \"" + utils::string::log_escape_hex(utils::string::limitTo(200, rm->m_data)) + "\"]");
```
martinhsv: We probably shouldn't restrict encoding of the double-quote character to only the 'data' item. That character could plausibly also appear in other items that can be filled via variable expansion in a rule. The most obvious additional item is 'msg', but 'tag' could also be affected.

src/utils/string.cc
```diff
@@ -267,6 +267,28 @@ void replaceAll(std::string *str, const std::string& from,
     }
 }
 
+std::string log_escape_hex(std::string s) {
```
martinhsv: The new functionality is highly comparable to the pre-existing function toHexIfNeeded().

The name for the new function proposed here doesn't really communicate what it's doing (the fn only cares about the double quote character) and it's unnecessarily dissimilar to the existing function.

I would suggest a name something like toHexSingleChar might be advantageous, along with an additional interface change:

It's pretty easy to make this new function at least slightly more generic by passing in the char we care about. I.e. the function can include a second parameter of type 'char' that indicates which character should be encoded when the function is called.

As for the guts of the function:

As stated earlier, the implementation details of this function are very similar to toHexIfNeeded(). It's preferable from a maintenance perspective not to adopt a different implementation methodology here. I.e. just copy the internals of the existing function and then change the 'if' condition.

martinhsv (commented):
See comments both for when a new function should be called, and comments about specifics about a new function.

src/rule_message.cc
airween: Yes, you are right. I reviewed mod_security2's code, and found that more field are also escaped beside the `[data]` and as you mentioned the `[msg]`. These are also escaped:

[[id]](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re.c#L2227-L2228), [[rev]](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re.c#L2231-L2232), [[msg]](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re.c#L2241-L2242), [[ver]](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re.c#L2272-L2273), [[tag]](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/re.c#L2298-L2299)

I think that would be fine to add these too - then the v3 logformat would be even more compatible with v2.

What do you think?

src/utils/string.cc
airween: > The new functionality is highly comparable to the pre-existing function toHexIfNeeded().

yes, I saw that, but I didn't want to modify.

> The name for the new function proposed here doesn't really communicate what it's doing (the fn only cares about the double quote character) and it's unnecessarily dissimilar to the existing function.

the name comes from v2 naming - it's the same.
 
> I would suggest a name something like toHexSingleChar might be advantageous, along with an additional interface change:
> 
> It's pretty easy to make this new function at least slightly more generic by passing in the char we care about. I.e. the function can include a second parameter of type 'char' that indicates which character should be encoded when the function is called.
> 
> As for the guts of the function:
> 
> As stated earlier, the implementation details of this function are very similar to toHexIfNeeded(). It's preferable from a maintenance perspective not to adopt a different implementation methodology here. I.e. just copy the internals of the existing function and then change the 'if' condition.

Right, I see. I'm going to review and modify.

src/rule_message.cc
martinhsv: [msg] and [tag] for sure, per my earlier comments.

[ver] ==> this is probably unlikely, but, it does support variable expansion, so: yes, include that one too
[rev] ==> even more obscure since it doesn't support population from untrusted input, but since an admin could, in theory, hard-code the double-quote character, I guess I'll say: yes, do this one too
[id] ==> I'll say no here. It's a bit debatable, but since there's no way to include a double-quote character here, I'd lean towards avoiding an extra specific encoding call.

airween: This is done by commit 105c590.

src/utils/string.cc
airween: Thanks for sharing your idea.

I modified the function, but a little differently than you described above. Hope you will like that.

> The name for the new function proposed here doesn't really communicate what it's doing (the fn only cares about the double quote character) and it's unnecessarily dissimilar to the existing function.

Not exactly. As I explained, I would have follow the behavior of v2 [log escape fn](https://github.com/SpiderLabs/ModSecurity/blob/v2/master/apache2/msc_util.c#L1304-L1307). That changes the characters below 31, above 127 or if it's the `"` or `\` (34 or 92).

The function [toHexIfNeeded()](https://github.com/SpiderLabs/ModSecurity/blob/v3/master/src/utils/string.cc#L139-L152) does the similar, except it does not escape `"` and `\`. Now I added a new prototype of the `toHexIfNeeded()` with an optional boolean argument, which marks the caller wants to encode special chars too (`"` and `\`). The old prototype is still exists, it calls the new one with `false` at the second argument.

Now it fits for the necessary behavior, and easy to understand it, I guess.

I keep this review unresolved, let me know if you have any additional note.

src/utils/string.h
```diff
@@ -72,6 +73,7 @@ void replaceAll(std::string *str, const std::string& from,
     const std::string& to);
 std::string removeWhiteSpacesIfNeeded(std::string a);
 std::string parserSanitizer(std::string a);
+std::string log_escape_hex(std::string s);
```
martinhsv: This should be removed, since a function by that name is no longer needed and there is no function definition.

src/utils/string.cc
```diff
@@ -135,13 +135,18 @@ std::string string_to_hex(const std::string& input) {
...
+
+std::string toHexIfNeeded(const std::string &str, bool escape_spec) {
+    // escape_spec: escape special chars or not
+    // spec chars: '"' (quotation mark, ascii 34), '\' (backslash, ascii 92)
     std::stringstream res;
 
     for (int i = 0; i < str.size(); i++) {
         int c = (unsigned char)str.at(i);
-        if (c < 32 || c > 126) {
+        if (c < 32 || c > 126 || (escape_spec == true && (c == 34 || c == 92))) {
```
martinhsv, line 145: This is an ok alternative to what I suggested in the earlier review, but see comment regarding having two functions.

src/utils/string.h
```diff
@@ -62,6 +62,7 @@ std::string limitTo(int amount, const std::string &str);
 std::string removeBracketsIfNeeded(std::string a);
 std::string string_to_hex(const std::string& input);
 std::string toHexIfNeeded(const std::string &str);
+std::string toHexIfNeeded(const std::string &str, bool escape_spec);
```
martinhsv: Rather than having two (overloaded) functions, with one calling the other, I think this would be a good case for only having a single function but with the second parameter defaulted. E.g. :

toHexIfNeeded(const std::string &str, bool escape_spec = false);

It's not a big deal but it makes the code slightly cleaner and avoids an extra function call on the call stack.

martinhsv (changes requested):

src/utils/string.h
airween: Oh, sorry - removed. Thanks.

src/utils/string.h
airween: Thanks, I re-factorized these methods as you suggested here, see 6dd00be. I keep this review open.

martinhsv (approved):
All good. Thanks.

PR #2797, opened 2022-09-07
Multipart parsing fixes and new MULTIPART_PART_HEADERS collection

martinhsv:

PR #2857, opened 2023-01-04
Fix: FILES_TMP_CONTENT may sometimes lack complete content

martinhsv:


dune73:
Thank you @martinhsv.

coldtobi:
This issue have been assigned  CVE-2023-24021.

dune73:
Oh wow. I was not aware somebody was intending to assign a CVE to this. Do you have any information @coldtobi or did you launch the process @martinhsv?

It's a bit unfortunate it comes out of the blue, since we could have written a better advisory than the one now listed at NIST. https://nvd.nist.gov/vuln/detail/CVE-2023-24021 Namely because the problem is bigger than it seems.

What I do not like about the situation is that the Changelog makes it look rather innocent, when it can be abused for a buffer overflow. This gives attackers taking a deeper look an advantage over users who read the changelog and think it's no big deal.

coldtobi:
@dune73 I'm currently preparing a security update of modsecurity-crs for Debian LTS/ELTS  and for that I need also to update modsecurity-apache (due to #2797).  During that  I learned about this issue and my LTS/ELTS security team collegues asked me to file an CVE myself, so that I'll be able to tackle this issue.

The description is unfortunately the only bit of information I had about the implications, and I'm not happy with it either. If you have something better,   send it my way ;-) (tobi at debian.org) and I will try to get the description updated via MITRE.

dune73:
Will do. Thank you Tobi.

martinhsv:
A recent comment posted in this PR contains some incorrect information alleging buffer overflow.

The confirmed issue related to this PR is as described in the PR Title -- and more fully discussed in the release notes published here: https://www.trustwave.com/en-us/resources/security-resources/software-updates/announcing-modsecurity-version-297/ 

The original finder was gieltje, while a fix was composed by @airween, which was subsequently merged with only minor modification. Thanks to both.

But no buffer overflow related to this bug has ever been demonstrated, nor has anyone claimed to have been able to cause a buffer overflow as a result of this bug.

dune73:
I stand corrected. 

Using the term "buffer over-read" (CWE-126) is more appropriate than "buffer overflow" (CWE-121 / CWE-122). Thank you for pointing this out @martinhsv.

PR #2862, opened 2023-01-18
Fix: Setting of variable REQUEST_FILENAME

PrajwalKrishna:
Currently `REQUEST_FILENAME` variable includes the domain name from the URL while according to the documentation this is not the expected behaviour.

The details of the existing bug can be found on an issue opened for the same - https://github.com/SpiderLabs/ModSecurity/issues/2853

### DIFF
--- a/test/test-cases/regression/variable-REQUEST_FILENAME.json
+++ b/test/test-cases/regression/variable-REQUEST_FILENAME.json
@@ -74,11 +74,12 @@
       ]
     },
     "expected":{
+      "http_code": 403,
       "debug_log":"Target value: \"/one/two/login.php\" \\(Variable: REQUEST_FILENAME\\)"
     },
     "rules":[
       "SecRuleEngine On",
-      "SecRule REQUEST_FILENAME \"@contains test \" \"id:1,phase:3,pass,t:trim\""
+      "SecRule REQUEST_FILENAME \"@eq /one/two/login.php \" \"id:1,phase:3,deny,status:403,t:trim\""
     ]
   }
 ]
--- a/test/test-cases/regression/variable-REQUEST_FILENAME.json
+++ b/test/test-cases/regression/variable-REQUEST_FILENAME.json
@@ -39,5 +39,46 @@
       "SecRuleEngine On",
       "SecRule REQUEST_FILENAME \"@contains test \" \"id:1,phase:3,pass,t:trim\""
     ]
+  },
+  {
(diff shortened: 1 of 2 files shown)


test/test-cases/regression/variable-REQUEST_FILENAME.json
```diff
@@ -39,5 +39,46 @@
...
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "debug_log":"Target value: \"/one/two/login.php\" \\(Variable: REQUEST_FILENAME\\)"
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule REQUEST_FILENAME \"@contains test \" \"id:1,phase:3,pass,t:trim\""
```
airween: This makes no sense. In this test variable `REQUEST_FILENAME` does not contain pattern `test`. Would be fine to place a rule which shows the expected result (furthermore you should use some disruptive action like `deny` and check the result in the `expected` key)

Just my 2 cents...

test/test-cases/regression/variable-REQUEST_FILENAME.json
```diff
@@ -74,11 +74,12 @@
       ]
     },
     "expected":{
+      "http_code": 403,
       "debug_log":"Target value: \"/one/two/login.php\" \\(Variable: REQUEST_FILENAME\\)"
     },
     "rules":[
       "SecRuleEngine On",
-      "SecRule REQUEST_FILENAME \"@contains test \" \"id:1,phase:3,pass,t:trim\""
+      "SecRule REQUEST_FILENAME \"@eq /one/two/login.php \" \"id:1,phase:3,deny,status:403,t:trim\""
```
airween: Again just my 2 cents... :smile: 
To compare two strings, you should use `@strEq` instead of `@eq`.

airween:
Hi @PrajwalKrishna,

thanks for your first time contribution!

The mentioned variable has changed [meantime](https://github.com/owasp-modsecurity/ModSecurity/pull/3048), could you check that the behavior that you explained is still exists?

PrajwalKrishna:
Thanks @airween seems that the variable has been fixed. Would be closing this PR unless you think it is a good idea to make this test part of UTs.

airween:
Hi @PrajwalKrishna, thanks for possibility, now I'm closing this PR, I think the behavior is what we expect.

PR #2870, opened 2023-01-28
Set correct line end in case of `MULTIPART_PART_HEADER` variable

airween:
Looks like variable `MULTIPART_PART_HEADER` variable contains the EOL characters.

This can lead false positive, and we (CRS team) discovered with help of an issue [3046](https://github.com/coreruleset/coreruleset/issues/3046). There is a full detail of the unexpected behavior, with many examples.

Here is what I get in the debug.log:

```
Multipart: Added part header "Content-Disposition" "form-data; name="file"; filename="New Text Document.txt"".
Multipart: Added part header "Content-Type" "text/plain; charset=utf-8".
Multipart: Content-Disposition name: file.
Multipart: Content-Disposition filename: New Text Document.txt.
MultipartPartTmpFile: Create filename= /tmp/20230128-212401-16749374418.382181-file-uEZjC1
Multipart: Created temporary file 1 (mode o0): /tmp/20230128-212401-16749374418.382181-file-uEZjC1
Multipart: Added part header line:Content-Disposition: form-data; name="file"; filename="New Text Document.txt"

Multipart: Added part header line:Content-Type: text/plain; charset=utf-8

```

As it seems, the log contains the `Added part header line:` with the header of the part of multipart payload. Both lines ended with EOL - but not as like in case of `Added part header`.

After the fix, the same part of debug log is this:

```
Multipart: Added part header "Content-Disposition" "form-data; name="file"; filename="New Text Document.txt"".
Multipart: Added part header "Content-Type" "text/plain; charset=utf-8".
Multipart: Content-Disposition name: file.
Multipart: Content-Disposition filename: New Text Document.txt.
MultipartPartTmpFile: Create filename= /tmp/20230128-213008-167493780825.319085-file-89EDD6
Multipart: Created temporary file 1 (mode o0): /tmp/20230128-213008-167493780825.319085-file-89EDD6
Multipart: Added part header line:Content-Disposition: form-data; name="file"; filename="New Text Document.txt"
Multipart: Added part header line:Content-Type: text/plain; charset=utf-8
```

The lines are without `\r\n`, and both variables (part header and part header line) are the same. This is the same behavior as in case of mod_security2 (this is why I thought it's an unexpected behavior).

I added a new test case, anyone can check the regression test if picks up that commit.


Kalakaarboyz1 (approved):

src/request_body_processor/multipart.cc
```diff
@@ -935,6 +935,7 @@ int Multipart::process_part_header(std::string *error, int offset) {
                 "Multipart: Added part header \"" + header_name \
                 + "\" \"" + header_value + "\".");
             if (len_without_termination > 0) {
+                m_buf[len_without_termination] = '\0';
```
martinhsv: Your diagnosis appears to be correct, but I don't think we want to solve it this way.
 
m_buf is the unparsed buffer of data. It's typically undesirable to overwrite the pre-parsed buffer when extracting values from it. Although I don't see any tangible issues that would occur with the current code base, it may set us up for a problem later if someone does not realize that the original buffer has been altered.

Perhaps consider instead merely altering the assignment line to use the overloaded function that includes length? E.g.

m_mpp->m_last_header_line.assign(m_buf, len_without_termination);

Since it is the extracted value that needs the adjusted length, that's the right place to do it.

martinhsv (changes requested):

src/request_body_processor/multipart.cc
airween: > Perhaps consider instead merely altering the assignment line to use the overloaded function that includes length? > E.g. 
> 
> m_mpp->m_last_header_line.assign(m_buf, len_without_termination);

You are right, I'm going to review the PR and change this.

airween: Thanks for the feedback, it's done - see 6623c0ae.

martinhsv (approved):
This all looks fine now. Thanks @airween

PR #2883, opened 2023-03-14
build: Avoid bashisms

Polynomial-C:
otherwise configure fails to find pcre2 when /bin/sh does not point to bash:

```
  configure: using pcre v8.45
  ./configure: 16601: test: xno: unexpected operator
  ./configure: 16601: test: xno: unexpected operator
  checking for libpcre2 config script... no
  configure: *** pcre2 library not found.
  configure: error: pcre2 library is required
```


martinhsv:
Thanks @Polynomial-C

PR #2890, opened 2023-04-10
Fix: possible segfault on startup if duplicate ip+CIDR in ip match list

martinhsv:


src/utils/msc_tree.cc
```diff
@@ -259,6 +259,7 @@ int InsertNetmask(TreeNode *node, TreeNode *parent, TreeNode *new_node,
 
             node->count++;
             node->netmasks = reinterpret_cast<unsigned char *>(malloc(node->count * sizeof(unsigned char)));
+            memset(node->netmasks, 0, (node->count *  sizeof(unsigned char)));
```
mirkodziadzka-avi, line 262: @martinhsv This memset should probably go below the null check 2 lines down. Same for the other 2 in this PR.

src/utils/msc_tree.cc
```diff
@@ -410,6 +411,7 @@ TreeNode *CPTAddElement(unsigned char *ipdata, unsigned int ip_bitmask, CPTTree
                 node->count++;
                 new_node = node;
                 node->netmasks = reinterpret_cast<unsigned char *>(malloc(node->count *  sizeof(unsigned char)));
+                memset(node->netmasks, 0, (node->count *  sizeof(unsigned char)));
```
mirkodziadzka-avi, line 414: Here is another potential crash if we are out of memory.  Malloc above may return null and there is no check below for this condition.

martinhsv:
Thanks @mirkodziadzka-avi . The location of the first memset has been corrected via #2891

PR #2892, opened 2023-04-20
Avoid IP duplicates in tree

marcstern:
When you merge shared configs or you use mod_define (or core Define) to combine several IP depending on the environment, you can easily have many redundant IP in the tree. This can sometimes lead to a huge number of IP, most of them being redundant.
This one-line PR (3 lines for clarity) simply checks if it exists before adding it to the tree.


marcstern:
Sorry, this doesn't look correct.
I'm not sure where to perform the check exactly.

PR #2898, opened 2023-05-05
Add editorconfig to help OSS contributors

brandonpayton:
When switching between many different open source code bases, it can be cumbersome to constantly change editor settings in order to respect a given project's basic style preferences. EditorConfig is a popular format for representing those preferences so editors can automatically respect them.

Many editors support EditorConfig out of the box, and many others have EditorConfig plugins:
https://editorconfig.org/#pre-installed

This commit adds an EditorConfig to the ModSecurity project. It is most likely not perfect because the chosen indentation varies a bit throughout the project, but hopefully it provides a good basis for most ModSecurity editing.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,26 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file
+[*]
+end_of_line = lf
+indent_style = space
+# Scripts without suffixes in the project root tend to indent by two spaces
+indent_size = 2
+
+# Most of the project files indent by four spaces
+[*/**]
+indent_size = 4
+
+# Test files indent by two spaces
+[test/**]
+indent_size = 2
+
+# The config parser file indents by both two and four spaces,
+# so we choose to indent by two spaces as a common denominator.
+[*.yy]
+indent_size = 2
+
+[{Makefile,Makefile.am}]
+indent_style = tab
+


airween:
Hi @brandonpayton,

thanks for sharing this idea.

Generally I'm not against of this, but as you wrote:

> It is most likely not perfect because the chosen indentation varies a bit throughout the project

so I don't know how useful is this.

If we want to maintain the indentation through the whole code, we should choose a tool and implement it - eg. [astyle](https://astyle.sourceforge.net/).

Perhaps this is a good topic for discussion.

theseion:
IMO, formatters and editorconfig are complementary. Depending on your IDE, you may be able to run the formatter often, e.g., on save, but it's different that using editorconfig because it formats the code after you've already written it. editorconfig changes how you write code in the first place. You can still fix up things with a formatter after the fact but if both are close in their configuration, I think that helps. It's certainly less disruptive than seeing the code you just wrote being rearranged all the time.

My 2 cents: add an editorconfig. It doesn't hurt.

marcstern:
editorconfig looks a very good start. And a formatter like astyle could be added as well later.
With good policies, the code could be easier to read (and so to maintain).
And it could definitely help with PR.

airween:
Thanks @theseion and @marcstern, I'm going to merge this soon.

@marcstern: do you think could we add this to `v2/master` too?

airween (approved):
LGTM

brandonpayton:
Thank you, @airween!

PR #2912, opened 2023-06-06
Do not escape special chars in regex pattern with macro

### LINKED ISSUE #2357: Incorrect escaping in @rx operator with macro expansion
When using macro expansion with @rx operator, escaping is incorrectly applied.

Example:
 SecRule ARGS "@rx localhost|%{TX.OtherHosts}"
will lead to something like  SecRule ARGS "@rx localhost\\|myhost1\\|myhost2"
This screws up all regexes, as regex special characters will be escaped, even outside the macro-expanded varaiables.

In case I want to test against %{TX.MyHost} that isn't a regex, I can use the @contain operator.

martinhsv:
Closes #2357

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,6 +1,8 @@
 DD mmm YYYY - 2.9.x (to be released)
 -------------------
 
+ * Do not escape special chars in rx pattern with macro
+   [Issue #2357 - @marcstern, @martinhsv]
  * Substitute two equals-equals operators in build
    [Issue #2883 - @Polynomial-C]
 
--- a/apache2/re_operators.c
+++ b/apache2/re_operators.c
@@ -1,6 +1,6 @@
 /*
 * ModSecurity for Apache 2.x, http://www.modsecurity.org/
-* Copyright (c) 2004-2022 Trustwave Holdings, Inc. (http://www.trustwave.com/)
+* Copyright (c) 2004-2023 Trustwave Holdings, Inc. (http://www.trustwave.com/)
 *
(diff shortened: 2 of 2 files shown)


marcstern:
Thanks for the fix Martin.
The strdup can be avoided; we can use directly re_pattern->value:
`regex = msc_pregcomp_ex(msr->mp, re_pattern->value, PCRE_DOTALL | PCRE_DOLLAR_ENDONLY, &errptr, &erroffset, msc_pcre_match_limit, msc_pcre_match_limit_recursion);`

martinhsv:
Yes, I had noted the possibility of omitting the additional duplication.

However, I opted to leave the pre-existing model in place (by which I mean, to the extent that it creates yet another copy of the string to pass to the regex engine) because of the data models that are in use.

The function expand_macros() populates re_pattern, which is of type msc_string. And in such structures, the 'value_len' can (in principle) be greater than the null-terminated length of 'value'. Also, just thinking from a code-safety perspective, one might wonder whether a code error (now or in the future) might result in an msc_string where the value_len is correct, but the buffer pointed to by 'value', is (incorrectly) not null-terminated (there is no such bug now -- expand_macros() will always correctly null-terminate the returned string, but, still, there can sometimes be merit in safeguarding against some future bug).

Although it makes no functional difference right now, I suspected that it might be slightly cleaner/clearer to logically differentiate the variable as an msc_string from what will be passed to the regex engine. I may revisit that prior to merge.

PR #2930, opened 2023-07-17
Don't spray XML parser errors to stderr

erkia:
Currently, all XML parser errors are written to stderr, which in case of Apache means that the messages end up randomly in error_log, unformatted. 

This pull request adds custom XML error handling function, which just discards all the messages at the moment.

### DIFF
--- a/apache2/msc_xml.c
+++ b/apache2/msc_xml.c
@@ -19,6 +19,9 @@ xml_unload_external_entity(const char *URI, xmlCharEncoding enc)    {
     return NULL;
 }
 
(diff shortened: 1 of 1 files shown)


martinhsv:
Hello @erkia ,

Could you provide a couple of examples of such messages? Ideally include xml bodies to  reproduce them and whether any special configuration is required to reproduce the issue (e.g. higher LogLevel in the apache config).

marcstern:
Non-standard messages in error log are not clean, but no message at all isn't good either.
We should use msr_log(msr, ...), no?

PR #2813, opened 2022-10-11
fix(build): update pcre2 autoconf macro and dependencies

fzipi:
This PR fixes pcre2 builds for extended architectures. In summary:

- moved from custom m4 to using standard macro from autoconf-archive
- updated dependent automake files
- fixed calls to echo in configure.ac to standard, to be more portable

Signed-off-by: Felipe Zipitria <felipe.zipitria@owasp.org>

Fixes #2810

Issue #2810, opened 2022-10-09
PCRE2 library is not found in all debian supported architectures


fzipi:
**Describe the bug**

We are building modsecurity containers for various architectures. While adding support for pcre2 in newer versions, we found that for some architectures the pcre2 config is not found.

Example builds failing here: [linux/arm64/v8](https://github.com/coreruleset/modsecurity-docker/actions/runs/3210566616).

This happens probably because the package in debian is called `libpcre2-8` (see package [dev files list](https://packages.debian.org/sid/amd64/libpcre2-dev/filelist)), and here are the results of installing and querying in the nginx default debian based container:

```
root@04e33073eeea:/# apt-get install libpcre2-dev
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  libpcre2-16-0 libpcre2-32-0 libpcre2-8-0 libpcre2-posix2
The following NEW packages will be installed:
  libpcre2-16-0 libpcre2-32-0 libpcre2-dev libpcre2-posix2
The following packages will be upgraded:
  libpcre2-8-0
1 upgraded, 4 newly installed, 0 to remove and 21 not upgraded.
Need to get 1483 kB of archives.
After this operation, 3997 kB of additional disk space will be used.
Do you want to continue? [Y/n] Y
Get:1 http://deb.debian.org/debian bullseye/main amd64 libpcre2-8-0 amd64 10.36-2+deb11u1 [248 kB]
Get:2 http://deb.debian.org/debian bullseye/main amd64 libpcre2-16-0 amd64 10.36-2+deb11u1 [232 kB]
Get:3 http://deb.debian.org/debian bullseye/main amd64 libpcre2-32-0 amd64 10.36-2+deb11u1 [220 kB]
Get:4 http://deb.debian.org/debian bullseye/main amd64 libpcre2-posix2 amd64 10.36-2+deb11u1 [49.1 kB]
Get:5 http://deb.debian.org/debian bullseye/main amd64 libpcre2-dev amd64 10.36-2+deb11u1 [733 kB]
Fetched 1483 kB in 0s (8869 kB/s)
debconf: delaying package configuration, since apt-utils is not installed
(Reading database ... 20501 files and directories currently installed.)
Preparing to unpack .../libpcre2-8-0_10.36-2+deb11u1_amd64.deb ...
Unpacking libpcre2-8-0:amd64 (10.36-2+deb11u1) over (10.36-2) ...
[...]

PR #2939, opened 2023-08-02
Configure: add additional name to pcre2 pkg-config list

martinhsv:


martinhsv:
This change originated from #2810 and #2813

PR #2940, opened 2023-08-08
V2/mst/nullcheck

marcstern:
Added several check for NULL pointers


marcstern:

PR #2944, opened 2023-08-11
Support for "filename*" in multipart (precedence over "filename")

marcstern:
Reference: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition


marcstern:
Error: it was meant to be merged with v2

PR #2946, opened 2023-08-11
Fix for https://github.com/SpiderLabs/ModSecurity/issues/610

marcstern:


marcstern:

PR #2947, opened 2023-08-11
status 400 instead of 500 on XML/JSON parsing error

marcstern:


marcstern:

marcstern:

PR #2948, opened 2023-08-11
Remove redundant tags (and actions)

marcstern:
This removes redundant tags like tag:xss,tag:bla,tag:xss.
Very useful when using mod_macro or Define, so the same tag may be added at several places.
Does the same for other actions like logdata.
Detects redundancy only if the content of the action is 100% the same.

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -972,7 +972,7 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
     #endif
 
     /* Add rule to the recipe. */
-    if (msre_ruleset_rule_add(dcfg->ruleset, rule, rule->actionset->phase) < 0) {
+    if (rule->actionset && msre_ruleset_rule_add(dcfg->ruleset, rule, rule->actionset->phase) < 0) {
         return "Internal Error: Failed to add rule to the ruleset.";
     }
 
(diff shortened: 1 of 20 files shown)


martinhsv:
Hi @marcstern ,

Isn't this pretty similar to #2685 which we have discussed previously?

As before, my general leaning is that the ModSecurity engine should avoid second-guessing what the rule writers have done by doing autocorrections.

marcstern:
Hi @martinhsv, it's indeed a formal PR for what we discussed.
I know you were not in favor of this, but I'm proposing it anyway. In this case, it's really a simple clean-up that saves RAM and CPU, so I really think it has a value.
Also keep in mind that rule writers, in professional or hosted environments, don't have the full control like you have when working on a stand-alone version. You inherit (part of) rules that you can extend but you don't always have the full picture and, often, you combine different pieces that have to set a tag but can be combined together as well.

martinhsv:
Hello @marcstern ,

I'm going to go ahead and close this.

For the main proposal -- which you have agreed is identical to your earilier PR -- it is for the same reasons. While I understand the outcome of #2685 was not your own preference, I did articulate my reasons.

Note that the current version of this PR contains a lot of disparate changes unrelated to the stated purpose of the PR. PRs that are either very large and/or contain changes unrelated to each are problematic to manage.

PRs are more likely to be actively considered if they are both tightly focussed and result in a tangible improvement for the users.

PR #2949, opened 2023-08-11
Regex in setvar variables (issues 2927)

marcstern:
Regex in setvar variables
https://github.com/SpiderLabs/ModSecurity/issues/2927


marcstern:

PR #2950, opened 2023-08-11
Logging enhancement: in case of no id, show file info

marcstern:
Logging enhancement:
- show if rule is chained
- in case of no id, show file info


marcstern:
This obsoletes https://github.com/owasp-modsecurity/ModSecurity/pull/2940

marcstern:
Corrupted. Replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3052

PR #2955, opened 2023-08-11
Handle capture as tx.1=char in validateByteRange

marcstern:


martinhsv:
Hello @marcstern ,

Could you please describe what problem this is intended to address?

marcstern:
This allows to know what's the offending character.
Very useful in complex configurations with multiple <locationmatch>, defined variables, macros, etc.

martinhsv:
So this is recording every finding, right?

I'm a little hesitant about such expansive use of the tx.0, tx.1, ... special variables. Most operators that support capture use only tx.0 or rx, of course, which may use all of tx.0 through tx.9. But here, (unless I'm mistaken) there could potentially be thousands.

The rest of the code assumes that there are a maximum of 10 of these variables. At a minimum we would have a problem because items tx.11 and greater would not currently get reset via apr_table_unset as happens now with tx.0 through tx.9.

marcstern:
What about limiting it to 9 then?
It would be easy: if (capture && count < 10)

marcstern:
I adapted the PR

marcstern:

PR #2958, opened 2023-08-14
Check return code of apr_procattr_io_set()

marcstern:


martinhsv:
Merged. Thanks @marcstern

PR #2965, opened 2023-08-16
Fix for issue 2849

marcstern:
Fix ErrorDocument processing if status is changed during phase 4


marcstern:
Duplicate of https://github.com/SpiderLabs/ModSecurity/pull/2850

PR #2971, opened 2023-08-21
Update verify_ssn.cc to fix new format

### LINKED ISSUE #2970: verifySSN: Area code can be larger than 740
The United States Social Security Number area code used to startwith a number less than 740 bt this is no longer the case:

https://www.ssa.gov/history/ssn/geocard.html

There are now routinely SSNs with area code larger than 740.

https://github.com/SpiderLabs/ModSecurity/blob/60f802e4801c8a4fee8e2caac90462e53651971f/src/operators/verify_ssn.cc#L103

jakubsuchy:
Fixes https://github.com/SpiderLabs/ModSecurity/issues/2970

### DIFF
--- a/src/operators/verify_ssn.cc
+++ b/src/operators/verify_ssn.cc
@@ -100,7 +100,7 @@ bool VerifySSN::verify(const char *ssnumber, int len) {
         goto invalid;
 
     /* More tests */
-    if (area >= 740 || area == 666)
+    if (area == 666)
         goto invalid;
 
     return true;


airween:
Hi @jakubsuchy,

thanks for first time contributing!

lidmodsecurity3 has an awesome regression test framework and many test cases. Could you add some new tests to check your modification to the [existing](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/test/test-cases/regression/operator-verifyssn.json) one? You can add at least a new case with a new format.

After that (I hope) the CI workflow will start (seems like your PR wasn't checked yet).

src/operators/verify_ssn.cc
```diff
@@ -100,7 +100,7 @@ bool VerifySSN::verify(const char *ssnumber, int len) {
         goto invalid;
 
     /* More tests */
-    if (area >= 740 || area == 666)
+    if (area == 666)
```
fzipi, line 103:
```suggestion
    if ((area < 1000 && area >= 900) || area == 666))
```

PR #2975, opened 2023-08-24
Configure: do not check for pcre1 if pcre2 requested

### LINKED ISSUE #2966: PCRE2 support still requires PCRE1
libmodsecurity 2.9.7 compiles fine using --with-pcre2 only when both pcre1 and pcre2 dev packages are installed, which is wrong...
Moreover, the resulting binary links against both libraries... which is further wrong.

ldd mod_security2.so |grep pcre
ldd: warning: you do not have execution permission for `./mod_security2.so'
	libpcre.so.1 => not found
	libpcre2-8.so.0 => /usr/lib64/libpcre2-8.so.0 (0x0000ffffb9194000)
        ...
If only pcre2 dev package is installed, it fails at ./configure --with-pcre2 saying:

In other words, --with-pcre2 still requires pcre1 dev.
This question is the same as the [previous one](https://github.com/SpiderLabs/ModSecurity/issues/2750),however, this issue is not resolved in v2.9.7.

martinhsv:
Closes #2966

### DIFF
--- a/build/find_pcre.m4
+++ b/build/find_pcre.m4
@@ -21,69 +21,73 @@ AC_ARG_WITH(
     [test_paths="${with_pcre}"],
     [test_paths="/usr/local/libpcre /usr/local/pcre /usr/local /opt/libpcre /opt/pcre /opt /usr"])
 
-AC_MSG_CHECKING([for libpcre config script])
-
-dnl # Determine pcre lib directory
-if test -z "${with_pcre}"; then
-    test_paths="/usr/local/pcre /usr/local /usr"
+if test "x${with_pcre2}" != "x" && test "x${with_pcre2}" != "xno"; then
+    AC_MSG_NOTICE([pcre2 specified; omitting check for pcre])
 else
-    test_paths="${with_pcre}"
-fi
+    AC_MSG_CHECKING([for libpcre config script])
 
-for x in ${test_paths}; do
-    dnl # Determine if the script was specified and use it directly
-    if test ! -d "$x" -a -e "$x"; then
-        PCRE_CONFIG=$x
-        pcre_path="no"
-        break
+    dnl # Determine pcre lib directory
+    if test -z "${with_pcre}"; then
+        test_paths="/usr/local/pcre /usr/local /usr"
+    else
+        test_paths="${with_pcre}"
     fi
 
-    dnl # Try known config script names/locations
-    for PCRE_CONFIG in pcre-config; do
-        if test -e "${x}/bin/${PCRE_CONFIG}"; then
-            pcre_path="${x}/bin"
+    for x in ${test_paths}; do
+        dnl # Determine if the script was specified and use it directly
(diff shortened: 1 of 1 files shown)


martinhsv:
The downside with merging this PR in the near term is that if anyone is still using alp2 then that build will fail (I believe it is obsolete and no longer used, but of course, it's difficult to be certain).

marcstern:
Doesn't work for me. I still have pcre included

martinhsv:
Hi @marcstern ,

Hmm. I would think that should have worked. Note that you need to have the modification included before the ./autogen step.

A consequence of this is that it won't work if you start with the tar file (from here: https://github.com/SpiderLabs/ModSecurity/releases/download/v2.9.7/modsecurity-2.9.7.tar.gz ) and then begin with the ```configure``` step. In other words, do a clone from github, apply the patch, and then start with autogen.

If that's already what you did and still see your reported effect, could you post relevant content from your config.log file(any line with 'pcre')? With the patch, I would expect to see something like:
```
...
configure:16344: pcre2 specified; omitting check for pcre
configure:16445: checking for libpcre2 config script
configure:16475: result: /usr/bin/pcre2-config
configure:16507: using pcre2 v10.40
...
```

marcstern:
> ./autogen.sh
libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build'.
libtoolize: copying file 'build/ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'build'.
libtoolize: copying file 'build/libtool.m4'
libtoolize: copying file 'build/ltoptions.m4'
libtoolize: copying file 'build/ltsugar.m4'
libtoolize: copying file 'build/ltversion.m4'
libtoolize: copying file 'build/lt~obsolete.m4'
configure.ac:18: installing 'build/compile'
configure.ac:17: installing 'build/missing'
alp2/Makefile.am: installing 'build/depcomp'

martinhsv:
@marcstern ,

Perhaps I wasn't clear, but you still have to run ```configure``` after ```autogen.sh```.
The ```configure``` step will produce a 'config.log' file. and pcre1 should not get referenced included in the build commands in the Makefile.
I.e.
```
(1) git clone https://github.com/SpiderLabs/ModSecurity
(2) get checkout v2/master
(3) git apply <the location of the patch file for #2975>
(4) ./autogen.sh
(5) ./configure  --enable-pcre-study --enable-pcre-jit --with-pcre2
```

After step (5) the config.log file should exist and include pcre references akin to what I posted here: https://github.com/SpiderLabs/ModSecurity/pull/2975#issuecomment-1693361681

martinhsv:
@marcstern ?

marcstern:
My fault, it's working correctly. Sorry

martinhsv:
Proceeding to merge.

If it turns out that someone is still using the obsolete alp2 parser, see the comment: https://github.com/SpiderLabs/ModSecurity/pull/2975#issuecomment-1692480894

PR #2980, opened 2023-09-02
build: don't directly use the .libs directory

orbea:
The build system should not directly use the `.libs` directory which is intended for internal use by the libtool implementation. Additionally when linking with internal a library the `.la` file should be used as this provides all the information required by the libtool implementation.

When using Slibtool instead of GNU Libtool the build fails because Slibtool is more strict about this and fails to find the required files to link with.

Gentoo Bug: https://bugs.gentoo.org/913484

### DIFF
--- a/examples/multiprocess_c/Makefile.am
+++ b/examples/multiprocess_c/Makefile.am
@@ -6,15 +6,14 @@ multi_SOURCES = \
 	multi.c
 
 multi_LDADD = \
+	$(top_builddir)/src/libmodsecurity.la \
 	$(SSDEEP_LDADD) \
 	$(LUA_LDADD) \
 	$(MAXMIND_LDADD) \
 	$(GLOBAL_LDADD)
 
 multi_LDFLAGS = \
-	-L$(top_builddir)/src/.libs/ \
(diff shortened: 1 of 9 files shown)


orbea:
@martinhsv Can you please review this PR? It should be straight forward.

airween:
Hi @orbea,

thanks for sending for your first contribution!

Unfortunately for some reason, the regression tests didn't run when you sent this PR. Although the PR does not affect relevant parts, could you sent it again or add some minimal modification (fetch the newer version of the source tree), and commit that? Let me see that the CI starts now.

(I can't find even find your PR [here](https://github.com/owasp-modsecurity/ModSecurity/actions?page=4)).

Thanks!

orbea:
@airween I tried to rebase and force push the branch, but it failed on account of code duplication. So tried to make PR https://github.com/owasp-modsecurity/ModSecurity/pull/3097, but it needs maintainer approval.

airween:
@orbea: would you mind to create a new PR from a fresh cloned repository?

I'm going to close this one - thanks for your effort!

PR #2983, opened 2023-09-17
Fix memory leaks in lmdb code (new'd strings)

martinhsv:


martinhsv:
The bug in this case only affected installations using the ```--with-lmdb``` option.

Issue #2973, opened 2023-08-24
Invalid PCRE2 JiT check


marcstern:
In msc_pcre.c, we have the following code at several places:
```
#ifdef WITH_PCRE_STUDY
    #ifdef WITH_PCRE_JIT
    int jit;
    #endif
#endif
[...]
#ifdef WITH_PCRE_STUDY
   #ifdef WITH_PCRE_JIT
      if (msr->txcfg->debuglog_level >= 4) {
         #ifdef WITH_PCRE2
            rc = regex->jit_compile_rc;
         #else
            rc = msc_fullinfo(regex, PCRE_INFO_JIT, &jit);
         #endif
         if ((rc != 0) || (jit != 1)) {
                    *error_msg = apr_psprintf(msr->mp,
                            "Rule %pp [id \"%s\"][file \"%s\"][line \"%d\"] - "
                            "Execution error - "
                            "Does not support JIT (%d)",
                            rule,((rule->actionset != NULL)&&((rule->actionset->id != NULL)&&
                                    (rule->actionset->id != NOT_SET_P))) ? rule->actionset->id : "-",
                            rule->filename != NULL ? rule->filename : "-",
                            rule->line_num,rc);
                    msr_log(msr, 4, "%s.", *error_msg);
                }
            }
   #endif
#endif
```
jit is not initialized when using PCRE2.
Fix: initialize jit to 1 (will potentially be changed to 0 if msc_fullinfo succeeds).
Better fix: this code is copied 7 times in this file; we should put it in a static function.

Question: is it normal that, for PCRE2, we don't check if jit is enabled? Shouldn't we call msc_fullinfo succeeds() as well?

martinhsv:
Hi @marcstern ,

(I guess you actually mean the file re_operators.c ...)

You are correct about the ```int jit``` variable being tested while possibly uninitialized. When I implemented the jit support for pcre2, I made a note to return to this and look into a couple of anomalies in this code (see below).

Have you actually noted any problems with functionality problems related to this?

The odd thing here (and predating all pcre2 support) is that the only difference the ```jit``` int has is the production of that error message content. And I was never able to actually see the error produced in my testing. I believe this is because there is no ```return -1``` which usually triggers the actual message output.

martinhsv:
@marcstern ?

DId you actually notice any functionality problem here?

marcstern:
It's indeed in re_operators.c.
[...]

PR #2987, opened 2023-09-25
'jit' variable not initialized when WITH_PCRE2 is defined

marcstern:


martinhsv:
Hello @marcstern ,

Did you see my inquiry at #2973  ?

marcstern:
Hi @martinhsv,
About functionality problem? Yes, it sometimes creates invalid entries in the debug log.
The current code is clearly buggy, as the value is not initialized, so we have an undefined behaviour.

PR #2993, opened 2023-10-04
Removed redundant processing for "sanitizeXXX" actions

marcstern:
Store "sanitizeMatched" & "sanitizeMatchedBytes" as "sanitiseMatched" & "sanitiseMatchedBytes" to avoid double processing at run-time (only 1 table scan instead of 2).

It also fixes a potential NULL pointer problem (last 3 changes):
```
if (rule->actionset) {
                matched_bytes = apr_table_get(rule->actionset->actions, "sanitizeMatchedBytes") ? 1 : 0;
            }
            if(!matched_bytes)
                matched_bytes = apr_table_get(rule->actionset->actions, "sanitiseMatchedBytes") ? 1 : 0;
```
In case rule->actionset is NULL, last line uses NULL->actions

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -972,7 +972,7 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
     #endif
 
     /* Add rule to the recipe. */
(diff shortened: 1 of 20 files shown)


marcstern:

Issue #2989, opened 2023-09-27
Add support for lua 5.4; remove curl deprecation


3eka:
**Describe the bug**

I have custom (latest) installations of **lua** library and **curl** utility  installed under /pz/SFW directory...

- Lua 5.4.6
- cURL 8.3.0

GCC 12 is used on OpenIndiana (Solaris fork) Illumos based OS

**Logs and dumps**
Lua 5.4 is not supported:
```
:; make
...
libtool: compile:  /usr/gcc/12/bin/gcc -DHAVE_CONFIG_H -I. -DSOLARIS2=11 -D_POSIX_PTHREAD_SEMANTICS -D_REENTRANT -I/usr/include/libxml2 -I/pz/SFW/apr/include -I/pz/SFW/apr/include -I/pz/SFW/include -I/usr/apache2/2.4/include -I/pz/SFW/include -DWITH_CURL_SSLVERSION_TLSv1_2 -DWITH_CURL -I/usr/include/libxml2 -DWITH_LUA -DWITH_PCRE_STUDY -DMODSEC_PCRE_MATCH_LIMIT=1500 -DMODSEC_PCRE_MATCH_LIMIT_RECURSION=1500 -DREQUEST_EARLY -DCACHE_LUA -DPERFORMANCE_MEASUREMENT -DMSC_LARGE_STREAM_INPUT -DWITH_REMOTE_RULES -I/pz/SFW/include -DWITH_PCRE2 -I/pz/SFW/include -DWITH_YAJL -I/usr/include/yajl -lstdc++ -m64 -march=native -std=c89 -std=c99 -D_POSIX_PTHREAD_SEMANTICS -D_XOPEN_SOURCE=600 -D_FILE_OFFSET_BITS=64 -D__EXTENSIONS__ -I/pz/SFW/include -I/pz/SFW/include/openssl -I/pz/SFW/apr/include -I/pz/SFW/include/nghttp3 -I/usr/include -I/usr/gnu/include -I/usr/include/readline -I/usr/include/ncurses -I/usr/include/idn2 -I/pz/SFW/apr/include -I/pz/SFW/include -MT mod_security2_la-msc_lua.lo -MD -MP -MF .deps/mod_security2_la-msc_lua
.Tpo -c msc_lua.c  -fPIC -DPIC -o .libs/mod_security2_la-msc_lua.o                                                                                                                                            
msc_lua.c: In function 'lua_execute':                                                                                                                                                                         
[...]

PR #2996, opened 2023-10-16
Allow lua version 5.4

martinhsv:


martinhsv:
Issue raised via #2989

PR #2999, opened 2023-10-20
Allow to store only the length of ARGS values in ARGS_COMBINED_SIZE

marcstern:
Allow to store only the length of ARGS values in ARGS_COMBINED_SIZE.
The default is to also store length of ARGS names.
This can be a problem, especially with JSON with deep nesting, where ARGS_COMBINED_SIZE may be huge while the total of the ARGS values is low.
This change implements a configuration option to choose between both behaviours.

### DIFF
--- a/apache2/re_variables.c
+++ b/apache2/re_variables.c
@@ -154,7 +154,9 @@ static int var_args_combined_size_generate(modsec_rec *msr, msre_var *var, msre_
     te = (apr_table_entry_t *)arr->elts;
     for (i = 0; i < arr->nelts; i++) {
         msc_arg *arg = (msc_arg *)te[i].val;
+#ifndef ARGS_COMBINED_SIZE_VALUE_ONLY
         combined_size += arg->name_len;
+#endif
         combined_size += arg->value_len;
     }
 
--- a/configure.ac
+++ b/configure.ac
@@ -546,19 +546,19 @@ AC_ARG_ENABLE(server-context-logging,
   log_server_context=''
 ])
 
-# Disable logging of server context
-AC_ARG_ENABLE(server-context-logging,
+# Store only length of ARGS values in ARGS_COMBINED_SIZE
+AC_ARG_ENABLE(args-combined-size-value-only,
               AS_HELP_STRING([--enable-args-combined-size-value-only],
                              [Store only the length of ARGS values in ARGS_COMBINED_SIZE. The default is to also store length of ARGS names.]),
 [
   if test "$enableval" != "no"; then
-    log_server_context=
+    args_size_value=
   else
-    log_server_context="-DARGS_COMBINED_SIZE_VALUE_ONLY"
+    args_size_value="-DARGS_COMBINED_SIZE_VALUE_ONLY"
   fi
 ],
 [
-  log_server_context=''
+  args_size_value=''
 ])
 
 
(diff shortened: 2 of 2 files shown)


martinhsv:
Hi @marcstern ,

Have you tried addressing your underlying goal by using existing rule constructs? Perhaps something like:
```
SecRule ARGS "@gt 0" "id:4002,phase:2,t:length,setvar:'tx.arg_val_len=+%{MATCHED_VAR}'"
```

marcstern:
I'm currently using that, and it introduces a big delay with a lot of ARGS

martinhsv:
This is not a good candidate for a compiler flag.

Even if we did want to include an additional builtin sum like that (and I'm far from clear that that is needed), choosing which sum occurs based on a compiler flag would not be the way to do this.

For example, what is to stop someone six months from now saying that sometimes they want to test ARGS_COMBINED_SIZE including the names (and other times without the names per your interest here) -- and both options need to be available in the same executable?

marcstern:
Your argument is "I refuse it because someone could want more?".
I cannot understand this.

Currently, ARGS_COMBINED_SIZE is totally unusable in many setups; why not allowing someone who doesn't currently use it to use it for their needs?

martinhsv:
@marcstern ,

It is much broader than that.

First consider that there is little positive reason to favour a compile flag as a functionality mechanism for something like this. Compiler flags are an excellent choice if the functionality difference results in different build dependencies -- for example with pcre1 v2. pcre2. For other situations, such a selection is usually more questionable.

Second is consistency. The existing ModSecurity code base includes many different types of controls, and it's important to strive to make new changes fit within the existing framework. I struggle to think of a good set of examples where variant functionality of this type is currently implemented with compiler flags.

On the other hand, there are many examples of variant functionality where something other than compiler flags have been used. For example with a perceived need for something slightly different from t:base64Decode, variant behaviour was not implemented via a compiler flag, but with t:base64DecodeExt. Likewise with t:normalisePath/t:normalisePathWin and t:urlDecode/t:urlDecodeUni. Want to change from the default audit log format? We don't use a compiler flag to select that, but a separate runtime configuration item.

Thirdly is indeed extensibility. If some foreseen 'stage 2' implementation is a likely need/want in the future, an intermediate 'stage 1' implementation **might** be ok if the 'stage 2' implementation builds on 'stage 1'. But that is not the case in this PR. To implement 'stage 2' (having both with-names and without-names sums usable by the same executable) here, would require either scrapping this intermediate implementation and starting over or carrying forward two separate mechanisms, which is an obvious maintenance headache.

marcstern:
Hi Martin,

I'm usually in favor of having a config-type mechanism instead of a compile flag, as it allows the admin to choose instead of the developer. I already proposed several PR with a compile flag because the change (that you call stage 1 if I understand well) is smaller and it was easier to accept in the past.

I'm also usually in favor of having multiple possibilities instead of havingto choose one (think here about ARGS_COMBINED_SIZE  vs. ARGS_VALUES_COMBINED_SIZE). Here, it would, from a functional point of view, an excellent solution obviously. However, from a performance point of view (which is definitely a critical points currently in many environments where we introduce ModSecurity), it would add some treament (instead of removing it with my PR). Because of the current performance problems, I definitely wouldn't favour an extra-functionality in this case. The example of the transformations cannot be compared as, here, the calculation would always be done even if you don't use that feature. Unless you add a run-time flag to disable these calculations (both ARGS_COMBINED_SIZE  & ARGS_VALUES_COMBINED_SIZE), which may be an option.

This PR may potentially be enhanced in the future (last point above), but doesn't have any negative side-effect (as you may not use the new compile flag) and it has two positive effects for the people wanting that functionality.
In these conditions, I don't understand why it should be refused.

PR #3011, opened 2023-11-02
removeByTag isn't executed if no rule id is present in the rule

marcstern:
There's a check for (rule->actionset->id != NULL) that doesn't check for target exceptions in case a rule has no id. This must be a copy-paste from the code checking exceptions by id.
This PR removes that check.
A little enhancement could be to remove the "id=NULL" from the debug log, but this wasn't done here.


marcstern:
Sorry, this was for v2

PR #3012, opened 2023-11-02
ctl:ruleRemoveByTag isn't executed if no rule id is present in the rule

marcstern:
There's a check for (rule->actionset->id != NULL) that doesn't check for target exceptions in case a rule has no id. This must be a copy-paste from the code checking exceptions by id.
This PR removes that check.
A little enhancement could be to remove the "id=NULL" from the debug log, but this wasn't done here.

### DIFF
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -575,7 +575,6 @@ int msre_ruleset_rule_matches_exception(msre_rule *rule, rule_exception *re)   {
                                     &my_error_msg);
                             if (rc >= 0)    {
                                 match = 1;
-                                break;
                             }
                         }
                     }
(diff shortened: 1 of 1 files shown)


martinhsv:
Hello @marcstern ,

Could you please provide a use case where this PR would result in a modification of functionality?

marcstern:
Very basic (illustrative) example:
```
SecRule REQUEST_HEADERS:Content-Type "json" "phase:1,ctl:ruleRemoveByTag=JSON"
[...]
SecRule REQUEST_BODY "^{" "phase:2,tag:JSON,deny,msg:'JSON detected in body'"
```
A correctly typed JSON is (incorrectly) blocked.
If we add an id to the second rule, the exception is taken into account and the request is not blocked.
With the PR, the request is accepted even if the rule has no id.

martinhsv:
If I try to include to include your sample rule:
```
SecRule REQUEST_HEADERS:Content-Type "json" "phase:1,ctl:ruleRemoveByTag=JSON"
```

... then Apache will fail to start and the error message is produced: 'ModSecurity: No action id present within the rule'

Do you experience something different?

marcstern:
It has to be compiled with "./configure --disable-rule-id-validation"

PR #3019, opened 2023-11-22
Log involved collection key in case of DBM write error

marcstern:
This adds the key name in the log.


marcstern:
Ignore this, I created a clean one as a lot of other stuff was merged

PR #3020, opened 2023-11-27
V2/mst/log coll key2

marcstern:
This adds the key name in the log (and its size, as this is the problem most of the time).


marcstern:
Problem with this branch, replaced with https://github.com/SpiderLabs/ModSecurity/pull/3032

PR #3025, opened 2023-12-12
fix missing libxml2 headers

pgnd:
bug fix, https://github.com/SpiderLabs/ModSecurity/issues/3023

### DIFF
--- a/src/modsecurity.cc
+++ b/src/modsecurity.cc
@@ -24,6 +24,7 @@
 #ifdef WITH_LIBXML2
 #include <libxml/xmlschemas.h>
 #include <libxml/xpath.h>
+#include <libxml/parser.h>
 #endif
 #ifdef MSC_WITH_CURL
 #include <curl/curl.h>
--- a/src/request_body_processor/xml.h
+++ b/src/request_body_processor/xml.h
@@ -16,6 +16,7 @@
 #ifdef WITH_LIBXML2
 #include <libxml/xmlschemas.h>
 #include <libxml/xpath.h>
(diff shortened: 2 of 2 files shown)


martinhsv:
Hi @pgnd,

WIth libxml2 v2.12.3 now published, ModSecurity now builds successfully using that version.

It looks to me like this PR is no longer necessary.

Would you like to check and see if you agree?

pgnd:
@martinhsv

available from repos
```
	rpm -qa | grep libxml2
		libxml2-2.12.3-1.fc39.x86_64
		libxml2-devel-2.12.3-1.fc39.x86_64
		python3-libxml2-2.12.3-1.fc39.x86_64
```


sourced for local builds from

	https://copr.fedorainfracloud.org/coprs/pgfed/libxml2/builds/

DISbling application of patch
```
	cat libmodsecurity.spec
		...
		%prep
		...
-		%patch -P 0 -p1
+		# %%patch -P 0 -p1
		...
```

@

	https://copr-dist-git.fedorainfracloud.org/cgit/pgfed/nginx-mainline/libmodsecurity.git/tree/libmodsecurity.spec?id=3b43d1c35857da78bb30c645d9296b873eab21a9#n121

indeed results, now, in an OK build

	https://copr.fedorainfracloud.org/coprs/pgfed/nginx-mainline/build/6774049

~~i'm (still) building a downstream pkg which consumes it; don't _anticipate_ any problems ...~~

EDIT:  downstream pkg using this^^ libmodsecurity build also builds & execs OK.  looks good here.

PR #3032, opened 2024-01-09
Detailed error message when writing collections

marcstern:
This adds the key name in the log (and its size, as this is the problem most of the time).


marcstern:
Corrupted commit, replaced by https://github.com/marcstern/ModSecurity/tree/v2/mst/log_coll_key4

PR #3042, opened 2024-01-28
docs: Fix organization name in references and security e-mail

airween:
* replace "SpiderLabs" with "owasp-modsecurity" in Github links
* changing the e-mail address of security reports


dune73:
Thanks

PR #3052, opened 2024-02-05
In case no rule id exists, log file name & line number in debug log

marcstern:
Fixes several NULL pointers use.
In case no rule id exists, log file name & line number in debug log.
Obsoletes https://github.com/owasp-modsecurity/ModSecurity/pull/2950 (corrupted).

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -30,6 +30,13 @@
     APLOG_USE_MODULE(security2);
 #endif
 
+# Returns the rule id if existing, otherwise the file name & line number
+static const char* id_log(msre_rule* rule) {
+    const char* id = rule->actionset->id;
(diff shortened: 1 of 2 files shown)


apache2/apache2_config.c
```diff
@@ -30,6 +30,13 @@
     APLOG_USE_MODULE(security2);
 #endif
 
+# Returns the rule id if existing, otherwise the file name & line number
```
airween: This is not a `C` comment, it's a preprocessor directive. You should replace it by this:
```C
// Returns the rule id if existing, otherwise the file name & line number
```

airween (changes requested):

apache2/apache2_config.c
marcstern: Done. I wanted to be too smart...

marcstern:

Issue #3053, opened 2024-02-06
libmodsecurity3: SecAction can't be disabled via ctl action


EsadCetiner:
**Describe the bug**

In libmodsecurity3, SecAction can't be disabled via a ctl action like with SecRules. This issue isn't present in ModSecurity2.

**Logs and dumps**

N/A

**To Reproduce**

1. Have a fresh Ubuntu 22.04 server setup, this bug was a bit tricky to reproduce when I was opening this issue.
2. ModSecurity was tested and installed using [digitalwave's ModSecurity repository](https://modsecurity.digitalwave.hu/) and ``apt install nginx-extras libnginx-mod-http-modsecurity``
3. I installed this version of [CRSv4 RC-2](https://github.com/coreruleset/coreruleset/commit/f987e297113545b53d2479a58553e520a3145947) and [ModSecurity.conf](https://github.com/owasp-modsecurity/ModSecurity/commit/622eb9e6c860ade1e2195d316dfbd86ce5710445) with the ``SecRuleEngine`` directive set to on, but it shouldn't matter what rulesets are used.
4. Create this test conf file, but make sure to load it before any blocking rules (i.e ``Include /etc/nginx/modsecurity/coreruleset/rules/*.conf``)
```
SecRule REQUEST_HEADERS:Host "!@streq example.com" "id:1,phase:1,pass,t:none,nolog,ctl:ruleRemoveById=2"

SecAction \
  "id:2,phase:1,pass,t:none,nolog,ctl:ruleRemoveByTag=OWASP_CRS"
```
5. restart nginx then try to send an attack payload, rule 2 should be disabled yet it still disables OWASP_CRS ``curl 127.0.0.1?exec/bin/bash``
6. If you modify the test file to this, then rule 2 is disabled and everything works as expected:
```
SecRule REQUEST_HEADERS:Host "!@streq example.com" "id:1,phase:1,pass,t:none,nolog,ctl:ruleRemoveById=2"

SecRule REQUEST_FILENAME "@unconditionalMatch" \
  "id:2,phase:1,pass,t:none,nolog,ctl:ruleRemoveByTag=OWASP_CRS"
```
**Expected behavior**

SecActions rules should be disableable via a ctl action, just like SecRules.

**Server:**
 - ModSecurity version (and connector): libmodsecurity 3.0.12 and ModSec-Nginx 1.0.3
 - WebServer: Nginx 1.18.0
 - OS: Ubuntu 22.04

[...]

PR #3080, opened 2024-02-11
test: Logical, syntax and cosmetic fixes on test cases

airween:
Tests are a very important part of the library. While investigating an issue, I found some errors in the tests that could have led to false results, such as using a non-existent `@contais` operator instead of `@contains`. This PR fixes these and some other cases.

### DIFF
--- a/test/test-cases/regression/action-ctl_rule_remove_by_id.json
+++ b/test/test-cases/regression/action-ctl_rule_remove_by_id.json
@@ -18,7 +18,7 @@
         "Cookie": "PHPSESSID=rAAAAAAA2t5uvjq435r4q7ib3vtdjq120",
         "Content-Type": "text/xml"
       },
-      "uri":"/wp-login.php?whee&pwd=lhebs",
+      "uri":"/wp-login.php?whee=something&pwd=lhebs",
       "method":"GET",
       "body": [ ]
     },
@@ -28,7 +28,7 @@
     },
     "rules":[
         "SecRule REQUEST_FILENAME \"@endsWith /wp-login.php\" \"id:9002100,phase:2,t:none,nolog,pass,ctl:ruleRemoveById=1\"",
-        "SecRule ARGS \"@contais whe\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS'\""
+        "SecRule ARGS_NAMES \"@contains whee\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS'\""
     ]
   },
   {
@@ -36,7 +36,7 @@
     "version_min":300000,
     "title":"Testing CtlRuleRemoteById (2)",
     "expected":{
-      "debug_log": "Target value: .*Variable: ARGS:pwd"
+      "debug_log": "Target value: .*Variable: ARGS_NAMES:whee"
     },
     "client":{
       "ip":"200.249.12.31",
@@ -50,7 +50,7 @@
         "Cookie": "PHPSESSID=rAAAAAAA2t5uvjq435r4q7ib3vtdjq120",
         "Content-Type": "text/xml"
       },
-      "uri":"/wp-login.php?whee&pwd=lhebs",
+      "uri":"/wp-login.php?whee=something&pwd=lhebs",
       "method":"GET",
       "body": [ ]
     },
@@ -60,7 +60,7 @@
     },
     "rules":[
         "SecRule REQUEST_FILENAME \"@endsWith /wp-login.php\" \"id:9002100,phase:2,t:none,nolog,pass,ctl:ruleRemoveById=123\"",
-        "SecRule ARGS \"@contais whe\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS2'\""
+        "SecRule ARGS_NAMES \"@contains whee\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS2'\""
     ]
   }
 ]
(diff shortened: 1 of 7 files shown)


test/test-cases/regression/action-ctl_rule_remove_by_id.json
```diff
@@ -28,15 +28,15 @@
     },
     "rules":[
         "SecRule REQUEST_FILENAME \"@endsWith /wp-login.php\" \"id:9002100,phase:2,t:none,nolog,pass,ctl:ruleRemoveById=1\"",
-        "SecRule ARGS \"@contais whe\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS'\""
+        "SecRule ARGS_NAMES \"@contains whee\" \"id:1,phase:3,t:none,nolog,pass,tag:'CRS'\""
```
mirkodziadzka-avi, line 31: My first thought "Why did it worked before" ... and then I see this "expected" block wich rely on debug output analysis. Maybe not for this PR, but I think we should at least for all new tests do not use the debug output at all but rely on externally observable behaviour. Like making this rule a

```
SecRule REMOTE_ADDR "@unconditionalMatch" "id:1,phase:2,deny,status=403"
```
and then having 

```
    "expected":{
      "http_code": 200
    },
```

test/test-cases/regression/action-ctl_rule_remove_by_tag.json
```diff
@@ -34,7 +34,7 @@
       "debug_log":"Skipped rule id '2'. Skipped due to a ruleRemoveByTag action."
     },
     "rules":[
-      "SecRule ARGS:key \".\" \"id:4,ctl:ruleRemoveByTag=tag123",
+      "SecRule ARGS:key \".\" \"id:4,ctl:ruleRemoveByTag=tag123\"",
```
mirkodziadzka-avi, line 37: This should not have worked before. It should have been a parsing error. Maybe open a Bug?

test/test-cases/regression/issue-1850.json
```diff
@@ -44,7 +44,7 @@
     "rules": [
       "SecRuleEngine On",
       "SecDefaultAction \"phase:1,status:404,deny\"",
-      "SecRule REQUEST_URI \"@contains /\" \"id:2000001,phase:1,log,redirect:'http://1.1.1.1/failed.html',t:none,msg:\"Unauthorized administrator request'\""
+      "SecRule REQUEST_URI \"@contains /\" \"id:2000001,phase:1,log,redirect:'http://1.1.1.1/failed.html',t:none,msg:'Unauthorized administrator request'\""
```
mirkodziadzka-avi, line 47: Same here. Should have been a parsing error. Maybe open a Bug?

test/test-cases/regression/action-ctl_rule_remove_by_id.json
airween: My first idea was to make a small cleanup with minimal modifications.

Yes, this test works because it checks only the debug log message - which is problematic because there could be a hidden error (see #3053 - this test would have failed there).

But you're right, if we touch a test, should make it more cleaner.

See my new commit [@80e244](https://github.com/owasp-modsecurity/ModSecurity/pull/3080/commits/80e244388438b729e7cba4d8a6e48ac81be2b1d8) which contains this condition too.

test/test-cases/regression/action-ctl_rule_remove_by_tag.json
airween: > This should not have worked before. It should have been a parsing error.

Yes, definitely.

> Maybe open a Bug?

Unfortunately the parser has several issues. I don't know what would be the best step: open a new issue for each discovered (and already known) bug, or collect them into one issue? Or open a new one and collect the others there? (There are still few opened issue related to parser).

Btw I mark this review as resolved, actually it does not affect this PR.

test/test-cases/regression/issue-1850.json
airween: Same as [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3080/files/ed811f106296ab52f8b2ec3bf76b30cfd9e5ac50).

MirkoDziadzka (approved):
Looks good to me.

test/test-cases/regression/action-ctl_rule_remove_by_id.json
MirkoDziadzka: @airween This new condition in expected make only sense if the rule we disable would return another thing, right? For example `deny,status=403`.

Same for the other tests here. 
As I mentioned, not really needed as part of this PR, but adding the test for `200` without changing the rules does not really make sense to me.

test/test-cases/regression/rule-920200.json
```diff
@@ -18,7 +18,7 @@
         "Accept-Language":"en-us,en;q=0.5",
         "Accept":"*/*",
         "Keep-Alive":"300",
-	"Range": "bytes=1-10,11-20,21-30,31-40,41-50,51-60"
+        "Range": "bytes=1-10,11-20,21-30,31-40,41-50,51-60"
```
MirkoDziadzka, line 21: space change?

airween: Yes, an identation fix. Unfortunately it does not seem here in this diff output.

test/test-cases/regression/action-ctl_rule_remove_by_id.json
airween: Ah, right, I reverted my last commit.

MirkoDziadzka (approved):
looks good to me

marcstern (approved):

PR #3090, opened 2024-02-21
Ep/scoped for

devzero2000:
Patches in this series reduce the scope of variables in a for loop. In general, it is a commonly considered useful refactoring


dune73:
Thank you for your contribution @devzero2000. It's been a while. Good to see you are still working on ModSec.

src/parser/seclang-scanner.cc
airween, line 1: The file `seclang-scanner.cc` is a generated file by `flex` - see the [note](https://github.com/owasp-modsecurity/ModSecurity/blob/1146f4593e5e7695ea7e3af3a71e95f30649cabb/src/parser/seclang-scanner.cc#L6) in the header of the file. Every modifications will be lost after a new build process (eg. if you pass the `--enable-parser-generation` option to `./configure`. Please remove that from the PR.

airween (changes requested):

src/parser/seclang-scanner.cc
devzero2000: deleted the seclang-scanner.cc change from this PR

airween: As I see you completely removed the file from the repository.

I asked you only remove it from the PR (from the pull request), not from the repository. `seclang-scanner.cc` is a part of the source tree.

You should have only revert the commit [c121ec5](https://github.com/owasp-modsecurity/ModSecurity/pull/3090/commits/c121ec5314e5d6bc84d116029f139854fdbcb1f5).

devzero2000:
I'll do the pr again

airween:
> I'll do the pr again

Thanks!

PR #3092, opened 2024-02-22
fix: Change 'SecEngineStatus' to Off by default

### LINKED ISSUE #3085: SecStatusEngine should be "Off" in modsecurity.conf-recommended
**Describe the bug**

v2 states that
```
# NB: As of April 2022, there is no longer any advantage to turning this
# setting On, as there is no active receiver for the information.
```
and has the setting turned off for in https://github.com/owasp-modsecurity/ModSecurity/blob/705002be2ba23b01bd9c895a8d01ebd9fd141ceb/modsecurity.conf-recommended#L239

But v3 has it enabled in https://github.com/owasp-modsecurity/ModSecurity/blob/4e4f3291ad2f65df7b945d3fdc4ef67d6fcc631d/modsecurity.conf-recommended#L284.

**Expected behavior**

Both should be in sync, and turned off.

airween:
The default setting of `SecStatusEngine` is `Off` because there is no any advantage to turn that on.

This PR changes the value to `Off`.

Closes #3085.

### DIFF
--- a/modsecurity.conf-recommended
+++ b/modsecurity.conf-recommended
@@ -281,5 +281,7 @@ SecUnicodeMapFile unicode.mapping 20127
 # The following information will be shared: ModSecurity version,
 # Web Server version, APR version, PCRE version, Lua version, Libxml2
 # version, Anonymous unique id for host.
+# NB: As of April 2022, there is no longer any advantage to turning this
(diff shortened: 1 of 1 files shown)


dune73 (approved):

dune73:
Thank you. Merging now.

PR #3096, opened 2024-02-26
Clean up 'return' never will be executed.

gberkes:
Issue reference: https://sonarcloud.io/project/issues?open=AY1CfKFNrsSpWCKX0w9c&id=owasp-modsecurity_ModSecurity

There are three occurrences of this issue. I deleted pthread_exit(NULL) calls, because these execution points are not in
pthread execution scope. In process_request() I kept pthread_exit(NULL) and deleted return(NULL), because here we are
in pthread execution context. 
(FYI: 11.5 Thread Termination in [Advanced Programming in the UNIX® Environment, Third Edition](https://learning.oreilly.com/library/view/advanced-programming-in/9780321638014/))

### DIFF
--- a/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
+++ b/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
@@ -102,7 +102,7 @@ struct data_ms {
         delete modsecTransaction;
     }
 
-    pthread_exit(NULL);
+    pthread_exit(nullptr);
 }
 
 
--- a/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
+++ b/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
@@ -103,7 +103,6 @@ static void *process_request(void *data) {
     }
 
     pthread_exit(NULL);
-    return NULL;
 }
 
 
@@ -165,7 +164,6 @@ class ReadingLogsViaRuleMessage {
 
         delete rules;
         delete modsec;
-        pthread_exit(NULL);
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
```diff
@@ -103,7 +103,6 @@ static void *process_request(void *data) {
     }
 
     pthread_exit(NULL);
-    return NULL;
```
MirkoDziadzka, line 106: I agree with this change. The function is declared as void, so returning NULL is wrong.

But maybe we should also mark the whole function as `[[noreturn]]`, as `pthread_exit()` will not return.

MirkoDziadzka (approved):
looks good to me, one comment for possible improvements.

examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
gberkes: Thank you Mirko! I marked the functon as [[noreturn]].

examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
```diff
@@ -103,7 +103,6 @@ static void *process_request(void *data) {
     }
 
     pthread_exit(NULL);
```
airween: To avoid [Sonarcloud's](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&open=AY1CfKFIrsSpWCKX0w9K) next warning, you should replace `NULL` literal here by `nullptr`.

airween (changes requested):

examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
gberkes: Thank you. Done.

airween (approved):

PR #3097, opened 2024-02-27
build: don't directly use the .libs directory

orbea:
This is a duplicate of PR https://github.com/owasp-modsecurity/ModSecurity/pull/2980 to try to trigger the CI regression tests.

The build system should not directly use the `.libs` directory which is intended for internal use by the libtool implementation. Additionally when linking with internal a library the `.la` file should be used as this provides all the information required by the libtool implementation.

When using Slibtool instead of GNU Libtool the build fails because Slibtool is more strict about this and fails to find the required files to link with.

Gentoo Bug: https://bugs.gentoo.org/913484

### DIFF
--- a/test/Makefile.am
+++ b/test/Makefile.am
@@ -42,6 +42,7 @@ noinst_HEADERS = \
 
 
 unit_tests_LDADD = \
+	$(top_builddir)/src/libmodsecurity.la \
 	$(CURL_LDADD) \
 	$(GEOIP_LDADD) \
 	$(MAXMIND_LDADD) \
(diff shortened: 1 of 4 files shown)


orbea:
Trying to split the changes in half didn't seem to make a difference for Quality Gate....

airween:
> Trying to split the changes in half didn't seem to make a difference for Quality Gate....

Yeah, QG check is very weird - I'm sure that's a false positive hit (the code is definitely not NEW).

I have to find a solution - sorry, and thanks for your patience.

orbea:
> I have to find a solution - sorry, and thanks for your patience.

No worries and thanks for your time!

airween:
It sounds like a stupid question, but could you make a fresh clone and apply your modification in a separated new commit?

[Here](https://github.com/owasp-modsecurity/ModSecurity/pull/2980) I see this:

modsec_duplicate

and may be Sonarcloud thinks that you added the mentioned lines through the pull. (Because 4d7d1fa3 was made in Sept, 2023)

hydrapolic:
ping @orbea

airween:
May be we will copy this PR and send it again.

Meanwhile the CI workflow had changed a lot (eg. see the Windows support), so we insist to run the whole test.

PR #3098, opened 2024-02-29
Ep/scoped for: second pr

devzero2000:
this is the second request to do a merge for this topic of reducing the scope of variables in a for loop. Compared to the previous PR, the modification to the seclang-parser.cc file has been eliminated, as this is a file generated by lex.


airween (approved):
LGTM. Thanks.

PR #3104, opened 2024-03-06
V3/remove this throw call transaction h

gberkes:
Introducing the use of assertions to address throw; calls that lack try-catch blocks. Upon examining the caller code that utilized methods containing the questioned throw; calls, it became clear that, in the current state of development, there are no scenarios where execution could reach these throw; calls. However, we cannot guarantee this for future development. For instance, if someone attempts to use getCurrentMarker() without first verifying isInsideAMarker(), ModSecurity would encounter the throw; and terminate. The issue with the other throw; call is similar in that it is, fortunately, unreachable at the moment. However, it differs because this throw; is intended to handle a case that has not yet been developed.

Fortunately, I found an article titled "Effective Use of Assertions in C++" by Mike A. Martin in (ACM SIGPLAN Language Tips, page 3), which offers a neat way to handle such cases, specifically regarding argument validation and unreachable code. [Link to the article](https://dl.acm.org/doi/pdf/10.1145/240964.240969).

Following the guidance from this article, I addressed the issues and also included modifications to enrich assert error messages. Furthermore, I updated configure.ac to maintain the usual build procedure and modified README.md to introduce the new configure flag.

### DIFF
--- a/README.md
+++ b/README.md
@@ -213,6 +213,12 @@ $ cd test
 $ ./regression-tests
 $ ./unit-tests
  ```
+Please take care that the '<path/to/modsecurity>/test/cppcheck_suppressions.txt'
+file contains hardcoded 'filename:line number' suppressions. If you modify a
+file with explicit suppressions, you must update the 'cppcheck_suppressions.txt'
+accordingly to ensure the suppression references remain aligned with the
+intended lines of code. Failing to do so might trigger 'make check-static' to
+break the build with errors that are logically unrelated to your modifications.
 
 ### Debugging
 
@@ -235,7 +241,12 @@ $ ./configure --enable-assertions=yes
 $ make
 $ sudo make install
 ```
+"Assertions allow us to document assumptions and to spot violations early in the
+development process. What is more, assertions allow us to spot violations with a
+minimum of effort." https://dl.acm.org/doi/pdf/10.1145/240964.240969
 
+It is recommended to use assertions where applicable, and to enable them with
+'--enable-assertions=yes' during the testing and debugging workflow.
 
 ## Reporting Issues
 
--- a/README.md
+++ b/README.md
@@ -231,7 +231,7 @@ CFLAGS to disable the compilation optimization parameters:
 ```shell
 $ export CFLAGS="-g -O0"
 $ ./build.sh
-$ ./configure
+$ ./configure --enable-assertions=yes
 $ make
 $ sudo make install
 ```
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -321,7 +321,7 @@ class TransactionSecMarkerManagement {
     }
 
  private:
-    std::shared_ptr<std::string> m_marker = nullptr;
+    std::shared_ptr<std::string> m_marker;
 };
 
 /** @ingroup ModSecurity_CPP_API */
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -308,7 +308,7 @@ class TransactionSecMarkerManagement {
     }
 
     std::shared_ptr<std::string> getCurrentMarker() const {
-        assert(m_marker != nullptr);
+      assert((m_marker != nullptr) && "You might have forgotten to call and evaluate isInsideAMarker() before calling getCurrentMarker().");
         return m_marker;
     }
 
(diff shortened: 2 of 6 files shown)


headers/modsecurity/transaction.h
```diff
@@ -307,11 +308,8 @@ class TransactionSecMarkerManagement {
     }
 
     std::shared_ptr<std::string> getCurrentMarker() const {
-        if (m_marker) {
-            return m_marker;
-        } else {
-            throw;
```
airween, line 313: SonarCloud report is [here](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfJimrsSpWCKX0vrq)

src/rule_with_actions.cc
```diff
@@ -121,10 +122,8 @@ RuleWithActions::RuleWithActions(
                     m_actionsRuntimePos.push_back(a);
                 }
             } else {
+                assert(false && "The handling of RunTimeBeforeMatchAttemptKind has not been implemented yet.");
                 delete a;
-                std::cout << "General failure, action: " << a->m_name;
-                std::cout << " has an unknown type." << std::endl;
-                throw;
```
airween, line 127: SonarCloud report is [here](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfKAnrsSpWCKX0waI)

airween:
Just for the record: this PR fixes two SonarCloud issues in files:

* [headers/modsecurity/transaction.h](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfJimrsSpWCKX0vrq)
* [src/rule_with_actions.cc](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfKAnrsSpWCKX0waI)

I also added these references as conversation.

README.md
```diff
@@ -231,7 +231,7 @@ CFLAGS to disable the compilation optimization parameters:
 ```shell
 $ export CFLAGS="-g -O0"
 $ ./build.sh
-$ ./configure
+$ ./configure --enable-assertions=yes
```
airween, line 240: Please explain the behavior of this new option. First it's enough to place that in a single comment.

Later we should consider to create a separated documentation / extend the existing one with all options.

airween (changes requested):

headers/modsecurity/transaction.h
```diff
@@ -323,7 +321,7 @@ class TransactionSecMarkerManagement {
     }
 
  private:
-    std::shared_ptr<std::string> m_marker;
+    std::shared_ptr<std::string> m_marker = nullptr;
```
MirkoDziadzka: I think C++ gurantees this. `std::shared_ptr` is a class and therefore has a constructor and initialize itself.

src/rule_with_actions.cc
MirkoDziadzka: What we are doing here is a switch over `a->action_kind` withe the 2 cases `Action::ConfigurationKind` and `Action::RunTimeOnlyIfMatchKind`.

This `else` would be the default case of the switch, right.

Would it be better to replace the if/else with a switch and ask the compiler of our switch case is complete (e.g. all cases are implemented)

----

independent from this remark: Is the else case a "can never happen" or is it a "user configured invalid modsec code". In the second case, we change the semantic which could be a problem.

README.md
gberkes: Done: https://github.com/owasp-modsecurity/ModSecurity/pull/3104/commits/2c8684eeb0e99b25a4d488d802910cba927153bb

headers/modsecurity/transaction.h
gberkes: Yes. You're right. Fixed: https://github.com/owasp-modsecurity/ModSecurity/pull/3104/commits/cee3965f689ea52004d0e8b160054e227e6c5593

src/rule_with_actions.cc
gberkes: The refactoring of the if-then chain to a switch statement is complete. We reviewed the code together with @airween  and couldn't find a way to force execution into the else/default case.

airween:
Please pull the modifications from #3134, and apply this PR. Or you can resolve the conflicts [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3104/conflicts).

huberts90:
@gberkes
Could you please address requested change to complete this issue?

PR #3114, opened 2024-03-27
fix: Sonarcloud memleak fixes

airween:
This pull request fixes two memory leaks, reported by [SonarCloud](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&resolved=false&id=owasp-modsecurity_ModSecurity).

* [test/regression/regression.cc](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&resolved=false&id=owasp-modsecurity_ModSecurity&open=AY1CfKEhrsSpWCKX0w7f)
* [tools/rules-check/rules-check.cc](https://sonarcloud.io/project/issues?impactSoftwareQualities=RELIABILITY&resolved=false&id=owasp-modsecurity_ModSecurity&open=AY1CfKGQrsSpWCKX0xBV)

None of them modified the library itself.


test/regression/regression.cc
```diff
@@ -476,7 +476,9 @@ int main(int argc, char **argv) {
     std::string ver(MODSECURITY_VERSION);
     std::string envvar("MODSECURITY=ModSecurity " + ver + " regression tests");
 
-    putenv(strdup(envvar.c_str()));
+    char *envvarptr = strdup(envvar.c_str());
+
+    putenv(envvarptr);
```
mirkodziadzka-avi: Why do we need the strdup at all? The lifetime of envvar is main and it does not change, so we should be able to use envvar.c_str() directly

tools/rules-check/rules-check.cc
```diff
@@ -41,6 +41,7 @@ int main(int argc, char **argv) {
 
     if (*args == NULL) {
         print_help(argv[0]);
+        delete rules;
```
mirkodziadzka-avi: How about making rules a std::unique_ptr instead which takes care of the lifetime

tools/rules-check/rules-check.cc
```diff
@@ -50,6 +51,7 @@ int main(int argc, char **argv) {
         const char *arg = *args;
         std::string err;
         int r;
+        bool need_free = false;
```
mirkodziadzka-avi: How about making args a std::string?

test/regression/regression.cc
airween: > Why do we need the strdup at all?

I guess because `putenv()` might modify the string.

> The lifetime of envvar is main and it does not change, so we should be able to use envvar.c_str() directly

`envvar.c_str()` returns a `const char *` but `putenv()` does not allow it, only `char *`.

Never mind, see my newer commit [4085ff5](https://github.com/owasp-modsecurity/ModSecurity/pull/3114/commits/4085ff553624121d6cd76d20064394e4d6d21c1a), where I replaced `putenv()` by `setenv()`. Now SonarCloud has stopped complaining.

tools/rules-check/rules-check.cc
airween: Yep, seems that works (see next commit).

tools/rules-check/rules-check.cc
airween: Do you mean `arg` (and not `args`)? Because it makes sense to me, but after a quick review I don't understand the idea.

airween:
[This](https://github.com/owasp-modsecurity/ModSecurity/blob/9ea5b475b2e94c8b64eb3698e7da28c6bcec0921/tools/rules-check/rules-check.cc#L49-L62) block makes no sense.

The variable is initialized as an empty string in line [49](https://github.com/owasp-modsecurity/ModSecurity/blob/9ea5b475b2e94c8b64eb3698e7da28c6bcec0921/tools/rules-check/rules-check.cc#L49) in each cycle.

```C++
49.      std::string argFull("");
```
therefore the condition's left side (`argFull.empty()`) is always `true` which means the condition itself
```C++
54.        if (argFull.empty() == false) {
```
is always `false`. Moreover, in the two branches in that block after the `append()` the code jumps to the `next` label, where the cycle ends...

So I'm going to remove it.

tools/rules-check/rules-check.cc
airween: Okay I'm done. See the result after my last commit [cdb6b839](https://github.com/owasp-modsecurity/ModSecurity/pull/3114/commits/cdb6b839e1de54ce126a2cc8e97a9f2c2846b5a9)

airween: Please ignore the new SonarCloud [issue](https://sonarcloud.io/project/issues?resolved=false&pullRequest=3114&id=owasp-modsecurity_ModSecurity&open=AY6AlC0CBkWqhBwO48MX) - that's not because of this PR.

MirkoDziadzka (approved):
Looks a lot simpler, thanks.

tools/rules-check/rules-check.cc
```diff
@@ -46,41 +45,26 @@ int main(int argc, char **argv) {
...
-
-        if (arg[0] == '\"' && argFull.empty() == true) {
-            if (arg[strlen(arg)-1] == '\"') {
-                argFull.append(arg+1, strlen(arg) - 2);
-            } else {
-                argFull.append(arg+1);
-                goto next;
-            }
-        }
+        // strip arg from leading and trailing '"' chars
```
marcstern, line 56: We check that arg (*args) is not NULL.
Shouldn't we check that arg is not empty before erasing (cfr. +1 on next line)?
If it's empty, I imagine that find_last_not_of('\"') points to arg[0], so we'll erase from arg[1], which may be invalid memory.

marcstern (changes requested):

tools/rules-check/rules-check.cc
airween: > Shouldn't we check that arg is not empty before erasing (cfr. +1 on next line)?

The code does not erase the arg. It just removes the `"` characters - if there is one - from the end, and then from the begin.

That compositions (in those two lines) we can consider as a "strip()" function, which removes `"` characters from the front and from the end.

> If it's empty, I imagine that find_last_not_of('"') points to arg[0],

no, if the string is empty then `find_last_not_of('"')` returns [std::npos](https://cplusplus.com/reference/string/string/npos/) (see [find_last_not_of()](https://cplusplus.com/reference/string/string/find_last_not_of/) documentation).

Technically it means if it's empty it removes nothing.

I think the mentioned code is good as is.

marcstern (approved):

PR #3117, opened 2024-03-29
fix: Changed 'equal_range()' + loop by 'find()' in resolveFirst() methods

airween:
This PR fixes SonarCloud issues in these files:

* [src/anchored_set_variable.cc](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfJ_hrsSpWCKX0wV6)
* [src/collection/backend/in_memory-per_process.cc](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfJ-3rsSpWCKX0wUJ) and one [more](https://sonarcloud.io/project/issues?resolved=false&types=BUG&id=owasp-modsecurity_ModSecurity&open=AY1CfJ-3rsSpWCKX0wUM) place

Summary: [AnchoredSetVariable](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/headers/modsecurity/anchored_set_variable.h#L71) and [InMemoryCollection](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/collection/backend/in_memory-per_process.h#L72-L73) types are derived from [std::unordered_multimap](https://en.cppreference.com/w/cpp/container/unordered_multimap).

Both type has a `resolveFirst(key)` method ([AnchoredSetVariable](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/anchored_set_variable.cc#L110-L119), [InMemoryCollection](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/collection/backend/in_memory-per_process.cc#L226-L240)), which finds the first occurrence of key in the collection.

The current implemented version uses `equal_range()` ([doc](https://en.cppreference.com/w/cpp/container/unordered_multimap/equal_range)) to find the key in a loop, and breaks from that after first match. This C++ collection has another method to find it where we don't need any loop, that's the `find()` ([doc](https://en.cppreference.com/w/cpp/container/unordered_multimap/find)).

### DIFF
--- a/src/anchored_set_variable.cc
+++ b/src/anchored_set_variable.cc
@@ -111,9 +111,7 @@ std::unique_ptr<std::string> AnchoredSetVariable::resolveFirst(
     const std::string &key) {
 
     if (auto search = this->find(key); search != this->end()) {
-        auto b = std::make_unique<std::string>();
-        b->assign(search->second->getValue());
-        return b;
+        return std::make_unique<std::string>(search->second->getValue());
     }
 
     return nullptr;
--- a/src/anchored_set_variable.cc
+++ b/src/anchored_set_variable.cc
@@ -109,12 +109,13 @@ void AnchoredSetVariable::resolve(const std::string &key,
 
 std::unique_ptr<std::string> AnchoredSetVariable::resolveFirst(
     const std::string &key) {
-    auto range = equal_range(key);
-    for (auto it = range.first; it != range.second; ++it) {
-        std::unique_ptr<std::string> b(new std::string());
-        b->assign(it->second->getValue());
+
+    if (auto search = this->find(key); search != this->end()) {
+        auto b = std::make_unique<std::string>();
+        b->assign(search->second->getValue());
         return b;
     }
+
     return nullptr;
 }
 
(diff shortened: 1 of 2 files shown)


src/anchored_set_variable.cc
```diff
@@ -109,12 +109,13 @@ void AnchoredSetVariable::resolve(const std::string &key,
...
     const std::string &key) {
-    auto range = equal_range(key);
-    for (auto it = range.first; it != range.second; ++it) {
-        std::unique_ptr<std::string> b(new std::string());
-        b->assign(it->second->getValue());
+
+    if (auto search = this->find(key); search != this->end()) {
+        auto b = std::make_unique<std::string>();
+        b->assign(search->second->getValue());
         return b;
```
MirkoDziadzka: Not sure, but if you improve this, I think this can be optimized by constructing this inline and avoid the resizing between line 114 and line 115

Probably like (not tested)
```cpp
return std::make_unique<std::string>(search->second->getValue());
```

airween: Ah, you are right, this make sense. And yes, looks like it works. Changed in [16d0df0](https://github.com/owasp-modsecurity/ModSecurity/pull/3117/commits/16d0df0ff9598cedec29330944e09151dba8fea2), thanks!

fzipi (approved):
LGTM. I was looking if we can move the lock inside the if (before setting the value/expiration), but I guess the element found could be removed between the find step and the actual modification.

fzipi (approved):

marcstern (approved):

PR #3120, opened 2024-04-04
Check for null pointer dereference (almost) everywhere

marcstern:
Added a design doc explaining the approach


configure.ac
```diff
@@ -304,7 +304,20 @@ if test "$build_docs" -eq 1; then
     AC_CONFIG_FILES([doc/Makefile])
 fi
 
-
+AC_ARG_ENABLE(assertions,
+              AS_HELP_STRING([--enable-assertions],
+                             [Turn on assertions checks (undefine NDEBUG])),
+[
+  if test "${enableval}" = "yes"; then
+      assertions=-UNDEBUG
```
airween: Perhaps the right form is `-DNDEBUG` instead of  `-UNDEBUG`, isn't it?

configure.ac
```diff
@@ -304,7 +304,20 @@ if test "$build_docs" -eq 1; then
     AC_CONFIG_FILES([doc/Makefile])
 fi
 
-
+AC_ARG_ENABLE(assertions,
+              AS_HELP_STRING([--enable-assertions],
```
airween, line 311: You should add this flag to GH CI workflow globally.

airween (changes requested):

apache2/apache2_config.c
```diff
@@ -30,6 +30,14 @@
     APLOG_USE_MODULE(security2);
 #endif
 
+// Returns the rule id if existing, otherwise the file name & line number
+static const char* id_log(msre_rule* rule) {
+	char* id = rule->actionset->id;
```
airween: This should be `const`:
```C
        const char* id = rule->actionset->id;
```

configure.ac
airween: It would by highly recommended to add within this PR, because that will check all modifications.

airween (changes requested):

apache2/apache2_config.c
fzipi: What if `rule` is NULL? Or if `rule->actionset` is NULL?

apache2/apache2_config.c
```diff
@@ -30,6 +30,14 @@
     APLOG_USE_MODULE(security2);
 #endif
 
+// Returns the rule id if existing, otherwise the file name & line number
+static const char* id_log(msre_rule* rule) {
```
fzipi: Maybe `log_id` instead? 🤔

apache2/apache2_io.c
```diff
@@ -179,12 +179,13 @@ apr_status_t input_filter(ap_filter_t *f, apr_bucket_brigade *bb_out,
  * Reads request body from a client.
  */
 apr_status_t read_request_body(modsec_rec *msr, char **error_msg) {
+    assert(msr != NULL);
+    assert( error_msg!= NULL);
```
fzipi:
```suggestion
    assert(error_msg!= NULL);
```

fzipi (commented):
The use of assertions will _terminate the process_, right? Is that what you want?

airween:
> The use of assertions will _terminate the process_, right? Is that what you want?

You can control the behavior with a macro. See the identical solution in case of [libmodsecurity3](https://github.com/owasp-modsecurity/ModSecurity/pull/3104#issue-2171232016). Here @gberkes [introduced](https://github.com/owasp-modsecurity/ModSecurity/pull/3104/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R373) a new `CFLAG`, which can be controlled with a configure option. It's disabled by default, so the process won't terminate - but we can use that in our test environment.

configure.ac
```diff
@@ -305,6 +305,23 @@ if test "$build_docs" -eq 1; then
...
 
+# Add assert() usage
+
+AC_ARG_ENABLE(assertions,
+              AS_HELP_STRING([--enable-assertions],
+                             [Turn on assertions checks (undefine NDEBUG)]),
+[
+  if test "${enableval}" = "yes"; then
+      assertions='-UNDEBUG'
+      MODSEC_EXTRA_CFLAGS="$MODSEC_EXTRA_CFLAGS $assertions"
```
airween: Why do you extend this CFLAG here? Wouldn't be a better place in line [849](https://github.com/owasp-modsecurity/ModSecurity/pull/3120/files#diff-49473dca262eeab3b4a43002adb08b4db31020d190caaad1594b47f1d5daa810R849)?

configure.ac
```diff
@@ -827,6 +844,7 @@ else
     EXTRA_CFLAGS="-O2 -g -Wall $strict_compile"
   fi
 fi
+EXTRA_CFLAGS="$EXTRA_CFLAGS $assertions"
```
airween: I'm afraid this has no effect. I mean I run `./configure ... --enable-assertions`, but I don't see the `DEBUG` flag at all in generated files (`Makefile`):
```
$ grep -rI MODSEC_EXTRA_CFLAGS * | grep DEBUG
[EMPTY]
```

airween (changes requested):

configure.ac
marcstern: Sure. How to do that? I never had a look at this

configure.ac
marcstern: NDEBUG toggles the behavior of assert():
if defined, assert() does nothing
if not defined, assert() performs the check

configure.ac
marcstern: --disable-assertions => NDEBUG
--enable-assertions => nothing special

apache2/apache2_config.c
marcstern: It was static because it was duplicated in 2 files. I merged them

marcstern:
> The use of assertions will _terminate the process_, right? Is that what you want?

Yes, as it's an impossible path, unless the implementation is bogus

apache2/apache2_config.c
marcstern: They're never NULL, but you're right: I added an assert().

configure.ac
airween: Seems like it's already done in [dd400f7f](https://github.com/owasp-modsecurity/ModSecurity/pull/3120/commits/dd400f7fa31da8174a930806e9c46bc99062db06#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL35).

airween (approved):

airween:
Looks good to me. I think the SonarCloud reports are false positives, all mentioned duplicate code had added years ago. Merging now - thanks, Marc.

Issue #3111, opened 2024-03-25
Annoying DNS queries with @rbl operator


ne20002:
**Describe the bug**

Using a pihole in my network I realized that the rbl queries from Modsecurity are doing additional dns queries with the search domain (from resolv.conf) added.
This happens only for @rbl queries (or maybe modsecurity in general). It does not happen from other containers in the pod nor any other system in my network.

![grafik](https://github.com/owasp-modsecurity/ModSecurity/assets/39007279/debfab12-ce49-42be-85c5-6adf88fbed15)

I use the owasp/modsecurity-crs:nginx container from Docker hub.

**Expected behavior**

The additional queries are not done.

dune73:
That would be `0x42.ch` in your case?

ne20002:
Yes

dune73:
Do you happen to know if the apache / ModSec2 docker container has the same problem?

ne20002:
No, sorry. I only use the Nginx version.

airween:
@ne20002 thanks for the report.

Could you provide an example, how can I try the explained behavior?

> It does not happen from other containers in the pod nor any other system in my network.

This means the "vanilla" ModSecurity instance does not do this?

ne20002:
> This means the "vanilla" ModSecurity instance does not do this?

No, this just means that no other server in my network creates dns requests with the seen postfix. It's just the queries created by the @rbl checks.

This is the part in my config:

```
# xbl.spamhaus.org to block malicious/infected ips
SecRule IP:PREVIOUS_RBL_CHECK "@eq 1" "phase:1,id:910201,t:none,pass,nolog,skipAfter:END_RBL_LOOKUP"
  
SecRule REMOTE_ADDR "@rbl xbl.spamhaus.org" \
   "phase:1,id:910202,\
    t:none,pass,nolog,auditlog,\
    msg:'RBL Match for SPAM Source',\
    tag:'AUTOMATION/MALICIOUS',\
    severity:'CRITICAL',\
    setvar:'tx.msg=%{rule.msg}',\
    setvar:'tx.lfi_score=+%{tx.critical_anomaly_score}',\
    setvar:'tx.inbound_anomaly_score_pl1=+%{tx.critical_anomaly_score}',\
    setvar:ip.spammer=1,\
    expirevar:ip.spammer=86400,\
    setvar:ip.previous_rbl_check=1,\
    expirevar:ip.previous_rbl_check=86400,\
[...]

PR #3127, opened 2024-04-22
fix(rbl): typo in rbl check selector

fzipi:
Fixes rbl provider always using httpbl

### DIFF
--- a/src/operators/rbl.h
+++ b/src/operators/rbl.h
@@ -71,9 +71,9 @@ class Rbl : public Operator {
                 m_demandsPassword = true;
                 m_provider = RblProvider::httpbl;
             } else if (m_service.find("uribl.com") != std::string::npos) {
-                m_provider = RblProvider::httpbl;
+                m_provider = RblProvider::uribl;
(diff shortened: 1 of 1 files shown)


airween:
Hi @fzipi, thanks for this patch!

You are definitely right, that's a typo.

And it's relevant [here](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/operators/rbl.cc#L226), where the mentioned argument is used, and [this](https://github.com/owasp-modsecurity/ModSecurity/blame/v3/master/src/operators/rbl.cc#L182-L200) is where the decision is made. It's clearly visible that the method runs the same block, no matter what is the operator argument.

I wasn't able to check the issue #3111 so I don't know whether this solves or not, but I'm going to merge this PR now.

Thanks again!

airween (approved):

fzipi:
I think this only solves getting additional information from the provider in https://github.com/owasp-modsecurity/ModSecurity/blob/6217b4ec1eb7bce6d48716842d904caafc99c8ec/src/operators/rbl.cc#L182 method. The DNS query is not changed.

Issue #321, opened 2013-10-17
rules-updater.pl config file trailing whitespace sensitivity


rcbarnett-zz:
MODSEC-167: The rules-updater.pl is sensitive to trailing spaces in its config file. An example is if the repositoryURI value has a trailing space, that space is carried over to the get request. Making the trailing .\* non-greedy resolves this.

--- rules-updater.pl    2010-06-04 11:37:44.000000000 -0700
+++ rules-updater.pl.new    2010-08-05 13:03:00.500825094 -0700
@@ -177,7 +177,7 @@

```
    # Parse
    chomp;
```
-       my($var,$q1,$val,$q2) = (m/^\s_(\S+)\s+(['"]?)(._)(\2)\s*$/);
- ```
    my($var,$q1,$val,$q2) = (m/^\s*(\S+)\s+(['"]?)(.*?)(\2)\s*$/);
  
  # Fixup values
  $var = lc($var);
  ```

PR #3132, opened 2024-04-25
Add support to build libModSecurity v3 on Windows

eduar-hte:
This PR includes changes to support building libModSecurity v3 on Windows.

Work was initially done on a snapshot of v3.0.12 to work on the latest stable version of the library, and then changes were applied to v3/master up to the latest commit at the time the PR was created (commit 07e5a70).

The library builds with no warnings and was tested as a 64-bit DLL on Windows, running all unit & regression tests to validate the build. Additionally, the updated code was built on Linux, and unit & regression tests were executed on that version too. See below for more information on the results.

Building the library on Windows is done using MSVC C++ compiler, CMake and Conan package manager, so a new CMake build configuration was introduced for this platform. The following libModSecurity optional dependencies were integrated through available Conan package managers:

 * libxml2 2.12.6
 * libcurl 8.6.0
 * libmaxminddb 1.9.1
 * LUA 5.4.6
 * lmdb 0.9.31

These optional features can be enabled/disabled through flag variables in the CMake configuration file. The optional third-party dependencies GeoIP and ssdeep are not included in the port due to lack of Conan packages.

Additionally, a Windows Docker container configuration is included to simplify prerequisites setup and build of the library and associated binaries.

More information on building the library on Windows is available in build/win32/README.md.

# Miscellaneous

 * In libModSecurity configuration files, paths on Windows still need to be specified using (forward) slashes. This avoids having to introduce changes to the parser.
   * For example, the path `C:\TEMP\modsec.log` needs to be configured as `c:/temp/modsec.log` (casing can be different because Windows filename are case-insensitive).
 * `src/config.h` is generated with CMake using similar rules as `configure` for non-Windows builds. This also sets up `HAVE_xxx` defines (such as `HAVE_LUA`), though these seem not to be used in the codebase (similar `WITH_xxx` compiler definitions are used instead, though these are not set in `src/config.h`). 

# Summary of changes

The following is a summary of the changes (more detail is available in each commit):

 * `SharedFiles` required updating file locking support to work on Windows. Additionally, a larger update was done on this class to remove unused shared memory code. See commit for detailed analysis of the rationale and changes.
 * `SetENV::evaluate` updated to use `_putenv_s` instead of `setenv`, not available in the MSVC C++ compiler.
 * `MultipartPartTmpFile::Open` updated usage of `mkstemp` (not available in the MSVC C++ compiler) with `_mktemp_s` and `_open`.
 * `HttpsClient::download` updated to set CURL option to use the native CA store for certificate validation on Windows.
 * `expandEnv`, `createDir` & `cpu_seconds` (from `src/utils/system.cc`) updated to work on Windows. `expandEnv` on Windows uses Glob from the POCO C++ libraries.
 * `Env::evaluate` updated to handle the case that on Windows environment variables are case-insensitive.
 * Fixed a *use after free* bug in `ModSecurity::processContentOffset`.

NOTE: The list does not include changes to files to update included header files to compile with the MSVC++ compiler on Windows and other minor changes.

# Unit & regression test results

All changes to the latest version of the library on v3/master were built on Ubuntu 18.04 LTS and Ubuntu 22.04 LTS following the build instructions in [compilation recipes](https://github.com/owasp-modsecurity/ModSecurity/wiki/Compilation-recipes-for-v3.x), with equal results to the unmodified version of libModSecurity.

A 64-bit Windows build of the library obtains the same unit tests results as the Linux build. Regression test execution is also succesful, but 21 of the tests fail due to three different reasons:

 * Most of the failed tests reference paths not available on Windows (such as /tmp). These tests pass successfully if updated to reference a valid Windows path. Changing the path to the current directory (`.`) would allow tests to work on Windows and non-Windows platforms.
 * A few of the tests depend on the `/bin/echo` binary not available on Windows. These  tests run successfully using a Windows port of echo (see [UnxUtils](https://sourceforge.net/projects/unxutils/)) and updating the rules to reference the path of `echo.exe`
 * Two regression tests fail on the Windows build because the result does not match the expected one because the test depends on order not guaranteed by the underlying std C++ library used to store variables.

13 tests are skipped due to the library not including support for GeoIP and ssdeep at this time (`variable-GEO.json` depends on GeoIP, and `operator-fuzzyhash.json` depends on ssdeep).

The following is a list of the failed tests and analysis:

 * `test-cases/regression/action-ctl_audit_engine.json`
   * Test name: `auditengine : Config=Off, ctl:auditEngine=on`
     * The test configures `SecAuditLog /tmp/modsec_test_ctl_auditengine_auditlog_1.log`, which is not a valid Windows path. The test passes if the path is updated with a valid Windows path (for example, replace `/tmp` with the local path `.`).
 * `test-cases/regression/action-exec.json`
   * Test name: `Testing action :: exec (1/3)`
     * Rule `SecRule REQUEST_HEADERS \"@contains PHPSESSID\" \"id:1,t:lowercase,t:none,exec:/bin/echo\"` looks for a file (invalid LUA script) in `/bin/echo` which doesn't exist on Windows. The test passes if the path is updated in the rule and in the `parser_error` line to a valid Windows file (such as `c:/windows/system32/kernel32.dll`).
 * `test-cases/regression/auditlog.json`
   * Test name: `auditlog : messages verification - nolog,auditlog`
   * Test name: `auditlog : multiMatch data, match after last transform`
   * Test name: `auditlog : multiMatch data, match only after intermediate transform`
   * Test name: `auditlog : rule chain, multiMatch data, match after last transform`
   * Test name: `auditlog : rule chain, multiMatch data, match only after intermediate transform`
     * The tests configure `SecAuditLog` with a filename in the `/tmp/test` directory, which is not valid on Windows. The tests pass if the path is updated with a valid Windows path (for example, replace `/tmp/test` with the local path `.`).
 * `test-cases/regression/issue-2000.json`
   * Test name: `Testing audit log part H should output when deny - issue-2000`
     * The test configures `SecAuditLog /tmp/test/modsec_audit.log`, which is not a valid Windows path. The test passes if the path is updated with a valid Windows path (for example, replace `/tmp/test` with the local path `.`).
 * `test-cases/regression/issue-2423-msg-in-chain.json`
   * Test name: `Test match variable (1/n)`
   * Test name: `Test match variable (2/n)`
   * Test name: `Test match variable (3/n)`
     * Tests fails because order of matched variables is different in Windows build, due to iteration on underlying `std::unordered_multimap` (in `void AnchoredSetVariable::resolve(std::vector<const VariableValue *> *l, variables::KeyExclusions &ke)`) which doesn't guarantee ordered iteration.
 * `test-cases/regression/issue-2427.json`
   * Test name: `Tmp file retained for @inspectFile (1/2)`
   * Test name: `Tmp file retained for @inspectFile (2/2)`
     * The tests configure `SecUploadDir` & `SecTmpDir` with the path `/tmp`, which is not valid on Windows. The tests pass if the path is updated with a valid Windows path (for example, replace `/tmp` with the local path `.`).
 * `test-cases/regression/operator-inpectFile.json`
   * Test name: `Testing Operator :: @inspectFile (1/3)`
   * Test name: `Testing Operator :: @inspectFile (2/3)`
   * Test name: `Testing Operator :: @inspectFile (3/3)`
     * The rule `SecRule ARGS:res \"@inspectFile /bin/echo\" \"id:1,phase:2,pass,t:trim\"` tries to execute the `/bin/echo` binary which is not available on Windows. The tests run successfully using a Windows port of echo (see [UnxUtils](https://sourceforge.net/projects/unxutils/)) and updating the rules to reference the path of `echo.exe`.
 * `test-cases/regression/offset-variable.json`
   * Test name: `Variable offset - FILES_NAMES`
     * Test fails because order of matched files is different in Windows build, due to iteration on underlying `std::unordered_multimap` (in `void AnchoredSetVariable::resolve(std::vector<const VariableValue *> *l, variables::KeyExclusions &ke)`) which doesn't guarantee ordered iteration.
   * Test name: `Variable offset - FILES_TMP_CONTENT 1`
   * Test name: `Variable offset - FILES_TMP_CONTENT 2`
   * Test name: `Variable offset - MULTIPART_FILENAME`
   * Test name: `Variable offset - MULTIPART_NAME`
     * The tests configure `SecUploadDir` with the path `/tmp`, which is not valid on Windows. The tests pass if the path is updated with a valid Windows path (for example, replace `/tmp` with the local path `.`).
 * `test-cases/regression/request-body-parser-multipart.json`
   * Test name: `multipart parser (normal)`
     * The test configure `SecUploadDir` with the path `/tmp`, which is not valid on Windows. The test passes if the path is updated with a valid Windows path (for example, replace `/tmp` with the local path `.`).


airween:
Hi @eduar-hte,

first of all, let me congratulate you on this pull request: this is an awesome work.

Special thanks for detailed description you attached for the patch.

Although I'm sure the library has built in your environments (both on Linux and Windows), but as you can see there are some issues in our pipeline. Before we merge that, it's unavoidable to fix them.

The most critical issue is the `make check-static` result [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/8837628481/job/24290716767?pr=3132#step:9:2097). The fix is trivial, I think that would be enough to modify the [test/cppcheck_suppressions.txt](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/test/cppcheck_suppressions.txt#L57) file, and change line number, because that line [moved up](https://github.com/owasp-modsecurity/ModSecurity/pull/3132/files#diff-2909a7590dbc49a2bda561395a12b1d3019cd856bf6b4b7b988f335df51c961dL244) by removing of [these](https://github.com/owasp-modsecurity/ModSecurity/pull/3132/files#diff-2909a7590dbc49a2bda561395a12b1d3019cd856bf6b4b7b988f335df51c961dL176-L177) two lines.

Also you can take a look the SonarCloud issues, but I'm not sure those are really critical.
* [here](https://sonarcloud.io/project/security_hotspots?id=owasp-modsecurity_ModSecurity&pullRequest=3132&resolved=false&inNewCodePeriod=true) is a Docker privileging problem; I'm not a Docker user, especially not on Windows, so please help me to decide is this a critical problem or not
* the [others](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=3132&id=owasp-modsecurity_ModSecurity&open=AY8WiZ4eSIO09gNZIKye&tab=more_info) are so called "reliability" issues, perhaps there are few (or all) false positives, but we also need to investigate them.

After merging this PR with the main codebase, we need to decide how to proceed. It seems like Nginx supported on [Windows](http://nginx.org/en/docs/windows.html), but I don't know how do the modules operate on it. If the Nginx module architecture is also works on Windows, then we should make the [ModSecurity-nginx](https://github.com/owasp-modsecurity/ModSecurity-nginx) connector.

What do you think?

We also welcome you on our [Slack server](https://owasp.slack.com/) in the channel #project-modsecurity, if you are able to join.

eduar-hte:
> The most critical issue is the `make check-static` result [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/8837628481/job/24290716767?pr=3132#step:9:2097). The fix is trivial, I think that would be enough to modify the [test/cppcheck_suppressions.txt](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/test/cppcheck_suppressions.txt#L57) file, and change line number, because that line [moved up](https://github.com/owasp-modsecurity/ModSecurity/pull/3132/files#diff-2909a7590dbc49a2bda561395a12b1d3019cd856bf6b4b7b988f335df51c961dL244) by removing of [these](https://github.com/owasp-modsecurity/ModSecurity/pull/3132/files#diff-2909a7590dbc49a2bda561395a12b1d3019cd856bf6b4b7b988f335df51c961dL176-L177) two lines.

I adjusted this in commit d19047a but I reviewed the rest of the files I updated in the PR to adjust other lines as well, see commit 3f9b38c. I think these builds need to be launched again to see if the latest update solves all the issues.

> Also you can take a look the SonarCloud issues, but I'm not sure those are really critical.
> 
> * [here](https://sonarcloud.io/project/security_hotspots?id=owasp-modsecurity_ModSecurity&pullRequest=3132&resolved=false&inNewCodePeriod=true) is a Docker privileging problem; I'm not a Docker user, especially not on Windows, so please help me to decide is this a critical problem or not

I don't think this needs to change, the recommendation is to run the container with low privileges (for example, if you're using the container to run a server). This container, however, is used just to build the library when the container image is built, so the container is not expected to run afterwards. Additionally, admin privileges are required during most of the container image generation, in order to install the prerequisites (MSVC C++ compiler, CMake, Conan package manager, GIT).

> * the [others](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=3132&id=owasp-modsecurity_ModSecurity&open=AY8WiZ4eSIO09gNZIKye&tab=more_info) are so called "reliability" issues, perhaps there are few (or all) false positives, but we also need to investigate them.

I've updated some of these today in commits 14ed9d9, 96ff573, 9cd9024 & 459c0c0. These are my notes after reviewing the rest of the issues:

 * `src/utils/msc_tree.cc`: There are several reported issues in this file (about potential memory leak, null pointer dereference, removing unused variables, adding const to some function parameters, etc.) I only updated this file in commit 2207a92 to adjust the included header files to compile with MSVC, so I don't think these issues apply to this PR.
 * `src/utils/string.cc`: I just removed unused header files. I'd rather not introduce changes unrelated to this PR.
 * Recommends using fopen_s on `src/utils/shared_files.cc` & `src/utils/system.cc`, which is MSVC C++ compiler only so I think should be ignored.
 * `MultipartPartTmpFile::Open()` (in `src/request_body_processor/multipart.cc`):
   * There is a recommendation to replace a C-style array with std::string. I actually moved the declaration of this variable (`tstr`) and introduced default initialization to remove a call to memset. I could rollback my change, if necessary. I'd rather not change the call to strftime where the array is used.
   * Recommends merging an if statement with an enclosing one. This is how the unmodified code is written, and now this is part of an #ifdef block for non-Windows and Windows code, so actually having two statements prevents duplicating code.
 * There is a reported issue about nesting levels in `expandEnv` (in `src/utils/system.cc`). I introduced code in this function to implement Glob matching using the implementation from POCO C++ libraries, but I adhered to the existing code flow, which used these nesting levels (for the original implementation and a port to OpenBSD).
 * `shared_files.h`: There is a recommendation to "use the transparent equality "std::equal_to<>" and a custom transparent heterogeneous hasher with this associative string container". However, the interface of `SharedFiles` guarantees that all lookups and insertions on this map are done using `std::string` instances so adding this capability would be redundant.

> After merging this PR with the main codebase, we need to decide how to proceed. It seems like Nginx supported on [Windows](http://nginx.org/en/docs/windows.html), but I don't know how do the modules operate on it. If the Nginx module architecture is also works on Windows, then we should make the [ModSecurity-nginx](https://github.com/owasp-modsecurity/ModSecurity-nginx) connector.

We're integrating libModSecurity on a custom web server on Windows. I've only looked at the Apache & Nginx ModSecurity connectors as a reference on how to use the library. At some point I saw that Nginx should compile on Windows as well, but I didn't try it.

airween:
> > The most critical issue is the `make check-static` result [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/8837628481/job/24290716767?pr=3132#step:9:2097). 
> 
> I adjusted this in commit [d19047a](https://github.com/owasp-modsecurity/ModSecurity/commit/d19047a0dd87ecf6787f27065561704fdba4ecc0) but I reviewed the rest of the files I updated in the PR to adjust other lines as well, see commit [3f9b38c](https://github.com/owasp-modsecurity/ModSecurity/commit/3f9b38c0b01a8774379e064e8ae66f14947cad39). I think these builds need to be launched again to see if the latest update solves all the issues.

Thanks!

Because this is your first issue, the CI workflow does not start automatically when you update the PR, so you weren't able to see that there is another `check-static` [issue](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/8851062705/job/24326387773?pr=3132):

```
2024-04-27T07:09:25.0572337Z Checking src/modsecurity.cc ...
2024-04-27T07:09:25.0573014Z Defines:MS_CPPCHECK_DISABLED_FOR_PARSER=1
2024-04-27T07:09:25.0573902Z Undefines: MBEDTLS_MD5_ALT; MBEDTLS_SHA1_ALT; YYSTYPE; YY_USER_INIT
2024-04-27T07:09:25.0575313Z Includes: -Iheaders/ -I./ -Iothers/ -Isrc/ -Iothers/mbedtls/ -Isrc/parser/
2024-04-27T07:09:25.0576153Z Platform:Native
2024-04-27T07:09:25.0922287Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1...
2024-04-27T07:09:26.5722454Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;AIX...
2024-04-27T07:09:28.0162535Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;DRAGONFLY...
2024-04-27T07:09:29.4353636Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;FREEBSD...
2024-04-27T07:09:30.8117204Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;HPUX...
2024-04-27T07:09:32.2319299Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;LINUX...
2024-04-27T07:09:33.6444501Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;MACOSX...
2024-04-27T07:09:35.0453940Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;MSC_WITH_CURL...
2024-04-27T07:09:36.4378081Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;NETBSD...
2024-04-27T07:09:37.8554729Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;NO_LOGS...
2024-04-27T07:09:39.3005270Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;OPENBSD...
2024-04-27T07:09:40.7184693Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;SOLARIS...
2024-04-27T07:09:42.1278565Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WIN32...
2024-04-27T07:09:43.5504218Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_GEOIP...
2024-04-27T07:09:44.9903719Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_LIBXML2...
2024-04-27T07:09:47.2117050Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_MAXMIND...
2024-04-27T07:09:48.6488191Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_PCRE2...
2024-04-27T07:09:50.1179638Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_YAJL...
2024-04-27T07:09:51.5599001Z warning: src/modsecurity.cc,265,performance,redundantCopyLocalConst,The const variable 'startingAt' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'startingAt' to const reference.
2024-04-27T07:09:51.5602109Z warning: src/modsecurity.cc,267,performance,redundantCopyLocalConst,The const variable 'size' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'size' to const reference.
2024-04-27T07:09:51.5604892Z warning: src/modsecurity.cc,349,performance,redundantCopyLocalConst,The const variable 'startingAt' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'startingAt' to const reference.
2024-04-27T07:09:51.5607899Z warning: src/modsecurity.cc,351,performance,redundantCopyLocalConst,The const variable 'size' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'size' to const reference.
2024-04-27T07:09:51.5798124Z 118/218 files checked 45% done
```

Seems like these came with the new modifications.

> > * [here](https://sonarcloud.io/project/security_hotspots?id=owasp-modsecurity_ModSecurity&pullRequest=3132&resolved=false&inNewCodePeriod=true) is a Docker privileging problem; I'm not a Docker user, especially not on Windows, so please help me to decide is this a critical problem or not
> 
> I don't think this needs to change, the recommendation is to run the container with low privileges (for example, if you're using the container to run a server). This container, however, is used just to build the library when the container image is built, so the container is not expected to run afterwards. Additionally, admin privileges are required during most of the container image generation, in order to install the prerequisites (MSVC C++ compiler, CMake, Conan package manager, GIT).

Okay, thanks for clarification. We can mark that issue as false positive with this explanation.

> > * the [others](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=3132&id=owasp-modsecurity_ModSecurity&open=AY8WiZ4eSIO09gNZIKye&tab=more_info) are so called "reliability" issues, perhaps there are few (or all) false positives, but we also need to investigate them.
> 
> I've updated some of these today in commits [14ed9d9](https://github.com/owasp-modsecurity/ModSecurity/commit/14ed9d9ddb06490478249770ffc72f5813ce6dd2), [96ff573](https://github.com/owasp-modsecurity/ModSecurity/commit/96ff5733ea6afa8c2b134d095dd148b031dee22e), [9cd9024](https://github.com/owasp-modsecurity/ModSecurity/commit/9cd9024fe66d834618b2eff56e25640e12c031fd) & [459c0c0](https://github.com/owasp-modsecurity/ModSecurity/commit/459c0c009198b72f07802edf9b3fd5742b15f2c0). These are my notes after reviewing the rest of the issues:
> 
>     * `src/utils/msc_tree.cc`: There are several reported issues in this file (about potential memory leak, null pointer dereference, removing unused variables, adding const to some function parameters, etc.) I only updated this file in commit [2207a92](https://github.com/owasp-modsecurity/ModSecurity/commit/2207a92da7605ac04347bb20e97c45c8441ce71a) to adjust the included header files to compile with MSVC, so I don't think these issues apply to this PR.

thanks - not just for this but the others too.

Unfortunately - as you wrote too - not all commits apply to this PR. I mean some modifications makes library to build on Windows, but some other are code cleaning - which is really appreciated and very useful.
 
>     * `src/utils/string.cc`: I just removed unused header files. I'd rather not introduce changes unrelated to this PR.

Same as above: if I'm not wrong, this is a code cleaning.

I think it's important to split up the two things: feature request and clean up. Both are really important and very welcome, but when someone will inspect the code later it's hard to decide why the author sent his feature request PR with unnecessary modifications.

Because you also mentioned this above, I really hope you understand this.

> > After merging this PR with the main codebase, we need to decide how to proceed. It seems like Nginx supported on [Windows](http://nginx.org/en/docs/windows.html), but I don't know how do the modules operate on it. If the Nginx module architecture is also works on Windows, then we should make the [ModSecurity-nginx](https://github.com/owasp-modsecurity/ModSecurity-nginx) connector.
> 
> We're integrating libModSecurity on a custom web server on Windows. I've only looked at the Apache & Nginx ModSecurity connectors as a reference on how to use the library. At some point I saw that Nginx should compile on Windows as well, but I didn't try it.

Thank you, it seems very interesting (the custom web server).

Thanks for all.

eduar-hte:
> Because this is your first issue, the CI workflow does not start automatically when you update the PR, so you weren't able to see that there is another `check-static` [issue](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/8851062705/job/24326387773?pr=3132):
> 
> ```
> 2024-04-27T07:09:50.1179638Z Checking src/modsecurity.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WITH_YAJL...
> 2024-04-27T07:09:51.5599001Z warning: src/modsecurity.cc,265,performance,redundantCopyLocalConst,The const variable 'startingAt' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'startingAt' to const reference.
> 2024-04-27T07:09:51.5602109Z warning: src/modsecurity.cc,267,performance,redundantCopyLocalConst,The const variable 'size' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'size' to const reference.
> 2024-04-27T07:09:51.5604892Z warning: src/modsecurity.cc,349,performance,redundantCopyLocalConst,The const variable 'startingAt' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'startingAt' to const reference.
> 2024-04-27T07:09:51.5607899Z warning: src/modsecurity.cc,351,performance,redundantCopyLocalConst,The const variable 'size' is assigned a copy of the data. You can avoid the unnecessary data copying by converting 'size' to const reference.
> 2024-04-27T07:09:51.5798124Z 118/218 files checked 45% done
> ```
> 
> Seems like these came with the new modifications.

I think these need a suppression too, because the copy is needed to avoid the *use after free* that I fixed in commit 7cb67b0 (and which seems to have been introduced in ad28de4). The problem is that the current code is creating a reference to the last element in the list (lines 265 & 267) but then the last element is immediately removed (lines 266 & 268), which means that you have a dangling reference. Then the references are accessed in lines 273-274 & 278-279 which is incorrect. In this case, if you heed the warning, you're actually introducing a bug! :-)

(Lines 349 & 351 didn't have the problem but I marked the variables as const because they were not modified after assignment, and for consistency with the code in lines 265 & 267)

Can I add these suppressions in the code itself? If that works with your setup, I think it'd improve maintainability of the project, as you wouldn't need to update test/cppcheck_suppressions.txt when line numbers change because of unrelated updates.

eduar-hte:
> Can I add these suppressions in the code itself? If that works with your setup, I think it'd improve maintainability of the project, as you wouldn't need to update test/cppcheck_suppressions.txt when line numbers change because of unrelated updates.

I tried this in commit cc24a4f. Let's see if it works and what you think.

PS: I checked if by adding these lines I needed to update the line numbers of other suppressions on the same file and I think the ones about line 206 may not be necessary (I'll revert commit 30e1cd8 if I'm mistaken). Additionally, I don't think an r-value reference is necessary in the previous line, as the compiler will perform a return value optimization (RVO) to avoid a copy.

eduar-hte:
> I tried this in commit [cc24a4f](https://github.com/owasp-modsecurity/ModSecurity/commit/cc24a4fa4f2d640b238cb2fcce259740822022f4). Let's see if it works and what you think.

I enabled actions in my fork so I could try this. I found out that the current configuration used to run cppcheck needs the `--inline-suppr` argument for the inline suppressions to work. I've added that configuration in commit e9540ea. Once again, I hope you're ok with this.

> PS: I checked if by adding these lines I needed to update the line numbers of other suppressions on the same file and I think the ones about line 206 may not be necessary (I'll revert commit [30e1cd8](https://github.com/owasp-modsecurity/ModSecurity/commit/30e1cd854bb75f5ef5b77da2572895f445b701b5) if I'm mistaken). Additionally, I don't think an r-value reference is necessary in the previous line, as the compiler will perform a return value optimization (RVO) to avoid a copy.

No warnings/error were reported on line 206 either.

eduar-hte:
I also tried out a workflow to build the Windows version of the library adding the following configuration in a branch in my fork:

```yaml
  build-windows:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [windows-2022]
        platform: [x86, x86_64]
    steps:
      - uses: actions/checkout@v4
        with:
          submodules: true
      - name: Install Conan
        run: |
          pip3 install conan --upgrade
          conan profile detect
      - uses: ammaraskar/msvc-problem-matcher@master
      - name: vcbuild ${{ matrix.platform }}
        run: vcbuild.bat Release ${{ matrix.platform }}
        shell: cmd
```

It's simpler than the Dockerfile I added as a reference!

The 64-bit build takes less than 5 minutes, which is similar to the build time of the Linux/macOS versions. However, the 32-bit version took around 33 minutes to build. This is because the Conan dependencies need to be built in this configuration (while the 64-bit build can use the compiled binaries available in the package).

eduar-hte:
> I enabled actions in my fork so I could try this. I found out that the current configuration used to run cppcheck needs the `--inline-suppr` argument for the inline suppressions to work. I've added that configuration in commit [e9540ea](https://github.com/owasp-modsecurity/ModSecurity/commit/e9540ea563d2d50be4f245f6104728f5b6d2bbb3). Once again, I hope you're ok with this.

I ended up looking at inlining cppcheck suppressions a bit more and created a new PR (#3134) to remove those that reference line numbers so it hopefully streamlines the process of submitting changes (by avoiding the build errors and the need for additional commits to update `test/cppcheck_suppressions.txt`).

I've seen this recently also in PR #3104, where guidance is being introduced to check if this file needs updating after making changes to the library (see commit 2c8684e).

eduar-hte:
> After merging this PR with the main codebase, we need to decide how to proceed. It seems like Nginx supported on [Windows](http://nginx.org/en/docs/windows.html), but I don't know how do the modules operate on it. If the Nginx module architecture is also works on Windows, then we should make the [ModSecurity-nginx](https://github.com/owasp-modsecurity/ModSecurity-nginx) connector.

Though this was unrelated to the work we're doing with libModSecurity v3, I ended up taking a stab at building the ModSecurity-nginx module and have just created [ModSecurity-nginx PR #321](https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/321) for that! :-)

airween:
Okay, so now please pick up the changes from current v3/master, and let's see what happens.

src/modsecurity.cc
```diff
@@ -262,9 +262,11 @@ int ModSecurity::processContentOffset(const char *content, size_t len,
         std::string value;
         yajl_gen_map_open(g);
         vars.pop_back();
-        const std::string &startingAt = vars.back().str();
+        // cppcheck-suppress redundantCopyLocalConst ; copy required to avoid use after free, see commit 7cb67b0
+        const std::string startingAt = vars.back().str();
```
airween: You mentioned this modification in [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3132#issuecomment-2080954088) comment. I agree with this, but I have to tell you I would like to avoid the strings copy. JSON parsing has poor performance anyway, if we start to copy the strings, may be that will lead worse efficiency.

May be we could use `std::vector` there for `vars` and iterate with a different invariant, I mean

```C++
int l = vars.size();
while(l > 3) {
    const std::string &startingAt = vars[l];
    l--;
    ...
}
```

What do you think?

src/request_body_processor/multipart.cc
```diff
@@ -61,27 +66,35 @@ MultipartPartTmpFile::~MultipartPartTmpFile() {
 
 void MultipartPartTmpFile::Open() {
     struct tm timeinfo;
-    char tstr[300];
     time_t tt = time(NULL);
 
     localtime_r(&tt, &timeinfo);
 
-    memset(tstr, '\0', 300);
+    char tstr[300] {};
```
airween, line 73: SonarCloud [complains](https://sonarcloud.io/project/issues?resolved=false&sinceLeakPeriod=true&pullRequest=3132&id=owasp-modsecurity_ModSecurity&open=AY8WiZ1ySIO09gNZIKyb) for this solution, we have to take a look it later. (I see the `strftime` later). Actually you don't need to touch this.

airween (changes requested):

airween:
I started to review this PR, there are one relevant [request change](https://github.com/owasp-modsecurity/ModSecurity/pull/3132#discussion_r1592974630). I try to continue it in next days, I would like to review `shared_files.cc` and `system.cc` modifications. The other looks good to me.

src/request_body_processor/multipart.cc
eduar-hte: Yes, you could just do away with the initialization of the buffer altogether as it's being initialized in the next line, so zero-initialization is not needed here.

The buffer size is actually a bit excessive, as the format string needs 2 chars (for `/` & `-`), 8 for the date (`%Y%m%d`), 6 for the time (`%H%M%S`) and 1 more for the null terminator, so a size of 17 should be enough.

```cpp
char tstr[17];
strftime(tstr, sizeof(tstr)/sizeof(tstr[0]), "/%Y%m%d-%H%M%S", &timeinfo);
```

I commented on the SonarCloud reported issue before [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3132#issuecomment-2079836460), which I think is a false positive.

src/modsecurity.cc
eduar-hte: I actually didn't spend a lot of time in this code, at the time I was trying to get all the examples to compile and run on Windows and run into this crash, which I thought I may have introduced when I realized that it was actually a bug.

I noticed that other than the `reading_logs_with_offset` example, this function is actually not referenced anywhere else in libModSecurity. This feature was introduced in commit 027d50b, and at the time it was copying the element (which makes sense, as that version is correct). But then the references (and associated bug) were introduced in commit ad28de4, which probably didn't notice the issue. Curiously the references were introduced in lines 262 & 264, but not in 346 & 348, where the code is duplicated and were also updated.

eduar-hte:
> I started to review this PR, there are one relevant [request change](https://github.com/owasp-modsecurity/ModSecurity/pull/3132#discussion_r1592974630). I try to continue it in next days, I would like to review `shared_files.cc` and `system.cc` modifications. The other looks good to me.

`system.c` changes are simple, in WIN32 blocks and reference third-party implementations of glob and clock_gettime (which I initially didn't implement, as the macOS version doesn't currently implement this, but found out the public domain implementation from mingw-w64's winpthreads and decided to include it).

`shared_files.*` is more involved. I mean, the implementation in the PR is actually pretty simple and straightforward, but you need to go through all the history referenced in the commit message to analyze how the removed code was both dead-code and also had several minor issues introduced over time. I hope I was able to explain that well.

src/modsecurity.cc
eduar-hte: Another approach would be to interleave accessing the last element when the value is actually used later in the function. So we can still reference the back element, use it and then call pop_back.

In case it was not clear in my previous comment, I wonder if the question shouldn't be whether to remove the function instead.

eduar-hte: I just rewrote it using (reverse) iterators, as I frown on vector indexing.

I don't think performance is going to be much of a difference on a function that uses regexps and JSON anyway. My vote would have been on reviewing whether to remove the function altogether.

src/request_body_processor/multipart.cc
airween: I mean SonarCloud complains it before we use a C-style string here, and it suggest to use C++'s `std::string`. But the `strftime()` expects a `char *` so actually I don't know how could it be resolved. As you wrote here, we should mark it as false positive.

src/modsecurity.cc
airween: > I just rewrote it using (reverse) iterators, as I frown on vector indexing.

It was just a quick idea to avoid the removing of items from the array, and make it more easy to reference them.

What's the problem with the vector here?

> I don't think performance is going to be much of a difference on a function that uses regexps and JSON anyway. My vote would have been on reviewing whether to remove the function altogether.

Ah, thanks for noticing - this feature doesn't seem to be in use, right?

airween:
> `system.c` changes are simple, in WIN32 blocks and reference third-party implementations of glob and clock_gettime (which I initially didn't implement, as the macOS version doesn't currently implement this, but found out the public domain implementation from mingw-w64's winpthreads and decided to include it).

yepp, thanks, everything looks good.
 
> `shared_files.*` is more involved. I mean, the implementation in the PR is actually pretty simple and straightforward, but you need to go through all the history referenced in the commit message to analyze how the removed code was both dead-code and also had several minor issues introduced over time. I hope I was able to explain that well.

yes, that's very impressive fix - thanks.

Let's finish the `modsecurity.cc` affected conversation above, and I'll merge this.

src/modsecurity.cc
eduar-hte: > > I just rewrote it using (reverse) iterators, as I frown on vector indexing.
> 
> It was just a quick idea to avoid the removing of items from the array, and make it more easy to reference them.
> 
> What's the problem with the vector here?

I went with the iterator option as it's more idiomatic when iterating over STL containers. This new approach also does away with removing the items, as we just advance over three elements on the matches on every iteration while there are more than three matches left.

With regards to using a vector, because the matches are returned in a `std::list` by `searchAll`, to get them into a `std::vector` would require iterating and copying all the strings into a new container, which would be more expensive than the initial approach to fix the use after free issue by copying the element.

When I did the last update two days ago, I actually thought it the matches were already on a `std::vector`, so I used to `std::distance` to check the condition on every iteration, because on that container the operation is O(1) because the distance is computed using pointer arithmetic between the two iterator positions. However, this is not the case on a `std::list`, where you need to iterate from one element to the other to count the distance, so it's an O(n) operation. I just updated the code to have the condition be evaluated by counting the pending matches to be processed in the list and make this more efficient.

> > I don't think performance is going to be much of a difference on a function that uses regexps and JSON anyway. My vote would have been on reviewing whether to remove the function altogether.
> 
> Ah, thanks for noticing - this feature doesn't seem to be in use, right?

That seems to be the case with `processContentOffset`. The other changes in commit 027d50b introduce `Transformation::instantiate` from code that was previously in `test/unit/unit.cc`. At the time, that refactoring was needed only by `processContentOffset`. However, that function is now used to run Lua scripts too, so that part of the commit would need to be kept.

eduar-hte:
I added a new commit to set up a test suite using CTest for the Windows build.

This only required a minor update to `unit.cc` & `regression.cc` to have them set their exit code to the number of tests that failed, so that CTest can use this to determine whether the test failed or not. This doesn't affect the Unix execution of the tests because those determine whether a test failed by analyzing their console output (processed through the shell scripts `custom-test-driver` & `test-suite.sh` used to run them).

There's a new `/test/test-suite.in` with the list of tests currently in `Makefile.am` because I wanted to share the definition of the tests between the Unix and Windows builds. I don't know much about autoconf/automake so I'm just doing a simple `include` of the file from `Makefile.am`. That means that each line of the file has a `TESTS+=` prefix (necessary for `Makefile.am`), and not just the test filename, and I have to ignore that prefix when processing the file on the Windows build to setup the CTest test suite. I couldn't figure out how to generate the TESTS variable from a file with just a list of filenames with autoconf/automake.

I've updated the GitHub workflow that builds libModSecurity on Windows to also run the same tests that `make check` do on the Unix builds. There are a couple of steps that handle the issues I found when running the current tests on Windows:

 * One step creates a `tmp` directory at the root of the drive where the build and tests are done, which makes those tests that configure `SecAuditLog` work.
 * This step also creates a `bin` directory at the root of the drive too, and then copies `echo.exe` from the `usr/bin` tools included with Git in the `windows-2022` runner image to that directory, which allows tests that run `/bin/echo` to work.
   * `echo.exe` is also copied as `\bin\echo` for the test that checks if the file `/bin/echo` exist.
 * The other step (`Disable tests that don't work on Windows`) sets the `enabled` property to zero in the JSON tests that fail on Windows because the tests relies (incorrectly) on the order on which a `std::unordered_map` is iterated.
   * I had to update the `offset-variable.json` to remove a couple of comments in the JSON file, as this is not JSON compliant, and breaks the processing of the file to turn the test off.

```yaml
build-windows:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [windows-2022]
        platform: [x86_64]
        configuration: [Release]
    steps:
      - uses: actions/checkout@v4
        with:
          submodules: true
      - name: Install Conan
        run: |
          pip3 install conan --upgrade
          conan profile detect
      - uses: ammaraskar/msvc-problem-matcher@master
      - name: Build ${{ matrix.configuration }} ${{ matrix.platform }}
        shell: cmd
        run: vcbuild.bat ${{ matrix.configuration }} ${{ matrix.platform }}
      - name: Set up test environment
        working-directory: build\win32\build\${{ matrix.configuration }}
        env:
          BASE_DIR: ..\..\..\..
        shell: cmd
        run: |
          copy unit_tests.exe %BASE_DIR%\test
          copy regression_tests.exe %BASE_DIR%\test
          copy libModSecurity.dll %BASE_DIR%\test
          copy %BASE_DIR%\unicode.mapping %BASE_DIR%\test
          md \tmp
          md \bin
          copy "C:\Program Files\Git\usr\bin\echo.exe" \bin
          copy "C:\Program Files\Git\usr\bin\echo.exe" \bin\echo
      - name: Disable tests that don't work on Windows
        working-directory: test\test-cases\regression
        shell: cmd
        run: |
          jq "map(if .title == \"Test match variable (1/n)\" then .enabled = 0 else . end)" issue-2423-msg-in-chain.json > tmp.json && move /Y tmp.json issue-2423-msg-in-chain.json
          jq "map(if .title == \"Test match variable (2/n)\" then .enabled = 0 else . end)" issue-2423-msg-in-chain.json > tmp.json && move /Y tmp.json issue-2423-msg-in-chain.json
          jq "map(if .title == \"Test match variable (3/n)\" then .enabled = 0 else . end)" issue-2423-msg-in-chain.json > tmp.json && move /Y tmp.json issue-2423-msg-in-chain.json
          jq "map(if .title == \"Variable offset - FILES_NAMES\" then .enabled = 0 else . end)" offset-variable.json > tmp.json && move /Y tmp.json offset-variable.json
      - name: Run tests
        working-directory: build\win32\build
        run: |
          ctest -C ${{ matrix.configuration }} --output-on-failure
```

src/modsecurity.cc
airween: For the fix: thanks for explanation, everything is clean now. I didn't notice that `searchAll` returned with a `std::list` - thanks.

For the `processContentOffset` - I found that only in use in [examples/reading_logs_with_offset/read.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/examples/reading_logs_with_offset/read.cc#L24). Let's keep it there.

airween:
> I added a new commit to set up a test suite using CTest for the Windows build.
> 
> This only required a minor update to `unit.cc` & `regression.cc` to have them set their exit code to the number of tests that failed, so that CTest can use this to determine whether the test failed or not. This doesn't affect the Unix execution of the tests because those determine whether a test failed by analyzing their console output (processed through the shell scripts `custom-test-driver` & `test-suite.sh` used to run them).
> 
> There's a new `/test/test-suite.in` with the list of tests currently in `Makefile.am` because I wanted to share the definition of the tests between the Unix and Windows builds. I don't know much about autoconf/automake so I'm just doing a simple `include` of the file from `Makefile.am`. That means that each line of the file has a `TESTS+=` prefix (necessary for `Makefile.am`), and not just the test filename, and I have to ignore that prefix when processing the file on the Windows build to setup the CTest test suite. I couldn't figure out how to generate the TESTS variable from a file with just a list of filenames with autoconf/automake.

Thanks, let me check this in local - I'll try that soon.

> I've updated the GitHub workflow that builds libModSecurity on Windows to also run the same tests that `make check` do on the Unix builds. There are a couple of steps that handle the issues I found when running the current tests on Windows:

Sorry, but I don't see any modification if `.github` directory at all - see the [Files changed](https://github.com/owasp-modsecurity/ModSecurity/pull/3132/files) tab. Beside that I think if a new architecture appears in [workflow](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/.github/workflows/ci.yml), then we must see that in between tests results. But I also don't see even between your [tests](https://github.com/eduar-hte/ModSecurity/actions/runs/9067247186/job/24911912594).

 
>     * One step creates a `tmp` directory at the root of the drive where the build and tests are done, which makes those tests that configure `SecAuditLog` work.

Right,

>     * This step also creates a `bin` directory at the root of the drive too, and then copies `echo.exe` from the `usr/bin` tools included with Git in the `windows-2022` runner image to that directory, which allows tests that run `/bin/echo` to work.

right,
       
>       * `echo.exe` is also copied as `\bin\echo` for the test that checks if the file `/bin/echo` exist.

okay,
 
>     * The other step (`Disable tests that don't work on Windows`) sets the `enabled` property to zero in the JSON tests that fail on Windows because the tests relies (incorrectly) on the order on which a `std::unordered_map` is iterated.

right.

>       * I had to update the `offset-variable.json` to remove a couple of comments in the JSON file, as this is not JSON compliant, and breaks the processing of the file to turn the test off.

This is okay.

Thanks - it would be very to see the Windows test in the Github workflow now.

eduar-hte:
> Thanks - it would be very to see the Windows test in the Github workflow now.

oh, I've never submitted committed the changes to the workflow in my PR because I thought it'd be restricted. I'll try that now and let you know.

airween:
> > Thanks - it would be very to see the Windows test in the Github workflow now.
> 
> oh, I've never submitted committed the changes to the workflow in my PR because I thought it'd be restricted. I'll try that now and let you know.

No, there is no any restriction.

If you send the commit, GH will start the workflow at **your** branch separately. You can wait for the finish, if you want to check (but the workflow will start on main repository immediately, because the PR is active).

This is the goal (and a big advantage :smiley:).

src/modsecurity.cc
eduar-hte: > For the `processContentOffset` - I found that only in use in [examples/reading_logs_with_offset/read.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/examples/reading_logs_with_offset/read.cc#L24). Let's keep it there.

Yes, of course, that's how I came into the 'use after free' issue myself, when I was trying to get all examples and tests to run on Windows.

My point [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3132#discussion_r1593284475) was that if the issue was introduced over five years ago (commit ad28de4 is from Jan 18, 2019), it could be argued that the example has not been run in a while either.

eduar-hte:
> If you send the commit, GH will start the workflow at **your** branch separately. You can wait for the finish, if you want to check (but the workflow will start on main repository immediately, because the PR is active).

I added the Windows build yesterday.

I've made an update to also support building libModSecurity on Windows without some of the third party dependencies (LMDB, LUA, LibXML2, MaxMind & cURL), and have now configured builds without each of them.

NOTE: There is not a build in the GH workflow without LibXML2 (though the Windows build configuration now supports it) because the regression tests assume that XML support is always on. I guess that's the reason why this is not turned off either for Linux/macOS builds.

airween (approved):

airween:
> I added the Windows build yesterday.

yes, I saw it - it ran [successfully](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/9069486649/job/24919132024?pr=3132).
 
> I've made an update to also support building libModSecurity on Windows without some of the third party dependencies (LMDB, LUA, LibXML2, MaxMind & cURL), and have now configured builds without each of them.

really nice, thank you.

> NOTE: There is not a build in the GH workflow without LibXML2 (though the Windows build configuration now supports it) because the regression tests assume that XML support is always on. I guess that's the reason why this is not turned off either for Linux/macOS builds.

Yes, thank you for spotting that. That makes sense (because XML parser is a really important component of the library), but I don't understand why is there a test without YAJL. Never mind.

Thank you for this huge work, now everything is fine for me.

Please let me know if you finished the work or you still want to add something - I'm ready to merge this PR.

earias-cybele:
> Thank you for this huge work, now everything is fine for me.
> 
> Please let me know if you finished the work or you still want to add something - I'm ready to merge this PR.

That's great to hear. I'm not working on any further changes on the Windows port now, so I think you should just go ahead and integrate it (and then I can update the ModSecurity-nginx PR to sync with this -it currently references my repository in order to build libModSecurity on Windows).

eduar-hte:
> Thank you for this huge work, now everything is fine for me.
> 
> Please let me know if you finished the work or you still want to add something - I'm ready to merge this PR.

That's great to hear. I'm not working on any further changes on the Windows port now, so I think you should just go ahead and integrate it (and then I can update the ModSecurity-nginx PR to sync with this -it currently references my repository in order to build libModSecurity on Windows).

airween:
Merged, many thanks for this awesome work!

eduar-hte:
> > NOTE: There is not a build in the GH workflow without LibXML2 (though the Windows build configuration now supports it) because the regression tests assume that XML support is always on. I guess that's the reason why this is not turned off either for Linux/macOS builds.

I'll create a quick PR to annotate the regression tests that depend on libxml and have them skipped if the build doesn't include that feature.

Then I'll add the Windows build configuration to build without libxml.

> Yes, thank you for spotting that. That makes sense (because XML parser is a really important component of the library), but I don't understand why is there a test without YAJL. Never mind.

I'll try and take care of that too.

varkey98:
@airween Do you guys have a release timeline for this feature?

airween:
> Do you guys have a release timeline for this feature?

Yes: as soon as possible :smile:.

Seriously: the plan was that we make releases in last month, but there are few pending PR's which I would like to review and merge. Unfortunately I had no time to do this, but I hope I can continue the reviewing soon.

Note to everyone: I would like to improve the QA of the projects, eg. add more steps in CI/CD in case of mod_security2 and ModSecurity-nginx repositories. Last month I started to work on a special rule set. The aim is to cover all ModSecurity's target with the rules and make tests for them; tests can run with [go-ftw](https://github.com/coreruleset/go-ftw).

I don't want to wait until we finish that rule set, but with help we can progress with it, so please anyone who knows how ModSecurity works and able to write rules, take a look to [MRTS](https://github.com/owasp-modsecurity/MRTS). You should take few finished data files like [ARGS](https://github.com/owasp-modsecurity/MRTS/blob/main/util/generate_rules/data/CONF_002_TARGET_ARGS.yaml) or [ARGS_GET](https://github.com/owasp-modsecurity/MRTS/blob/main/util/generate_rules/data/CONF_004_TARGET_ARGS_GET.yaml), and I'm sure that will be easy to write data files for other target's.

varkey98:
Thanks for the reply @airween. I'd love to help out as well, but I'm a noob when it comes to Modsec rules, will try to learn more about it in the coming days :)

airween:
> Thanks for the reply @airween. I'd love to help out as well, but I'm a noob when it comes to Modsec rules, will try to learn more about it in the coming days :)

That would be awesome! You can reach us on [OWASP's Slack](https://owasp.slack.com/) server, on #project-modsecurity channel - feel free to ask anything!

Issue #3107, opened 2024-03-11
Enhancement: Improve log statement for SecArgumentsLimit issue instead of JSON parsing error


kkrupka:
**Describe the bug**

I came accross the issue, when I was sending data as JSON string: Send a JSON string with a specific array length (>1000 items) leads to a http status code 400 including the message "JSON parsing error: parse error: client cancelled parse via callback return value". If less than that threshold, JSON string can be parsed.

ModSecurity for Apache/2.9.7
Apache/2.4.57 (Debian 12)

**Logs and dumps**

Error message in modsec_audit.log
```
Message: JSON parsing error: parse error: client cancelled parse via callback return value

Message: Access denied with code 400 (phase 2). Match of "eq 0" against "REQBODY_ERROR" required. [file "/etc/modsecurity/modsecurity.conf"] [line "72"] [id "200002"] [msg "Failed to parse request body."] [data "JSON parsing error: parse error: client cancelled parse via callback return value\x0a"] [severity "CRITICAL"]
Apache-Error: [file "apache2_util.c"] [line 275] [level 3] [client XXXXX] ModSecurity: JSON parsing error: parse error: client cancelled parse via callback return value\\n [hostname "XXXXX"] [uri "XXXXX"] [unique_id "Ze6_ymoEZsb-foW2ptOCKwAAAAo"]
Apache-Error: [file "apache2_util.c"] [line 275] [level 3] [client XXXXX] ModSecurity: Access denied with code 400 (phase 2). Match of "eq 0" against "REQBODY_ERROR" required. [file "/etc/modsecurity/modsecurity.conf"] [line "72"] [id "200002"] [msg "Failed to parse request body."] [data "JSON parsing error: parse error: client cancelled parse via callback return value\\\\x0a"] [severity "CRITICAL"] [hostname "XXXXX"] [uri "XXXXX"] [unique_id "Ze6_ymoEZsb-foW2ptOCKwAAAAo"]
Action: Intercepted (phase 2)
Apache-Handler: proxy-server
Stopwatch: 1710145482393235 69656 (- - -)
Stopwatch2: 1710145482393235 69656; combined=801, p1=668, p2=10, p3=0, p4=0, p5=123, sr=132, sw=0, l=0, gc=0
Response-Body-Transformed: Dechunked
Producer: ModSecurity for Apache/2.9.7 (http://www.modsecurity.org/); OWASP_CRS/3.3.5.
Server: Apache
[...]

PR #3139, opened 2024-05-10
Enhanced logging [Issue #3107]

marcstern:
In case too many keys, indicate this in the audit log

### DIFF
--- a/apache2/msc_json.c
+++ b/apache2/msc_json.c
@@ -65,6 +65,7 @@ int json_add_argument(modsec_rec *msr, const char *value, unsigned length)
                     log_escape_ex(msr->mp, arg->value, arg->value_len));
         }
         msr->msc_reqbody_error = 1;
+        msr->json->yajl_error = apr_psprintf(msr->mp, "More than %ld JSON keys", msr->txcfg->arguments_limit);
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


apache2/msc_json.c
```diff
@@ -374,9 +375,12 @@ int json_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
 	if (msr->json->depth_limit_exceeded) {
            *error_msg = "JSON depth limit exceeded";
 	} else {
-           char *yajl_err = yajl_get_error(msr->json->handle, 0, buf, size);
-           *error_msg = apr_pstrdup(msr->mp, yajl_err);
-           yajl_free_error(msr->json->handle, yajl_err);
+        if (msr->json->yajl_error) *error_msg = msr->json->yajl_error;
```
airween, line 378: Could you set the correct indentation for this block? Just for the better readability.

airween (changes requested):

airween:
I see that you added a new "extra" line:
```
ModSecurity: JSON parsing error: More than 10 JSON keys [hostname "localhost"] [uri "/post"] [unique_id "ZkUWkhdCjBcNkeH57qvdrgAAAAA"]
ModSecurity: Warning. Match of "eq 0" against "REQBODY_ERROR" required. [file "/etc/modsecurity/modsecurity.conf"] [line "77"] [id "20
0002"] [msg "Failed to parse request body."] [data "JSON parsing error: More than 10 JSON keys"] [severity "CRITICAL"] [hostname "localhost"] [uri "/post"] [unique_id "ZkUWkhdCjBcNkeH57qvdrgAAAAA"]
```

The first line contains the cause and the second the triggered rule details (with the cause too).

It's good for me, just asking: do we need the first line? (it does not contain extra information)

Cc @kkrupka - is this what you would like to have?

kkrupka:
Perfect. Thanks!

marcstern:
I didn't add any line, it's automatically added.
Should be the case before the change (except that it would be the generic message about the "callback").

airween (approved):

airween:
> I didn't add any line, it's automatically added. Should be the case before the change (except that it would be the generic message about the "callback").

Ok, thanks.

I'm going to merge it now - thank you. Then it closes #3107.

PR #3149, opened 2024-05-23
fix: remove usage of insecure tmpnam

fzipi:
## what

- remove usage of tmpnam
- apr has primitives for file handling and unique name creation
- encapsulate mutex creation in one function `acquire_global_lock` so we DRY
- create the temporary file template (`GLOBAL_LOCK_TEMPLATE`) as a definition outside a function
- move preprocessor directives into the new function
- add prototype to modsecurity.h file to enable other functions to create global locks (if needed)

## why

- tmpnam is considered unsafe

## drawbacks

- static memory usage goes up a bit: `L_tmpnam` is 20 bytes, and `APR_MAX_PATH` should be ~ 4096.

## references

- https://sonarcloud.io/project/security_hotspots?id=owasp-modsecurity_ModSecurity&pullRequest=3148&sinceLeakPeriod=true

### DIFF
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -131,6 +131,7 @@ int acquire_global_lock(apr_global_mutex_t *lock, apr_pool_t *mp) {
     // get platform temp dir
     rc = apr_temp_dir_get(&temp_dir, mp);
     if (rc != APR_SUCCESS) {
+        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, NULL, "ModSecurity: Could not get temp dir");
         return -1;
     }
 
@@ -139,6 +140,7 @@ int acquire_global_lock(apr_global_mutex_t *lock, apr_pool_t *mp) {
 
     rc = apr_file_mktemp(&lock_name, path, 0, mp);
     if (rc != APR_SUCCESS) {
+        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, NULL, " ModSecurity: Could not create temporary file for global lock");
         return -1;
     }
     // below func always return APR_SUCCESS
@@ -146,6 +148,7 @@ int acquire_global_lock(apr_global_mutex_t *lock, apr_pool_t *mp) {
 
     rc = apr_global_mutex_create(&lock, filename, APR_LOCK_DEFAULT, mp);
     if (rc != APR_SUCCESS) {
+        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, NULL, " ModSecurity: Could not create global mutex");
         return -1;
     }
     return APR_SUCCESS;
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -122,6 +122,34 @@ msc_engine *modsecurity_create(apr_pool_t *mp, int processing_mode) {
     return msce;
 }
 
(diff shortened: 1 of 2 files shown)


airween (dismissed):

marcstern:
Sounds good overall, just a few remarks:
1. In case of error, we should log the reason and the path
2. We should create a "get_lock" function englobing apr_file_mktemp(), apr_file_name_get() & apr_global_mutex_create() instead of duplicating this code
3. temp_path_template should probably be initialized as a static value outside modsecurity_init() in case we want re-use that "get_lock" function at another place
@fzipi, can you modify the code accordingly?

fzipi:
Sure, makes sense. Will push the changes.

apache2/modsecurity.c
```diff
@@ -122,6 +122,37 @@ msc_engine *modsecurity_create(apr_pool_t *mp, int processing_mode) {
...
+int acquire_global_lock(apr_global_mutex_t *lock, apr_pool_t *mp) {
+    apr_status_t rc;
+    apr_file_t *lock_name;
+    const char *temp_dir;
+    const char *filename;
+
+    // get platform temp dir
+    rc = apr_temp_dir_get(&temp_dir, mp);
+    if (rc != APR_SUCCESS) {
+        ap_log_perror("ModSecurity: Could not get temp dir");
```
airween: Based on the [reference](https://nightlies.apache.org/httpd/trunk/doxygen/group__APACHE__CORE__LOG.html#ga2c7707a1e1930f5b5a2889ae66243aba) of `ap_log_perror()` it seems that few arguments are missing - see the correct syntax below in line [143](https://github.com/owasp-modsecurity/ModSecurity/pull/3149/files#diff-389d3c26e7333c1d93edeb63987800af15b4fe765e4a5a3c7455e13ca21a49c0R143).

I didn't check the cause of [failed CI tests](https://github.com/owasp-modsecurity/ModSecurity/pull/3149/checks), but I'm almost sure the reason is this.

airween (changes requested):

apache2/modsecurity.c
fzipi: Agreed. Did it for the others, but oversaw this one. Changed.

airween (dismissed):
LGTM

marcstern:
acquire_global_lock() prototype should be added in modsecurity.h in case we want ot reuse it elsewhere (or it should be marked as static for now).

The #ifdef __SET_MUTEX_PERMS should also move to acquire_global_lock().

When you're at it:
```
#ifdef GLOBAL_COLLECTION_LOCK
    msce->geo_lock = NULL;
#endif
```
This block can be removed, it's redundant with previous line

Thanks for you responsiveness

fzipi:
- consolidated all the preprocessor flags in the function.
- Also, I think the
```
#ifdef GLOBAL_COLLECTION_LOCK
    msce->geo_lock = NULL;
#endif
```
was a typo, and should be `msce->dbm_lock` instead (see previous line [188](https://github.com/owasp-modsecurity/ModSecurity/blob/6e82895afc6b336d144228ff84dee9cf88e77a78/apache2/modsecurity.c#L188)).

marcstern (approved):

PR #3150, opened 2024-05-24
msc_regexec() != PCRE_ERROR_NOMATCH

marcstern:
Return of msc_regexec() compared with PCRE_ERROR_NOMATCH (!=) to check if match.
Other errors may happen that would return -2, -3, ...
Matching would be incorrectly set in this case.
We should check if >= 0


marcstern:
Remark: all this code is duplicated all over the file and could be enclosed in a function:
```
        /* Figure out if we want to include this variable. */
        if (var->param == NULL) match = 1;
        else {
            if (var->param_data != NULL) { /* Regex. */
                char *my_error_msg = NULL;
                if (msc_regexec((msc_regex_t *)var->param_data, te[i].key, strlen(te[i].key), &my_error_msg) >= 0) match = 1;
            } else { /* Simple comparison. */
                if (strcasecmp(te[i].key, var->param) == 0) match = 1;
            }
        }
```

PR #3153, opened 2024-05-28
Missing null byte + optimization

marcstern:
Replace a memset to 0 by a single assignment and fixing the 0 byte missing at the end when MSC_LARGE_STREAM_INPUT is not defined


airween (approved):
LGTM

PR #3154, opened 2024-05-28
Set 'jit' variable's initial value

marcstern:


apache2/msc_pcre.c
```diff
@@ -31,11 +31,7 @@ static apr_status_t msc_pcre_cleanup(msc_regex_t *regex) {
         }
 #else
         if (regex->pe != NULL) {
-#if defined(VERSION_NGINX)
```
airween, line 34: Seems like you removed here *this* occurrence of `VERSION_NGINX` macro. But I think it would be nice to place it with a separated PR, where we remove the entire Nginx "support" from this module finally. And may be we should announce this elimination before we start it.

apache2/msc_pcre.c
```diff
@@ -152,19 +148,15 @@ void *msc_pregcomp_ex(apr_pool_t *pool, const char *pattern, int options,
 
     #ifdef WITH_PCRE_STUDY
         #ifdef WITH_PCRE_JIT
-                pe = pcre_study(regex->re, PCRE_STUDY_JIT_COMPILE, &errptr);
+                pe = pcre_study(regex->re, PCRE_STUDY_EXTRA_NEEDED|PCRE_STUDY_JIT_COMPILE, &errptr);
```
airween: I'm a little ashamed to admit it, but I couldn't find this `PCRE_STUDY_EXTRA_NEEDED` flag at all. What does it do? Could you add some note to description of this PR?

Isn't it necessary in case of pcre2?

apache2/msc_pcre.c
marcstern: Indeed. And support for IIS & NETWARE as well I imagine.
Here I removed it because there's no reason to make a difference between Nginx & Apache.

apache2/msc_pcre.c
marcstern: It's to have pcre_study() creating the extra data instead of doing in afterwards.
As PCRE1 is almost obsolete anyway, I'll revert this change that's not really mandatory.
Maybe we should challenge PCRE1 support also ...

airween (approved):

apache2/msc_pcre.c
airween: Perhaps we should bump the default PCRE engine to v2 (in a separated PR).

PR #3155, opened 2024-05-28
Possible double free

marcstern:
msr->msc_full_request_buffer is freed but not assigned to NULL. It could be freed again later


airween (approved):
LGTM

PR #3156, opened 2024-05-28
Memory leaks + enhanced logging

marcstern:
Log audit lock name in case of problem. 
Systematically log problems in update_rule_target_ex().
Fix some memory leaks in update_rule_target_ex().


apache2/re.c
```diff
@@ -252,22 +252,21 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
     if(rule != NULL)    {
 
         target_list = strdup(p2);
-        if(target_list == NULL)
-            return apr_psprintf(ruleset->mp, "Error to update target - memory allocation");;
+        if (target_list == NULL) {
+            my_error_msg = apr_psprintf(ruleset->mp, "Error to update target - memory allocation");
```
airween, line 256: Here (and at some other cases) you removed the `return` statement and after set of `my_error_msg` you jump to the label `end`. There you call an `msr_log(msr, 9, my_error_msg)` and always return with NULL.

But the calling places return this value, so may be that `char *` needs there - isn't it?

marcstern: You're right, I fixed the return value

marcstern:
replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3191

PR #3157, opened 2024-05-28
Removed code for SecStatusEngine

marcstern:
Removed code for SecStatusEngine.
Directives is still allowed but ignored.


fzipi:
If you remove SecRemoteRules, everything _curl_ could be removed IMHO from modsec (but not mlogc), which eliminates a big dependency.

fzipi:
I think probably splitting this into smaller PRs would be what we want, right? One for SecStatusEngine. Another for SecRemoteRules*?

airween:
I agree with @fzipi: we should split this PR into more smaller, but I think it's more important that we have to announce that we will eliminate these functions in next(-next) release.

Perhaps in first step we should add a warning message if someone uses any of them, and (if the user checks the logs after startup) then it can be visible our aim.

Also we should make these eliminations in v3 too, in parallel with v2, I guess.

So I would close this PR without merging - what do you think guys?

airween:
Also we should check the CI logs - all builds were fail.

marcstern:
My bad:
- SecStatusEngine is obsolete as there's no more server to receive this info.
- SecRemoteRules isn't obsolete as someone may have created a server to deliver config files

So, we indeed must split the PR.
Unless nobody uses SecRemoteRules, we cannot remove it (do we need a poll?).
curl dependency must stay as long as SecRemoteRules is supported.

As SecStatusEngine is already broken, I think there's no problem to remove the code, even without announcing it (in advance), as it doesn't do anything already (except potentially introducing a delay).

dune73:
I'm not a fan of the remote rules and namely how it was being implemented, but commercial rule vendors do use this and I am sure there are people who host their own rules centrally and then load them on startup. We have to keep this around for the time being.

marcstern:
I re-introduced the code for SecRemoteRules & SecRemoteRulesFailAction

fzipi:
Looks like pipeline is failing...

fzipi:
Why are the ISSUE_TEMPLATES also being modified in this PR? Maybe move those to a new PR?

PR #3164, opened 2024-06-02
Improve performance of VariableOrigin instances

eduar-hte:
## what

- The previous approach would create a `std::unique_ptr` and store it in a `std::list` in `VariableValue`'s `Origins`.
- The new approach stores `Origins` in a `std::vector` and constructs `VariableOrigin` elements in-place on insertion.
- Instead of having two heap-allocations for every added `VariableOrigin` instance, this performs only one.
- If multiple origins are added, `std::vector`'s growth strategy may even prevent the heap-allocation on additional insertions. 
  - When the vector is resized, there's the cost of copying the current elements.
- Introduced `reserveOrigin` method to notify when multiple insertions will be made, so that a call to `std::vector::reserve` can be made to perform a single allocation (and copy of previous elements) if necessary, and then just initialize the new elements in-place.

## why

This change increases performance of the `benchmark` test by about 2.5%.

Additionally, the refactored code simplifies code used to initialize `VariableOrigin` instances.

## misc

Included script to download OWASP Core Ruleset v4 for benchmark execution.

This is part of a series of PRs to improve performance of the library (1/n).


test/benchmark/download-owasp-v2-rules.sh
```diff
@@ -2,10 +2,7 @@
 #
```
airween: I think we must remove this file. CRS2 is unmaintained, I'm sure nobody uses that.

And from the other hand: the seclang parser fails to parse it :(
```
phase:4,rev:'2',ver:'OWASP_CRS/2.2.9',maturity:'9',accuracy:'9',chain,t:none,capture,ctl:auditLogParts=+E,block,msg:'ASP/JSPProblems loading the rules...
Rules error. File: owasp-v2/base_rules/modsecurity_crs_50_outbound.conf. Line: 39. Column: 387. Expecting an action, got:  source code leakage',logdata:'Matched Data: %{TX.0} found within %{MATCHED_VAR_NAME}: %{MATCHED_VAR}',id:'970903',tag:'OWASP_CRS/LEAKAGE/SOURCE_CODE_ASP_JSP',tag:'WASCTC/WASC-13',tag:'OWASP_TOP_10/A6',tag:'PCI/6.5.6',severity:'3'"
```

so just remove this.

airween (changes requested):

airween:
Thanks for this PR - the performance improvement is awesome! As I could measure - based on result of own `benchmark` tool it's **better** than 2.5-3% with CRS, I think it's about 5-6%.

I requested a change: could you remove the CRS2 test case? It's completely unnecessary.

test/benchmark/download-owasp-v2-rules.sh
eduar-hte: That's alright, will do (and rebase the PR too).

airween (approved):

PR #3166, opened 2024-06-07
Use standard apache logging format in error log

marcstern:
All credits to @arminabf  - see https://github.com/owasp-modsecurity/ModSecurity/pull/1997

Use the server context, like in all other places to use standard httpd format


marcstern:
Closed by mistake

PR #3172, opened 2024-06-13
CI improvement: First check syntax & always display error.log

marcstern:
First check syntax, in case no error.log is generated. this displays error on console.
Always show error.log (in sub-menu) after start. Even if error and CI processing stops.


marcstern:
replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3190

PR #3176, opened 2024-06-25
Passing address of lock instead of lock in acquire_global_lock()

marcstern:
Lock was not sent back to calling function


marcstern:
This now passed all my non-regression tests (thousands of requests)

marcstern:
replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3188

PR #3181, opened 2024-07-10
Invalid pointer access in case rule id == NOT_SET_P

marcstern:
Fixed invalid pointer access, by checking against NOT_SET_P before checking *id.

Fixed typo (duplicate string) in log message (not related).


marcstern:
replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3187

PR #3182, opened 2024-07-11
Update README.md: use submodule and use benchmark tool

airween:
## what

Update README.md: added information about how to use Git's submodules and how can anybody use the library's Benchmark tool.

## why

One of the recent PR contains some undocumented information.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,15 @@ As a dynamic library, don’t forget that libmodsecurity must be installed to a 
 
 ### Unix (Linux, MacOS, FreeBSD, …)
 
-On unix the project uses autotools to help the compilation process.
+On unix the project uses autotools to help the compilation process. Please note that if you work with `git`, do not forget to initialize submodules and update them. Here is a brief howto:
+```shell
+$ git clone https://github.com/owasp-modsecurity/ModSecurity ModSecurity
+$ cd ModSecurity
+$ git submodule init
+$ git submodule update
+```
+
+After that, you can start the build process:
 
 ```shell
 $ ./build.sh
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -57,7 +57,15 @@ As a dynamic library, don’t forget that libmodsecurity must be installed to a
 
 ### Unix (Linux, MacOS, FreeBSD, …)
 
-On unix the project uses autotools to help the compilation process.
+On unix the project uses autotools to help the compilation process. Please note that if you work with `git`, do not forget to initialize submodules and update them. Here is a brief howto:
+```shell
+$ git clone https://github.com/owasp-modsecurity/ModSecurity ModSecurity
```
fzipi: why not just using `git clone --recursive`?

README.md
```diff
@@ -235,6 +243,81 @@ $ make
 $ sudo make install
 ```
 
+### Benchmarking
+
+The source tree includes a Benchmark tool that can help measure library performance. The tool is located in the test/benchmark/ directory. The build process also creates the binary here, so you will have the tool after the compilation is finished.
```
Xhoenix: This is repeated.

README.md
```diff
@@ -235,6 +243,81 @@ $ make
...
+To run, just type:
+
+```shell
+cd test/benchmark
+$ ./benchmark
+Doing 1000000 transactions...
+
+```
+
+Okay, this is a bit much, but you can pass a lower value:
```
Xhoenix: It'd be better if it's: `You can also pass a lower value:`

README.md
```diff
@@ -235,6 +243,81 @@ $ make
...
+```shell
+$ time ./benchmark 1000
+Doing 1000 transactions...
+
+real	0m0.351s
+user	0m0.337s
+sys	0m0.022s
+```
+
+This is very fast because the benchmark uses the minimal `modsecurity.conf.default` configuration, which doesn't not too much rules:
```
Xhoenix: "which doesn't not too much rules"?

README.md
```diff
@@ -235,6 +243,81 @@ $ make
...
+
+```shell
+$ cat basic_rules.conf
+
+Include "../../modsecurity.conf-recommended"
+
+```
+
+To measure with real rules, run one of the download scripts in the same directory:
```
Xhoenix, line 283: How about `download scripts available in the same directory`?

README.md
airween: Thanks for all, fixed every issue.

Xhoenix (approved):

README.md
```diff
@@ -235,6 +243,79 @@ $ make
...
+
+Include "../../modsecurity.conf-recommended"
+
+Include "owasp-v3/crs-setup.conf.example"
+Include "owasp-v3/rules/*.conf"
+```
+
+Now the command will give much higher value.
+
+#### How the benchark works
```
fzipi:
```suggestion
#### How the benchmark works
```

README.md
```diff
@@ -235,6 +243,79 @@ $ make
...
+
+The tool is a simple wrapper application that uses the library. Creates a ModSecurity instance and a RuleSet instance, then runs a loop with the passed number. Creates a Transaction (object) and emulates real HTTP transactions.
+
+The transaction is an HTTP/1.1 GET request with some GET parameters. Added common headers then the response: headers and an XML body. Between phases it checks whether an intervention has taken place or not. All transactions are created with same data.
```
fzipi:
```suggestion
The tool is a straightforward wrapper application that utilizes the library. It creates a ModSecurity instance and a RuleSet instance, then runs a loop based on the specified number. Within this loop, it creates a Transaction object to emulate real HTTP transactions.

Each transaction is an HTTP/1.1 GET request with some GET parameters. Common headers are added, followed by the response headers and an XML body. Between phases, the tool checks whether an intervention has occurred. All transactions are created with the same data.
```

README.md
airween: Thanks, fixed.

README.md
airween: Fixed, thanks.

README.md
airween: Fixed, thanks.

fzipi (approved):

marcstern (approved):

PR #3187, opened 2024-07-20
Invalid pointer access in case rule id == NOT_SET_P 

marcstern:
Invalid pointer access in case rule id == NOT_SET_P


marcstern:
Replaces https://github.com/owasp-modsecurity/ModSecurity/pull/3181

airween (approved):

PR #3188, opened 2024-07-20
Passing address of lock instead of lock in acquire_global_lock()

marcstern:
Cretaed lock was not available in calling function


marcstern:
Replaces https://github.com/owasp-modsecurity/ModSecurity/pull/3176

airween (approved):

PR #3190, opened 2024-07-22
CI improvement: First check syntax & always display error/audit logs

marcstern:
First check syntax, in case no error.log is generated. this displays error on console.
Always show error.log (in sub-menu) after start. Even if error and CI processing stops. Same for audit log (even if empty for the moment)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           # '|| :' handles the case grep doesn't match, otherwise the script exits with 1 (error)
           errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log) || :
           if [[ -z "${errors}" ]]; then exit 0; fi
-          echo "Found errors/warnings in error.log"
+          echo "::error Found errors/warnings in error.log"
           echo "${errors}"
           exit 1
       - name: Show httpd error log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -54,7 +54,7 @@ jobs:
           # '|| :' handles the case grep doesn't match, otherwise the script exits with 1 (error)
           errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log) || :
           if [[ -z "${errors}" ]]; then exit 0; fi
-          echo "::error Found errors/warnings in error.log"
+          echo "::error::Found errors/warnings in error.log"
           echo "${errors}"
           exit 1
       - name: Show httpd error log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,11 +52,12 @@ jobs:
       - name: Search for errors/warnings in error log
         run: |
           errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log)
-          if [ -n "${errors}" ]; then
+          if [[ -n "${errors}" ]]; then
             echo "Found errors/warnings in error.log"
             echo "${errors}"
             exit 1
           fi
+          exit 0
       - name: Show httpd error log
         if: always()
         run: sudo cat /var/log/apache2/error.log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
         run: sudo systemctl restart apache2.service
       - name: Search for errors/warnings in error log
         run: |
-          errors="$(grep -E ":(?error|warn)[]]" /var/log/apache2/error.log)"
+          errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log)
           if [ -n "${errors}" ]; then
             echo "Found errors/warnings in error.log"
             echo "${errors}"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -63,4 +63,3 @@ jobs:
       - name: Show mod_security2 audit log
         if: always()
         run: sudo cat /var/log/apache2/modsec_audit.log
-        # For non-regression tests: /home/runner/work/ModSecurity/ModSecurity/tests/regression/server_root/logs/audit/audit.log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,8 +45,14 @@ jobs:
           sudo cp unicode.mapping /etc/apache2/
           sudo mkdir -p /var/cache/modsecurity
           sudo chown -R www-data:www-data /var/cache/modsecurity
+      - name: first check config (to get syntax errors)
+        run: sudo apachectl configtest
       - name: start apache with module
-        run: |
-          sudo systemctl restart apache2.service
-          sudo cat /var/log/apache2/error.log
-
+        run: sudo systemctl restart apache2.service
+      - name: Show httpd error log
+        if: always()
+        run: sudo cat /var/log/apache2/error.log
+      - name: Show mod_security2 audit log
+        if: always()
+        run: sudo cat /var/log/apache2/modsec_audit.log
+        # For non-regression tests: /home/runner/work/ModSecurity/ModSecurity/tests/regression/server_root/logs/audit/audit.log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,13 +51,12 @@ jobs:
         run: sudo systemctl restart apache2.service
       - name: Search for errors/warnings in error log
         run: |
-          errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log)
-          if [[ -n "${errors}" ]]; then
-            echo "Found errors/warnings in error.log"
-            echo "${errors}"
-            exit 1
-          fi
-          exit 0
+          # '|| :' handles the case grep doesn't match, otherwise the script exits with 1 (error)
+          errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log) || :
+          if [[ -z "${errors}" ]]; then exit 0; fi
+          echo "Found errors/warnings in error.log"
+          echo "${errors}"
+          exit 1
       - name: Show httpd error log
         if: always()
         run: sudo cat /var/log/apache2/error.log
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,6 +49,14 @@ jobs:
         run: sudo apachectl configtest
       - name: start apache with module
         run: sudo systemctl restart apache2.service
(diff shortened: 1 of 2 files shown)


marcstern:
Replaces https://github.com/owasp-modsecurity/ModSecurity/pull/3172

.github/workflows/ci.yml
```diff
@@ -45,8 +45,14 @@ jobs:
...
+      - name: first check config (to get syntax errors)
+        run: sudo apachectl configtest
       - name: start apache with module
-        run: |
-          sudo systemctl restart apache2.service
-          sudo cat /var/log/apache2/error.log
-
+        run: sudo systemctl restart apache2.service
+      - name: Show httpd error log
+        if: always()
```
theseion, line 72: Unnecessary.
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -45,8 +45,14 @@ jobs:
...
+        run: sudo cat /var/log/apache2/error.log
+      - name: Show mod_security2 audit log
+        if: always()
```
theseion, line 75:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -45,8 +45,14 @@ jobs:
...
+        run: sudo apachectl configtest
       - name: start apache with module
-        run: |
-          sudo systemctl restart apache2.service
-          sudo cat /var/log/apache2/error.log
-
+        run: sudo systemctl restart apache2.service
+      - name: Show httpd error log
+        if: always()
+        run: sudo cat /var/log/apache2/error.log
```
theseion, line 73:
```suggestion
        run: |
          errors="$(grep "ERROR" /var/log/apache2/error.log)"
          if [ -n "${errors}" ]; then
            echo "Found errors in error.log"
            echo "${errors}"
            exit 1
          fi
```

.github/workflows/ci.yml
```diff
@@ -45,8 +45,14 @@ jobs:
...
+      - name: Show mod_security2 audit log
+        if: always()
+        run: sudo cat /var/log/apache2/modsec_audit.log
```
theseion, line 76:
```suggestion
        run: |
          errors="$(grep "ERROR" /var/log/apache2/modsec_audit.log)"
          if [ -n "${errors}" ]; then
            echo "Found errors in modsec_audit.log"
            echo "${errors}"
            exit 1
          fi
```

theseion (changes requested):

.github/workflows/ci.yml
theseion: This is just a proposal, the `grep` match might need to be adjusted in order to fail the job when it shouldn't and to properly match errors. I do think that the job should fail whenever an error is encountered though, since that's not a normal situation and should be investigated.

airween:
Thanks @theseion for the review - @marcstern I think the suggestions (for using `grep` command) are useful. I'm not sure anyone will check Github logs after a CI run, so if the workflow shows there is any error in the log, that might be much helpfully.

.github/workflows/ci.yml
marcstern: I'm far from a specialist with this CI, but it seems to me that the error log won't be displayed in case of error if we remove that line. In case the service doesn't start, we want to know why, so it need to be displaye _always_.

.github/workflows/ci.yml
marcstern: Very good idea to grep for ERROR and mark as failed if found (and maybe WARNING) in the future.
However, as follow-up of my previous comment, we should also have the complete log available.
PS: the syntax of error log status is ":error]" & ":warn]"

airween: @marcstern - could you align the CI? Then we should merge this PR.

.github/workflows/ci.yml
theseion: You're absolutely right.

.github/workflows/ci.yml
theseion: Not sure it makes much sense here, but it doesn't hurt either.

.github/workflows/ci.yml
theseion: > However, as follow-up of my previous comment, we should also have the complete log available.

Makes sense.

marcstern: > @marcstern - could you align the CI? Then we should merge this PR.

done

.github/workflows/ci.yml
airween: @marcstern, @theseion - this is an unresolved conversation - what do you want to do with this?

airween:
@theseion - thanks for your help again.

Could you take a review again and resolve conversation where it's possible?

.github/workflows/ci.yml
theseion: Depends. Probably not necessary to check for anything and ok to simply dump the log.

.github/workflows/ci.yml
marcstern: Especially useful when the test fail and you want to see which rule triggered (or not)

.github/workflows/ci.yml
marcstern: This is a generic way to display the audit log. It's especially useful if a test fail and you want to see which rule triggered (or not).
On top of that, we have to add specific searches for pattern associated to requests, to ensure they were blocked (with the right rule) or they were accepted. That will be included in the "run-time" checks we'll implement soon (I hope).

.github/workflows/ci.yml
```diff
@@ -45,8 +45,22 @@ jobs:
...
+        run: sudo systemctl restart apache2.service
+      - name: Search for errors/warnings in error log
         run: |
-          sudo systemctl restart apache2.service
-          sudo cat /var/log/apache2/error.log
-
+          # '|| :' handles the case grep doesn't match, otherwise the script exits with 1 (error)
+          errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log) || :
+          if [[ -z "${errors}" ]]; then exit 0; fi
+          echo "Found errors/warnings in error.log"
```
theseion: Use GitHub commands for improved visibility in the log:
```suggestion
          echo "::error::Found errors/warnings in error.log"
```

.github/workflows/ci.yml
```diff
@@ -45,8 +45,22 @@ jobs:
...
+          echo "${errors}"
+          exit 1
+      - name: Show httpd error log
+        if: always()
+        run: sudo cat /var/log/apache2/error.log
+      - name: Show mod_security2 audit log
+        if: always()
+        run: sudo cat /var/log/apache2/modsec_audit.log
+        # For non-regression tests: /home/runner/work/ModSecurity/ModSecurity/tests/regression/server_root/logs/audit/audit.log
```
theseion: Can this comment be removed?

theseion (changes requested):
Two small changes. @marcstern please resolve the comments. Unfortunately, as owner of the review I can't resolve them myself.

.github/workflows/ci.yml
marcstern: Is "::error::" a kind of standard used somewhere else in the project?

.github/workflows/ci.yml
marcstern: done

.github/workflows/ci.yml
airween: I think the correct form is `::error`, without trailing `::` - see [this](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions) article. In CRS we use [this](https://github.com/coreruleset/coreruleset/blob/main/util/crs-rules-check/rules-check.py#L789) syntax too.

But I think @theseion will correct this.

marcstern: I fixed it

theseion: Nope, the correct syntax is `::error::`. As you can see [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-error-message), the message follows after the second `::`. It's likely that GitHub is forgiving with how it parses the message, which is why the example airween posted above works.

theseion: It's not a "standard", but a way to format log output on GitHub.

.github/workflows/ci.yml
theseion: This conversation is still unresolved.

.github/workflows/ci.yml
```diff
@@ -45,8 +45,21 @@ jobs:
...
+        run: sudo systemctl restart apache2.service
+      - name: Search for errors/warnings in error log
         run: |
-          sudo systemctl restart apache2.service
-          sudo cat /var/log/apache2/error.log
-
+          # '|| :' handles the case grep doesn't match, otherwise the script exits with 1 (error)
+          errors=$(grep -E ':(?error|warn)[]]' /var/log/apache2/error.log) || :
+          if [[ -z "${errors}" ]]; then exit 0; fi
+          echo "::error Found errors/warnings in error.log"
```
```suggestion
          echo "::error::Found errors/warnings in error.log"
```

theseion (changes requested):

.github/workflows/ci.yml
airween: Oh, so it's like
```
::error arg1=val1,arg2=val2,...argN=valN::
```
the trailing `::` is after the argument list of `error` "tag"?

theseion: Correct.

.github/workflows/ci.yml
airween: @marcstern - could you check this conversation?

.github/workflows/ci.yml
airween: @marcstern - could you check this one too? If you agree the suggestion, you can commit it here immediately.

.github/workflows/ci.yml
airween: Is there any unsolved issue with this? If not, @theseion could you resolve this?

theseion: As the review owner I cannot resolve comments. There's an open issue about this. It has to be either you or @marcstern, or you (temporarily) disable the merge check that requires all conversations to be resolved before merging.

airween (approved):

airween (approved):
LGTM

theseion (approved):

theseion:
It looks like my review doesn't count towards the merge check. Not sure why.

theseion:
@airween I found this: https://github.com/orgs/community/discussions/67622. Did you assign a specific role to me?

PR #3191, opened 2024-07-22
Memory leaks + enhanced logging

marcstern:
Log audit lock name in case of problem.
Systematically log problems in update_rule_target_ex().
Fix some memory leaks in update_rule_target_ex().

### DIFF
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -489,11 +489,7 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
 end:
     if (my_error_msg) {
         if (msr) msr_log(msr, 9, my_error_msg);
-#if !defined(MSC_TEST)
-        else {
-            ap_log_error(APLOG_MARK, APLOG_INFO, 0, NULL, " ModSecurity: Successfully appended variable");
-        }
-#endif
+        else ap_log_error(APLOG_MARK, APLOG_INFO, 0, NULL, my_error_msg);
     }
     if (target_list != NULL) free(target_list);
     if (replace != NULL) free(replace);
--- a/apache2/re.c
+++ b/apache2/re.c
@@ -333,14 +333,8 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
                 if (match == 1)  {
                     rc = msre_parse_targets(ruleset, p, rule->targets, &my_error_msg);
                     if (rc < 0) {
-                        if (msr) {
-                            msr_log(msr, 9, "Error parsing rule targets to replace variable");
-                        }
-#if !defined(MSC_TEST)
-                        else {
-                            ap_log_error(APLOG_MARK, APLOG_ERR, 0, NULL, " ModSecurity: Error parsing rule targets to replace variable");
-                        }
-#endif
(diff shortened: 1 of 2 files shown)


marcstern:
Replaces https://github.com/owasp-modsecurity/ModSecurity/pull/3156

apache2/re.c
```diff
@@ -284,29 +282,22 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
...
                 name = param;
             }
 
-            if(apr_table_get(ruleset->engine->variables, name) == NULL)   {
-                if(target_list != NULL)
-                    free(target_list);
-                if(replace != NULL)
-                    free(replace);
-                if(msr) {
-                    msr_log(msr, 9, "Error to update target - [%s] is not valid target", name);
```
airween, line 299: Reviewed the logic and seems like this generated a duplicated error log entry, right?

I mean this line `msr_log()` and the next `return apr_psprintf()`.

If it's true, it's a fix that you removed or it caused by accident?

apache2/re.c
```diff
@@ -372,28 +363,31 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
...
                     goto end;
                 }
-            } else {
+            }
+            else {
 
                 target = strdup(p);
-                if(target == NULL)
-                    return NULL;
+				if (target == NULL) {
```
airween: Seems like the indentation is wrong here. Could you correct it?

apache2/re.c
```diff
@@ -408,30 +402,22 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
...
                     name = param;
                 }
 
-                if(apr_table_get(ruleset->engine->variables, name) == NULL)   {
-                    if(target_list != NULL)
-                        free(target_list);
-                    if(replace != NULL)
-                        free(replace);
-                    if(msr) {
-                        msr_log(msr, 9, "Error to update target - [%s] is not valid target", name);
```
airween, line 424: Same as above (in line 299. in old version)

airween (changes requested):

apache2/re.c
marcstern: You're right, it's a mistake.

apache2/re.c
marcstern: I fixed it definitely in my editor

apache2/re.c
airween: I don't know it's a mistake or not - this is why I asked (though it was a bit misunderstood).

`msr_log()` produces a **debug.log** entry.

But after calling the function chain, I'm not sure if it's also generated in that log, rather in **error.log**. The chain starts [here](https://github.com/marcstern/ModSecurity/blob/v2/pr/mem_leak_re/apache2/apache2_config.c#L3848) (in **your** branch).

How can we check this?

marcstern: We return `my_error_msg` which contains the error message (if any) that will be displayed in the error.log. That'll be done by httpd.
To check this, we could add a CI test with a buggy config.

apache2/re.c
marcstern: Same answer as above

apache2/re.c
airween: Could you add a buggy config to test this?

airween (approved):

PR #3192, opened 2024-07-22
Use standard httpd logging format in error log

marcstern:
All credits to @arminabf - see https://github.com/owasp-modsecurity/ModSecurity/pull/1997

Use the server context, like in all other places to use standard httpd format.

Remark: We previously had a mix; some entries were using the standard httpd format, some not.
Entries using the standard httpd format had the [client] field duplicated.
In this PR, we use standard logging for all lines and remove the duplicated field.
Actual changes:
- entries that were not using the standard format will use it
- entries that were using the standard format won't have the [client] field duplicated anymore

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,4 +49,14 @@ jobs:
         run: |
           sudo systemctl restart apache2.service
           sudo cat /var/log/apache2/error.log
-
+      - name: Check error.log
+        run: |
+          # Send requests & check log format
+          # Valid request
+          curl -s http://127.0.01/ > /dev/null || echo $?
+          # Invalid request
+          curl -s http://127.0.01/%2e%2f > /dev/null || echo $? 
+          # Check log format
+          grep -F ModSecurity < /var/log/apache2/error.log | grep -vP "^\[[^\]]+\] \[security2:[a-z]+\] \[pid [0-9]+:tid [0-9]+\] (?:\[client [0-9.:]+\] )?ModSecurity" || exit 0
+          # grep -v succeeded => found some lines with invalid format
+          exit 1
--- a/apache2/apache2_util.c
+++ b/apache2/apache2_util.c
@@ -286,19 +286,16 @@ static void internal_log_ex(request_rec *r, directory_config *dcfg, modsec_rec *
 
 #if AP_SERVER_MAJORVERSION_NUMBER > 1 && AP_SERVER_MINORVERSION_NUMBER > 2
 	ap_log_rerror(APLOG_MARK, APLOG_ERR | APLOG_NOERRNO, 0, r,
-            "[client %s] ModSecurity: %s%s [uri \"%s\"]%s", r->useragent_ip ? r->useragent_ip : r->connection->client_ip, str1,
-            hostname, log_escape(msr->mp, r->uri), unique_id);
+            "ModSecurity: %s%s [uri \"%s\"]%s", str1, hostname, log_escape(msr->mp, r->uri), unique_id);
 #else
         ap_log_error(APLOG_MARK, APLOG_ERR | APLOG_NOERRNO, 0, r->server,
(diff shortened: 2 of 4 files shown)


fzipi:
Is there a way to actually test this change?

theseion:
Sounds good, but I'm with @fzipi. From the code it's not clear what the output will be. Does the `[client]` field still contain the same IP address that ModSecurity would have injected in every case?

marcstern:
> Sounds good, but I'm with @fzipi. From the code it's not clear what the output will be. Does the `[client]` field still contain the same IP address that ModSecurity would have injected in every case?

Yes, same logic: useragent_ip if present, client_ip otherwise

marcstern:
Code from httpd log.c
```
if (info->r) {
    len += apr_snprintf(buf + len, buflen - len,
                        info->r->connection->sbh ? "[client %s:%d] " : "[remote %s:%d] ",
                        info->r->useragent_ip,
                        info->r->useragent_addr ? info->r->useragent_addr->port : 0);
}
else if (info->c) {
    len += apr_snprintf(buf + len, buflen - len,
                        info->c->sbh ? "[client %s:%d] " : "[remote %s:%d] ",
                        info->c->client_ip,
                        info->c->client_addr ? info->c->client_addr->port : 0);
}
```

theseion:
That's good. But a test that asserts the log format would be even better.

marcstern:
I added some tests

theseion:
Thanks, that's great. @airween, isn't there a test suite for ModSecurity where the test could be added?

.github/workflows/ci.yml
```diff
@@ -49,4 +49,14 @@ jobs:
...
-
+      - name: Check error.log
+        run: |
+          # Send requests & check log format
+          # Valid request
+          curl -s http://127.0.01/ > /dev/null || echo $?
+          # Invalid request
+          curl -s http://127.0.01/%2e%2f > /dev/null || echo $? 
+          # Check log format
+          grep -F ModSecurity < /var/log/apache2/error.log | grep -vP "^\[[^\]]+\] \[security2:[a-z]+\] \[pid [0-9]+:tid [0-9]+\] (?:\[client [0-9.:]+\] )?ModSecurity" || exit 0
```
airween, line 60: I'm not sure this test cover the case what you wanted to fix, namely only one `[client]` field can be there - or you still accept the two fields log format?

This pattern:
```
\[pid [0-9]+:tid [0-9]+\] (?:\[client [0-9.:]+\] )
```
matches with this string:
```
[pid 6545:tid 6547] [client ::1:49050] [client ::1]
```
but only till the firs `[client]` field, so this substring:
```
[pid 6545:tid 6547] [client ::1:49050]
```

Was that your intention?

airween (changes requested):

airween:
> Thanks, that's great. @airween, isn't there a test suite for ModSecurity where the test could be added?

Unfortunately no, there isn't.

We have only this CI:

And as @marcstern wrote, he [added](https://github.com/owasp-modsecurity/ModSecurity/pull/3192/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR52-R62) a test (but I'm not sure that's what we expect - I'm waiting for Marc's answer).

But as I introduced few weeks ago, there is [MRTS](https://github.com/owasp-modsecurity/MRTS), where the aim is to create a full covered test suite. Feel free to review and contribute! :smile:

.github/workflows/ci.yml
marcstern: `^\[[^\]]+\] \[security2:[a-z]+\] \[pid [0-9]+:tid [0-9]+\] (?:\[client [0-9.:]+\] )?ModSecurity` doesn't match `[pid 6545:tid 6547] [client ::1:49050] [client ::1] ModSecurity ...`

airween: Ah, you're right - that was my mistake.

airween (approved):

PR #3193, opened 2024-07-22
Removed useless code

marcstern:
Mainly useless memset

### DIFF
--- a/apache2/msc_reqbody.c
+++ b/apache2/msc_reqbody.c
@@ -461,8 +461,8 @@ apr_status_t modsecurity_request_body_to_stream(modsec_rec *msr, const char *buf
         if(data == NULL)
             return -1;
 
-        memset(data, 0, msr->stream_input_length + 1 - buflen);
         memcpy(data, msr->stream_input_data, msr->stream_input_length - buflen);
+        data[msr->stream_input_length - buflen] = '\0';
 
         stream_input_body = (char *)realloc(msr->stream_input_data, msr->stream_input_length + 1);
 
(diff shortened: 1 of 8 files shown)


apache2/msc_reqbody.c
```diff
@@ -461,37 +461,30 @@ apr_status_t modsecurity_request_body_to_stream(modsec_rec *msr, const char *buf
...
 
-    memset(msr->stream_input_data, 0, msr->stream_input_length+1);
-
     if(first_pkt)   {
         memcpy(msr->stream_input_data, buffer, msr->stream_input_length);
     } else {
         memcpy(msr->stream_input_data, data, msr->stream_input_length - buflen);
         memcpy(msr->stream_input_data+(msr->stream_input_length - buflen), buffer, buflen);
     }
+    msr->stream_input_data[msr->stream_input_length] = '\0';
```
airween, line 485: If I'm right the target string's length can be 2 different value. But here it handled only the first case - will this be good?

apache2/re.c
```diff
@@ -326,14 +326,14 @@ char *update_rule_target_ex(modsec_rec *msr, msre_ruleset *ruleset, msre_rule *r
                     if(value != NULL && targets[i]->param != NULL)  {
                         if((strlen(targets[i]->param) == strlen(value)) &&
                                 strncasecmp(targets[i]->param,value,strlen(targets[i]->param)) == 0) {
-                            memset(targets[i]->name,0,strlen(targets[i]->name));
-                            memset(targets[i]->param,0,strlen(targets[i]->param));
+                            targets[i]->name[0] = '\0';
```
airween, line 329: With this evaluation the string will be cleared - was this the intention? Based on old version it seems like the string's length is more than 0.

apache2/re_operators.c
```diff
@@ -1573,12 +1561,11 @@ static const char *gsb_replace_tpath(apr_pool_t *pool, const char *domain, int l
...
     url = apr_palloc(pool, len + 1);
+    if (!url) return NULL;
     data = apr_palloc(pool, len + 1);
+    if (!data) return NULL;
 
-    memset(data, 0, len+1);
-    memset(url, 0, len+1);
-
-    memcpy(url, domain, len);
+    url[len] = 0;
```
airween: This would be better with `'\0'` (just for better readability).

airween (changes requested):

apache2/re.c
marcstern: memset sets all characters to '\0', which is useless, only the first character needs to be set to have a 0-ended string.

apache2/msc_reqbody.c
marcstern: First case, we copy `msr->stream_input_length` bytes.
Second case, we copy `msr->stream_input_length - buflen` bytes, then `buflen` bytes.
So, `msr->stream_input_length` bytes in both cases.

airween: Okay, now I reviewed the whole logic and I see. Thanks.

airween (approved):

PR #3194, opened 2024-07-22
msc_regexec() != PCRE_ERROR_NOMATCH

marcstern:
Return of msc_regexec() compared with PCRE_ERROR_NOMATCH (!=) to check if match.
Other errors may happen that would return -2, -3, ...
Matching would be incorrectly set in this case.
We must check if >= 0

### DIFF
--- a/apache2/re_variables.c
+++ b/apache2/re_variables.c
@@ -120,8 +120,7 @@ static int var_args_generate(modsec_rec *msr, msre_var *var, msre_rule *rule,
             if (var->param_data != NULL) { /* Regex. */
                 char *my_error_msg = NULL;
                 /* Run the regex against the argument name. */
-                if (!(msc_regexec((msc_regex_t *)var->param_data, arg->name,
-                    arg->name_len, &my_error_msg) == PCRE_ERROR_NOMATCH)) match = 1;
+                if (msc_regexec((msc_regex_t *)var->param_data, arg->name, arg->name_len, &my_error_msg) >= 0) match = 1;
             } else { /* Simple comparison. */
                 if (strcasecmp(arg->name, var->param) == 0) match = 1;
             }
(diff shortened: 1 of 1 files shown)


apache2/re_variables.c
```diff
@@ -120,8 +120,7 @@ static int var_args_generate(modsec_rec *msr, msre_var *var, msre_rule *rule,
             if (var->param_data != NULL) { /* Regex. */
                 char *my_error_msg = NULL;
                 /* Run the regex against the argument name. */
-                if (!(msc_regexec((msc_regex_t *)var->param_data, arg->name,
-                    arg->name_len, &my_error_msg) == PCRE_ERROR_NOMATCH)) match = 1;
+                if (msc_regexec((msc_regex_t *)var->param_data, arg->name, arg->name_len, &my_error_msg) >= 0) match = 1;
```
airween, line 123: There is another problem with the old implementation: `PCRE_ERROR_NOMATCH` is a macro of the old PCRE (PCRE3) library, but the function `msc_regex()` calls another inside function which can use the old (PCRE3) and the new (PCRE2) functions - these functions use their own macros.

I just took a quick review at PCRE(2|3) macros, I really hope that each macro with value >= 0 means that's a successfully execution.

And one more question here: if `msc_regex()` called with 4th argument (`&my_error_msg`), and there is something error (retval < 0), then there will be a string allocation and the error will be stored in this variable, right? Where will this pointer freed?

What do you think?

airween (changes requested):

apache2/re_variables.c
marcstern: You're right, `my_error_msg` won't be deallocated, but that's not important ... because it's not allocated.
That parameter is never used.
We'll have to fix that later: either remove the parameter or add the error message.

airween: Ah, now I see - `my_error_msg` address is passed, **checked(!)**, but never used... I didn't checked the code after [this] line, so I thought it's a necessary variable - but now I see it completely makes no sense.

Thanks.

airween (approved):

PR #3198, opened 2024-07-25
Add collection size in log in case of writing error

marcstern:
A collection is limited to 1008 bytes.
An error writing a collection is often due to a too big collection.
This adds the collection size to the log.


airween (approved):
LGTM.

PR #3202, opened 2024-07-31
Fixed assert() usage

marcstern:
- added some missing
- removed some invalid
- removed some that were not relevant in the context of the current function, when done in a called function

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -34,6 +34,7 @@
 const char* id_log(msre_rule* rule) {
     assert(rule != NULL);
     assert(rule->actionset != NULL);
+    assert(rule->ruleset != NULL);
     const char* id = rule->actionset->id;
     if (!id || !*id || id == NOT_SET_P) id = apr_psprintf(rule->ruleset->mp, "%s (%d)", rule->filename, rule->line_num);
 	return id;
@@ -188,6 +189,9 @@ static void copy_rules_phase(apr_pool_t *mp,
                              apr_array_header_t *child_phase_arr,
                              apr_array_header_t *exceptions_arr)
 {
+    assert(parent_phase_arr != NULL);
+    assert(child_phase_arr != NULL);
+    assert(exceptions_arr != NULL);
     rule_exception **exceptions;
     msre_rule **rules;
     int i, j;
@@ -196,11 +200,14 @@ static void copy_rules_phase(apr_pool_t *mp,
     rules = (msre_rule **)parent_phase_arr->elts;
     for(i = 0; i < parent_phase_arr->nelts; i++) {
         msre_rule *rule = (msre_rule *)rules[i];
+        assert(rule != NULL);
+        assert(rule->actionset != NULL);
         int copy = 1;
 
         if (mode == 0) {
             /* First rule in the chain. */
             exceptions = (rule_exception **)exceptions_arr->elts;
+            assert(exceptions != NULL);
             for(j = 0; j < exceptions_arr->nelts; j++) {
 
                 /* Process exceptions. */
@@ -290,13 +297,10 @@ static int copy_rules(apr_pool_t *mp, msre_ruleset *parent_ruleset,
                       msre_ruleset *child_ruleset,
                       apr_array_header_t *exceptions_arr)
 {
+    assert(parent_ruleset != NULL);
+    assert(child_ruleset != NULL);
+    assert(exceptions_arr != NULL);
     int ret = 0;
-
-    if (parent_ruleset == NULL || child_ruleset == NULL ||
-            exceptions_arr == NULL) {
-        ret = -1;
-        goto failed;
-    }
 
     copy_rules_phase(mp, parent_ruleset->phase_request_headers,
         child_ruleset->phase_request_headers, exceptions_arr);
@@ -318,6 +322,8 @@ failed:
  */
 void *merge_directory_configs(apr_pool_t *mp, void *_parent, void *_child)
 {
+    assert(_parent != NULL);
+    assert(_child != NULL);
     directory_config *parent = (directory_config *)_parent;
     directory_config *child = (directory_config *)_child;
     directory_config *merged = create_directory_config(mp, NULL);
@@ -785,11 +791,14 @@ void init_directory_config(directory_config *dcfg)
 static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
                             const char *p1, const char *p2, const char *p3)
 {
+    assert(cmd != NULL);
+    assert(dcfg != NULL);
     char *my_error_msg = NULL;
     //msre_rule *rule = NULL, *tmp_rule = NULL;
     char *rid = NULL;
     msre_rule *rule = NULL;
     extern msc_engine *modsecurity;
+    assert(modsecurity != NULL);
     int type_with_lua = 1;
     int type_rule;
     int rule_actionset;
@@ -1017,9 +1026,12 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
 static const char *add_marker(cmd_parms *cmd, directory_config *dcfg,
                               const char *p1, const char *p2, const char *p3)
 {
+    assert(cmd != NULL);
+    assert(dcfg != NULL);
     char *my_error_msg = NULL;
     msre_rule *rule = NULL;
     extern msc_engine *modsecurity;
+    assert(modsecurity != NULL);
     int p;
 
     #ifdef DEBUG_CONF
@@ -1068,11 +1080,14 @@ static const char *add_marker(cmd_parms *cmd, directory_config *dcfg,
 static const char *update_rule_action(cmd_parms *cmd, directory_config *dcfg,
                                       const char *p1, const char *p2, int offset)
 {
+    assert(cmd != NULL);
+    assert(dcfg != NULL);
     char *my_error_msg = NULL;
     msre_rule *rule = NULL;
     msre_actionset *new_actionset = NULL;
     msre_ruleset *ruleset = dcfg->ruleset;
     extern msc_engine *modsecurity;
+    assert(modsecurity != NULL);
 
     /* Get the ruleset if one exists */
     if ((ruleset == NULL)||(ruleset == NOT_SET_P)) {
@@ -1151,6 +1166,7 @@ static const char *cmd_action(cmd_parms *cmd, void *_dcfg, const char *p1)
 
 static const char *cmd_marker(cmd_parms *cmd, void *_dcfg, const char *p1)
 {
+    assert(_dcfg != NULL);
     directory_config *dcfg = (directory_config *)_dcfg;
     const char *action = apr_pstrcat(dcfg->mp, SECMARKER_BASE_ACTIONS, p1, NULL);
     return add_marker(cmd, (directory_config *)_dcfg, SECMARKER_TARGETS, SECMARKER_ARGS, action);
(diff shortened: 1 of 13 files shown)


apache2/apache2_config.c
```diff
@@ -290,14 +297,11 @@ static int copy_rules(apr_pool_t *mp, msre_ruleset *parent_ruleset,
                       msre_ruleset *child_ruleset,
                       apr_array_header_t *exceptions_arr)
 {
+    assert(parent_ruleset != NULL);
+    assert(child_ruleset != NULL);
+    assert(exceptions_arr != NULL);
     int ret = 0;
 
-    if (parent_ruleset == NULL || child_ruleset == NULL ||
```
airween, line 295: I think leaving these checks and NULL pointer handling parts is not a good idea.

As we discussed an `assert()` function can have two results:
* if someone uses `-DNDEBUG` `CFLAG`, then it terminates the program if condition is false (this is available with `--enable-assertions` configure option)
* if `-DNDEBUG` is not used (default configure setting) then `assert()` does nothing

Now consider if any of checked variable is NULL (condition is false), `-DNDEBUG` is not used -> `assert()` does nothing. Example: `parent_ruleset` is NULL. `assert()` does nothing, there is no other check, and in the next line we pass a NULL pointer to the `copy_rules_phase()` function.

Same situation in case of each removed NULL ptr check.

airween (changes requested):

apache2/apache2_config.c
marcstern: The goal of these assert() is to check conditions that can **never** happen. They are a watch dog that, once the development is complete, should be useless. This is why there's no "real" check in the normal code.
In this example, `parent_ruleset` cannot NULL; that's guaranteed by httpd (and our initialization). It cannot be NULL because of a run-time property. The only way it can be NULL is if the code is incorrect in its design (really design, not forgetting to check for NULL before calling the function). If it's incorrect in such a way, the whole behaviour of the application is screwed up. And, in almost all cases, even if we return correctly instead of referencing a NULL pointer (which is almost equivalent to an assert btw, except that logging is quite ugly), the code following the call will, at a certain moment, be in a totally inconsistent state. Because that inconsistent state may be something different than a NULL pointer, it's even worse because we could have memory corruption, which won't happen with a NULL pointer deferencing.

airween: > The goal of these assert() is to check conditions that can **never** happen.

Can we guarantee this at this moment?

> This is why there's no "real" check in the normal code.

I see, but until we can't guarantee the conditions, I think we can't leave the checks.

> In this example, `parent_ruleset` cannot NULL; 

`parent_ruleset` cannot. But I tried to review the functions where this function is called, and found this block:

https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/apache2_config.c#L429-L437

If I'm right, in this case the `rule_exceptions` will be NULL.

I think introduce `assert()` is a good direction, but at this moment we can't leave the regular checks. We just can't guarantee that every conditions are satisfied.

marcstern: I re-added all NULL check at run-time, with an error message on stderr (as it's a critical problem).

apache2/apache2_config.c
```diff
@@ -2755,11 +3129,16 @@ static const char *cmd_hash_engine(cmd_parms *cmd, void *_dcfg, const char *p1)
...
+    // Normally useless code, left to be safe for the moment
+    if (_dcfg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, NULL, "cmd_hash_param: _dcfg is NULL");
+        return NULL;
+    }
     directory_config *dcfg = (directory_config *)_dcfg;
 
-    if (dcfg == NULL) return NULL;
-
-    if (p1 == NULL) return NULL;
```
airween, line 2762: This check is still missing.

apache2/apache2_config.c
```diff
@@ -2777,12 +3156,18 @@ static const char *cmd_hash_param(cmd_parms *cmd, void *_dcfg, const char *p1)
...
+    // Normally useless code, left to be safe for the moment
+    if (_dcfg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, NULL, "cmd_hash_key: _dcfg is NULL");
+        return NULL;
+    }
     directory_config *dcfg = (directory_config *)_dcfg;
     char *p1 = NULL;
 
-    if (dcfg == NULL) return NULL;
-    if (_p1 == NULL) return NULL;
```
airween, line 2784: This check is still missing.

apache2/apache2_config.c
```diff
@@ -2978,11 +3375,16 @@ static const char *cmd_hash_method_rx(cmd_parms *cmd, void *_dcfg,
...
+    // Normally useless code, left to be safe for the moment
+    if (_dcfg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, NULL, "cmd_httpBl_key: _dcfg is NULL");
+        return NULL;
+    }
     directory_config *dcfg = (directory_config *)_dcfg;
 
-    if (dcfg == NULL) return NULL;
-
-    if (p1 == NULL) return NULL;
```
airween, line 2985: This check is still missing.

apache2/apache2_config.c
```diff
@@ -3034,11 +3440,12 @@ static const char *cmd_pcre_match_limit_recursion(cmd_parms *cmd,
 static const char *cmd_geo_lookup_db(cmd_parms *cmd, void *_dcfg,
         const char *p1)
 {
+    assert(cmd != NULL);
+    assert(p1 != NULL);
     const char *filename = resolve_relative_path(cmd->pool, cmd->directive->filename, p1);
     char *error_msg;
     directory_config *dcfg = (directory_config *)_dcfg;
-    if (dcfg == NULL) return NULL;
```
airween, line 3040: This check is still missing.

apache2/apache2_config.c
```diff
@@ -3085,12 +3494,14 @@ static const char *cmd_unicode_codepage(cmd_parms *cmd,
...
                                      const char *p1, const char *p2)
 {
+    assert(cmd != NULL);
+    assert(p1 != NULL);
+    assert(p2 != NULL);
     const char *filename = resolve_relative_path(cmd->pool, cmd->directive->filename, p1);
     char *error_msg;
     long val = 0;
     directory_config *dcfg = (directory_config *)_dcfg;
-    if (dcfg == NULL) return NULL;
```
airween, line 3092: This check is still missing.

apache2/apache2_config.c
```diff
@@ -3120,11 +3531,12 @@ static const char *cmd_unicode_map(cmd_parms *cmd, void *_dcfg,
 static const char *cmd_gsb_lookup_db(cmd_parms *cmd, void *_dcfg,
                                      const char *p1)
 {
+    assert(cmd != NULL);
+    assert(p1 != NULL);
     const char *filename = resolve_relative_path(cmd->pool, cmd->directive->filename, p1);
     char *error_msg;
     directory_config *dcfg = (directory_config *)_dcfg;
-    if (dcfg == NULL) return NULL;
```
airween, line 3126: This check is still missing.

apache2/re_operators.c
```diff
@@ -1025,10 +1047,6 @@ static int msre_op_rx_execute(modsec_rec *msr, msre_rule *rule, msre_var *var, c
     assert(var != NULL);
     assert(error_msg != NULL);
     msc_regex_t *regex = (msc_regex_t *)rule->op_param_data;
-    if (!regex) {
```
airween, line 1028: This check is still missing.

apache2/re_operators.c
```diff
@@ -2619,12 +2656,14 @@ static int msre_op_endsWith_execute(modsec_rec *msr, msre_rule *rule, msre_var *
...
 static int msre_op_strmatch_param_init(msre_rule *rule, char **error_msg) {
+    assert(rule != NULL);
+    assert(rule->ruleset != NULL);
+    assert(error_msg != NULL);
     const apr_strmatch_pattern *compiled_pattern;
     char *processed = NULL;
     const char *pattern = rule->op_param;
     unsigned short int op_len;
 
-    if (error_msg == NULL) return -1;
```
airween, line 2627: This check is removed, but I don't see the logic here: first the code checks if it NULL, returns if it is. Then sets it to NULL...

apache2/re_operators.c
```diff
@@ -3207,12 +3251,18 @@ static int cpf_verify(const char *cpfnumber, int len) {
...
+    if (error_msg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, rule->ruleset->mp, NULL, "msre_op_verifyCPF_init: error_msg is NULL");
+        return -1;
+    }
     const char *errptr = NULL;
     int erroffset;
     int options = 0;
     msc_regex_t *regex;
 
-    if (error_msg == NULL) return -1;
```
airween, line 3215: Same as above.

apache2/re_operators.c
```diff
@@ -4105,11 +4165,6 @@ static int msre_op_fuzzy_hash_init(msre_rule *rule, char **error_msg)
     threshold_str = data;
 #endif
 
-    if (error_msg == NULL)
```
airween, line 4108: Same as above.

apache2/re_operators.c
```diff
@@ -4241,9 +4296,15 @@ static int msre_op_fuzzy_hash_execute(modsec_rec *msr, msre_rule *rule,
...
+    assert(rule != NULL);
+    assert(rule->ruleset != NULL);
+    assert(error_msg != NULL);
+    if (error_msg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, rule->ruleset->mp, NULL, "msre_op_inspectFile_init: error_msg is NULL");
+        return -1;
+    }
     char *filename = (char *)rule->op_param;
 
-    if (error_msg == NULL) return -1;
```
airween, line 4246: Same as above.

airween (changes requested):
I found some removed and still missing check.

apache2/re_operators.c
marcstern: we check the pointer and we assign the content of the pointer

apache2/re_operators.c
marcstern: same answer

apache2/re_operators.c
marcstern: same answer

apache2/re_operators.c
marcstern: same answer

apache2/apache2_config.c
marcstern: I added it

apache2/apache2_config.c
marcstern: done

apache2/apache2_config.c
marcstern: done

apache2/apache2_config.c
marcstern: done

apache2/apache2_config.c
marcstern: done

apache2/apache2_config.c
marcstern: done

apache2/re_operators.c
marcstern: I incorrectly added the check for memory allocation problem but there's no allocation

airween (approved):

PR #3208, opened 2024-08-05
Build on macOS with Apple silicon (arm64)

eduar-hte:
## what

Support building libModSecurity on macOS with Apple silicon (arm64)

## why

Apple has been moving the macOS platform to Apple silicon (arm64) over the last few years (as of June 2023, the entire Mac lineup uses Apple silicon chips).

This PR fixes build configuration issues in pcre & yajl that prevented the library from compiling on newer versions of macOS (which probably apply to the x86_64 version of the OS too).

Additionally, it updates the GitHub quality assurance workflow to run now on macOS 14, which is an Apple silicon (arm64) runner image (see [here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,7 +78,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        os: [macos-12]
+        os: [macos-14]
         configure:
           - {label: "with parser generation", opt: "--enable-parser-generation" }
           - {label: "wo curl",    opt: "--without-curl" }
@@ -91,10 +91,12 @@ jobs:
           - {label: "with pcre2", opt: "--with-pcre2" }
     steps:
       - name: Setup Dependencies
-        # autoconf, curl, pcre2 not installed because they're already
+        # curl, pcre2 not installed because they're already
         # included in the image
         run: |
-          brew install automake \
+          brew install autoconf \
+                       automake \
+                       libtool \
                        yajl \
                        lmdb \
                        lua \
--- a/build/pcre.m4
+++ b/build/pcre.m4
@@ -27,30 +27,41 @@ else
 
     AC_MSG_CHECKING([for libpcre config script])
 
-    for x in ${test_paths}; do
-        dnl # Determine if the script was specified and use it directly
-        if test ! -d "$x" -a -e "$x"; then
-            PCRE_CONFIG=$x
-            pcre_path="no"
-            break
-        fi
+    AC_CHECK_PROG([PCRE_CONFIG_IN_ENV], [pcre-config], [yes], [no])
 
-        dnl # Try known config script names/locations
-        for PCRE_CONFIG in pcre-config; do
-            if test -e "${x}/bin/${PCRE_CONFIG}"; then
-                pcre_path="${x}/bin"
+    if test "$PCRE_CONFIG_IN_ENV" = "yes"; then
+        AC_MSG_NOTICE([pcre-config found in envinronment])
+
+        PCRE_CONFIG=pcre-config
+        pcre_path="no"
+    else
+        AC_MSG_NOTICE([pcre-config not found in environment. checking known paths])
+
+        for x in ${test_paths}; do
+            dnl # Determine if the script was specified and use it directly
+            if test ! -d "$x" -a -e "$x"; then
+                PCRE_CONFIG=$x
+                pcre_path="no"
                 break
-            elif test -e "${x}/${PCRE_CONFIG}"; then
-                pcre_path="${x}"
+            fi
+
+            dnl # Try known config script names/locations
+            for PCRE_CONFIG in pcre-config; do
+                if test -e "${x}/bin/${PCRE_CONFIG}"; then
+                    pcre_path="${x}/bin"
+                    break
+                elif test -e "${x}/${PCRE_CONFIG}"; then
+                    pcre_path="${x}"
+                    break
+                else
+                    pcre_path=""
+                fi
+            done
+            if test -n "$pcre_path"; then
                 break
-            else
-                pcre_path=""
             fi
         done
-        if test -n "$pcre_path"; then
-            break
-        fi
-    done
+    fi
 
     if test -n "${pcre_path}"; then
         if test "${pcre_path}" != "no"; then
--- a/build/yajl.m4
+++ b/build/yajl.m4
@@ -78,6 +78,29 @@ else
 #    fi
 fi
 
+# FIX: if the include directory in CFLAGS ends with "include/yajl",
+# remove the suffix "/yajl". the library header files are included
+# using the prefix (for example, #include <yajl/yajl_tree.h>), and
+# this is even the case for the library itself (for example,
+# yajl_tree.h includes yajl/yajl_common.h).
+
+new_cflags=""
+
+for flag in $YAJL_CFLAGS; do
+    case "$flag" in
+        -I*/include/yajl)
+            new_flag="${flag%/yajl}"
+            new_cflags="$new_cflags $new_flag"
+            ;;
+        *)
+            new_cflags="$new_cflags $flag"
+            ;;
+    esac
+done
+
+YAJL_CFLAGS="$new_cflags"
+
+
 if test -z "${YAJL_LDADD}"; then
     if test -z "${YAJL_MANDATORY}"; then
         if test -z "${YAJL_DISABLED}"; then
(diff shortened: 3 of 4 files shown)


eduar-hte:
> This PR fixes build configuration issues in pcre & yajl that prevented the library from compiling on newer versions of macOS (which probably apply to the x86_64 version of the OS too).

The `configure` step of the build on macOS 14 with Apple silicon would trigger the following error:

```
```

Having address the `pcre` issue, the build would still fail with the following error:

```
Error: transaction.cc:19:10: fatal error: 'yajl/yajl_tree.h' file not found
#include <yajl/yajl_tree.h>
         ^~~~~~~~~~~~~~~~~~
1 error generated.
make[3]: *** [libmodsecurity_la-transaction.lo] Error 1
```

eduar-hte:
> Additionally, it updates the GitHub quality assurance workflow to run now on macOS 14, which is an Apple silicon (arm64) runner image (see [here](https://github.com/actions/runner-images?tab=readme-ov-file#available-images)).

I replaced the `macos-12` (x86_64) runner with the `macos-14` (arm64) instead of just adding it in order not to multiply the number of configurations to build in the workflow.

NOTE: This doesn't imply dropping support for x86_64 on macOS though, but rather have visibility on the more representative architecture on the platform now.

airween (approved):
LGTM

airween:
Thanks again @eduar-hte.

eduar-hte:
airween@, I'm seeing builds today failing on the new `macos-14` runner image.

I found about it while working on another PR and experiencing failures on the mac builds, which I initially thought were related to those changes, but it seems they were unrelated.

It's always the same regression tests:

```
(   2/  2/   4): test/test-cases/regression/config-secremoterules.json
```

but the configuration changes (for example `macOS (wo ssdeep)` or `macOS (wo libxml)`). However, the same configuration then builds on a follow up build, so it doesn't appear to be specific to one of them.

For example:

image

but then:

image

I think I saw you experience this yourself working on the `CHANGES` PR today, right? (this [build](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10267749317)) shows the same issue, but a follow up build run successfully.

eduar-hte:
> It's always the same regression tests:
> 
> ```
> (   2/  2/   4): test/test-cases/regression/config-secremoterules.json
> ```

The strangest thing is that if I'm not reading things wrong, the test file only has *three* tests, and when the builds fail, they show that *four* were executed.

If you take a look at the `CHANGES` PR build that failed that I mentioned before, the `macOS (wo ssdeep)` configuration triggers the error and shows the listing above.

However, in the same build (and I would expect regression test `.json` file because all are built using the same repository commit) the `macOS (wo libxml)` configuration, which ran the tests successfully, shows:

```
(   3/  0/   3): test/test-cases/regression/config-secremoterules.json
```

airween:
Hi @eduar-hte,

yes, this is a very weird behavior and I'm not sure I understand what I see.

It's weird because when you sent the PR, all checks passed successfully. When I merge the PR, tests run again. In *this* case, the tests on macOS were failed. Then I re-run them again by hand, and all of them finished successfully.

And yes, there is a pending PR where I collect merged PR's for `CHANGES`, and after each commit the macOS tests were failed too.

It is very annoying that the failed tests are not consistent: every time a different test fails.

@theseion, @fzipi do you have any ideas what happens here?

airween:
> > It's always the same regression tests:
> > ```
> > (   2/  2/   4): test/test-cases/regression/config-secremoterules.json
> > ```

yes, I noticed that.

> The strangest thing is that if I'm not reading things wrong, the test file only has _three_ tests, and when the builds fail, they show that _four_ were executed.

yes, you're right.

> If you take a look at the `CHANGES` PR build that failed that I mentioned before, the `macOS (wo ssdeep)` configuration triggers the error and shows the listing above.

Maybe, I didn't noticed that. But now I looked up few other jobs and found this one:

[Here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10269582746/attempts/1) you can see that the macOS (wo geoip) was the success test, and macOS (with pcre2) was the first failed build.

This was a job after merging your PR.

> However, in the same build (and I would expect regression test `.json` file because all are built using the same repository commit) the `macOS (wo libxml)` configuration, which ran the tests successfully, shows:
> 
> ```
> (   3/  0/   3): test/test-cases/regression/config-secremoterules.json
> ```

Look at [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10269616749/job/28416185054#step:8:74)

```
(   0/  0/   3): test/test-cases/regression/config-secremoterules.json
```

I think we definitely need to investigate this.

Do you have any macOS-14 environment?

eduar-hte:
> > If you take a look at the `CHANGES` PR build that failed that I mentioned before, the `macOS (wo ssdeep)` configuration triggers the error and shows the listing above.
> 
> Maybe, I didn't noticed that. But now I looked up few other jobs and found this one:
> (...)

Yes, my point was that it's clearly related to macOS 14, but not the PR that I was currently working on (which I ended up spending a couple of hours trying to think how my changes could have broken those regression tests).

> I think we definitely need to investigate this.

This is strange and started happening just today so I think it's related to a change in environment.

I went back to see all the actions executed in my fork for the `macos-apple-silicon` branch, and I noticed that all of them that run successfully (other than the *in development* changes in the branch, of course) were running macOS 14.5 (see [here](https://github.com/eduar-hte/ModSecurity/actions/runs/10254399048/job/28369092184) from 4 days ago) and 14..4.1 (see [here](https://github.com/eduar-hte/ModSecurity/actions/runs/9213340390/job/25347003905) from 3 months ago (!) when I worked on this).

And as far as I can see, all of the builds that fail are running on macOS 14.6, a GitHub runner image which was released *yesterday* and is currently being deployed, see [here](https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20240805.3).

> Do you have any macOS-14 environment?

Yes, I have a macOS mini that is running macOS 14.5, and all the tests run successfully there after cloning the repository and building `v3/master`.

I'll try to see if I can update it to macOS 14.6 and see if I can reproduce it there.

eduar-hte:
> And as far as I can see, all of the builds that fail are running on macOS 14.6, a GitHub runner image which was released _yesterday_ and is currently being deployed, see [here](https://github.com/actions/runner-images/releases/tag/macos-14-arm64%2F20240805.3).

The `CHANGES` PR build that failed (see [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10267749317/job/28409052099)) ran on this new runner image (macOS 14.6):

```
Current runner version: '2.317.0'
Operating System
  macOS
  14.6
  23G80
Runner Image
  Image: macos-14-arm64
  Version: 20240805.3
```

while the one that run successfully two hours later (see [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10269616749)) ran on the older one (macOS 14.5):

```
Current runner version: '2.317.0'
Operating System
  macOS
  14.5
  23F79
Runner Image
  Image: macos-14-arm64
  Version: 20240728.1
```

eduar-hte:
> I'll try to see if I can update it to macOS 14.6 and see if I can reproduce it there.

I've updated my mac mini and found that the test `Include remote rules - failed download (Abort)` fails:

```
Test name: Include remote rules - failed download (Abort).
Reason: 
failed!
Expected a parser error.
Expected: Failed to download: HTTP response code said error
Produced: Rules error. File: https://gist.githubusercontent.com/zimmerle/a4c1ec028999f7df71d0cc80f4f271ca/raw/4c74363bf4eae974180f1a82007196e58729dd16/modsecurity-regression-test-secremoterules-bonga.txt. Line: 1. Column: 0. 
 - Failed to download: Failure when receiving data from the peer
``` 

The test configuration depends on a specific error message from curl/OS (`HTTP response code said error`), which seems to have changed (`Failure when receiving data from the peer`).

I'll create a new quick PR to simplify the test configuration and look just for `Failed to download`, which is the prefix added by the libModSecurity parser when logging the parser error.

PS: Someone should take a look at why `custom-test-driver` & `test-suite.sh` mix up their *total tests* count (probably when a test fail).

eduar-hte:
> I'll create a new quick PR to simplify the test configuration and look just for `Failed to download`, which is the prefix added by the libModSecurity parser when logging the parser error.

A build with this change running on macOS 16 shows that this change solves the issue, see [here](https://github.com/eduar-hte/ModSecurity/actions/runs/10274343366/job/28430841427#step:5:4).

I had to trigger the build a couple of times because the first two ones were assigned a macOS 14.5 runner image, which would not have proven that the change addressed the issue.

PR #3209, opened 2024-08-05
Add cleanup methods to complete C based ABI

eduar-hte:
## what

Add cleanup methods to C based ABI to allow the library to be used by languages that do not link with the standard C library.

## why

The C/C++ interface of libModSecurity assumes that clients that use the library are also able to link with the standard C library to free any resources allocated by libModSecurity functions and returned to the client (transferring their ownership to them).

If the client wants to build a shared library version of libModSecurity (such as .dll on Windows) and use it through the exported API, it will not be able to deallocate these resources.

This PR adds two cleanup methods to allow clients that depend just on the library's ABI to free those resources.

These methods follow the approach and terminology of existing cleanup methods in the C API (such as `msc_rules_cleanup` & `msc_cleanup`) and aims to provide completeness for the lifecycle management of all objects allocated by the library.

## changes

 * `void msc_rules_error_cleanup(const char *error)`
   * Deallocates an error message returned by the `msc_rules_xxx` methods.
 * `void msc_intervention_cleanup(ModSecurityIntervention *it)`
   * Deallocates the `url` and/or `log` members of the `ModSecurityIntervention` structure that may be allocated in a call to `msc_intervention`.

### DIFF
--- a/examples/simple_example_using_c/test.c
+++ b/examples/simple_example_using_c/test.c
@@ -68,6 +68,8 @@ int main (int argc, char **argv)
     msc_process_response_body(transaction);
     msc_process_logging(transaction);
 end:
+    if(error != NULL)
+        msc_rules_error_cleanup(error);
     msc_rules_cleanup(rules);
     msc_cleanup(modsec);
 
--- a/src/rules_set.cc
+++ b/src/rules_set.cc
@@ -311,6 +311,21 @@ extern "C" int msc_rules_add(RulesSet *rules, const char *plain_rules,
 }
 
 
+/**
+ * @name    msc_rules_error_cleanup
+ * @brief   Deallocates an error message buffer returned by a msc_rules_xxx function.
+ *
+ * This is a helper function to free the error message buffer allocated
+ * by a msc_rules_xxx function.
+ *
+ * @param error Error message pointer.
+ *
+ */
+extern "C" void msc_rules_error_cleanup(const char *error) {
+    free((void*) error);
+}
+
+
 extern "C" int msc_rules_cleanup(RulesSet *rules) {
     delete rules;
     return true;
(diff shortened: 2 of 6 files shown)


examples/simple_example_using_c/test.c
```diff
@@ -68,6 +68,8 @@ int main (int argc, char **argv)
     msc_process_response_body(transaction);
     msc_process_logging(transaction);
 end:
+    if(error != NULL)
```
airween, line 71: My 2 cents: it would be better to check inside the called function (`msc_rules_error_cleanup()`) that the pointer is not null - what do you think?

Reasons:
* simplifies caller code
* encapsulation
* consistent, reusable

src/rules_set.cc
```diff
@@ -311,6 +311,21 @@ extern "C" int msc_rules_add(RulesSet *rules, const char *plain_rules,
...
+ * @brief   Deallocates an error message buffer returned by a msc_rules_xxx function.
+ *
+ * This is a helper function to free the error message buffer allocated
+ * by a msc_rules_xxx function.
+ *
+ * @param error Error message pointer.
+ *
+ */
+extern "C" void msc_rules_error_cleanup(const char *error) {
+    free((void*) error);
```
airween, line 325:
```C
    if (error != NULL) {
        free((void*) error);
    }
```

airween (changes requested):

examples/simple_example_using_c/test.c
eduar-hte: I only added the check there because it was an example and wanted to illustrate that it was not necessary to always call `msc_rules_error_cleanup`. I could clarify it with a comment to state this and that it's not a validation.

However, the function itself doesn't need to have the validation because `free` already handles this scenario (See [here](https://en.cppreference.com/w/c/memory/free): 'If ptr is a null pointer, the function does nothing.')

airween (approved):

Issue #3138, opened 2024-05-10
Encountering SIGSEGV when parsing multiple rule sets in parallel


rkrishn7:
**Describe the bug**

Hello!

I'm encountering an error when creating multiple rule sets and adding rules to each of them in a multi-threaded environment.

**Logs and dumps**

I've traced it back to `yylex` in `yy::seclang_parser::parse()`:

```
#0  0x00007bb94eea02d8 in yylex(modsecurity::Parser::Driver&) ()
   from /usr/lib/libmodsecurity.so.3
#1  0x00007bb94ee78c0e in yy::seclang_parser::parse() ()
   from /usr/lib/libmodsecurity.so.3
#2  0x00007bb94eeba170 in modsecurity::Parser::Driver::parse(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /usr/lib/libmodsecurity.so.3
#3  0x00007bb94eeba4a6 in modsecurity::Parser::Driver::parseFile(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) ()
   from /usr/lib/libmodsecurity.so.3
#4  0x00007bb94eed0a17 in modsecurity::RulesSet::loadFromUri(char const*) ()
   from /usr/lib/libmodsecurity.so.3
#5  0x00007bb94eed0b68 in msc_rules_add_file ()
   from /usr/lib/libmodsecurity.so.3

...(omitted for brevity)
```

Is this a known issue? Is `Rules` meant to be initialized and filled only once and in a single-threaded context?

airween:
Hi @rkrishn7,

could you share your (minimal) config (at least to emulate this behavior)? And I assume you have your own application, so a minimal code example also would be fine.

rkrishn7:
Hey @airween,

Sure, I was able to create a minimal reproducible example in C (using libmodsecurity v3.0.12):

```c
#include <pthread.h>
#include <stdio.h>
#include <stdlib.h>
#include <modsecurity/modsecurity.h>
#include <modsecurity/rules_set.h>

void *create_rules(void *arg)
{
    const char *plain_rules = arg;

    const char *err = NULL;

    RulesSet *rules;

    rules = msc_create_rules_set();
    if (rules == NULL)
    {
        return NULL;
    }

[...]

Issue #3054, opened 2024-02-07
Inadequate locking for multithreading in libModSecurity In-Memory persistent storage


martinhsv:
This issue does not apply to the main, current use case of libModSecurity with nginx, since that is a multi-process but single-threaded application. This finding is based on code inspection only; I did not build a multi-threaded environment to prove a crash or other tangible errors.

While adding the expirevar support I noticed that, while the in-memory option includes some internal locking, it does so only for updates.

The problem is that if ThreadA is performing a read-only action using an interator and, while that is ongoing, ThreadB makes an update to the data, it could invalidate the iterator that ThreadA is in the midst of using. (Note that the update by ThreadB might be a deletion, but an insertion could also invalidate the iterator that is in-use by ThreadA.)

airween:
Hi @martinhsv,

thanks for sharing the results of your inspection. I try to write a sample code that will help us to investigate the behavior that you explained, but it would be a help if you could provide a rule set which helps to cause the (possible) behavior?

airween:
Hi @martinhsv,

thanks again for your report.

I made a small test source which embeds the libmodsecurity3 library to a multi threaded application.

The sad news is that it not necessary to add the `expirevar` to make a segfault.

If the debug log is turned on, it is guaranteed to cause a segfault.
```
Thread 3 "multithr" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff3c226c0 (LWP 877)]
0x00007ffff7e6aca3 in modsecurity::RulesSet::debug(int, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) () from /lib/x86_64-linux-gnu/libmodsecurity.so.3
```
If I turn off the debug.log, then I got a segfault in `evaluate`:
```
Thread 3 "multithr" received signal SIGSEGV, Segmentation fault.
[Switching to Thread 0x7ffff3c226c0 (LWP 910)]
0x00007ffff7e6af67 in modsecurity::RulesSet::evaluate(int, modsecurity::Transaction*) () from /lib/x86_64-linux-gnu/libmodsecurity.so.3
```

I'm going to finish that example tool soon (add the correct `Makefile.am` and so on...), and will add that to `examples/` directory (or first just share that through a gist on Github).

[...]

PR #3210, opened 2024-08-05
Fixed shared files deadlock in a multi-threaded Windows application

eduar-hte:
## what

Fix potential deadlock in multi-threaded usage of the library on Windows.

## why

The shared files Windows implementation introduced in PR #3132 works in multi-process single-threaded contexts but it doesn't work correctly in single-process multi-threaded contexts.

## miscellaneous

The issue is that the `LockFileEx` Win32 function works on a per-handle basis. In a multi-process context, each process will have called `SharedFiles::add_new_handler` when initializing the `SharedFile` and obtained a handle, and thus locking will work.

When running ModSecurity in a single process using multiple threads, the initialization of the `SharedFile` will happen once and the handle will be shared by all threads. Then, if two threads try to write to the same shared file concurrently, they may deadlock as one of them will lock the file (by calling `LockFileEx`) and then proceed to write to the file. If before writing to the file and unlocking it, another thread calls `LockFileEx` on the same handle, the attempt to write to the file will lock generating a deadlock.

The new implementation replaces usage of `LockFileEx`/`UnlockFileEx` with a named mutex to lock access to the shared file.

 * A named mutex is used to support multi-process scenarios.
   * The mutex name is generated using the filename to support multiple shared files (such as that for the debug and audit logs).
 * This assumes that both process will initialize the `SharedFile` instance using the same filename (which is expected as they'd be using the same configuration file)

### DIFF
--- a/src/utils/shared_files.cc
+++ b/src/utils/shared_files.cc
@@ -17,8 +17,7 @@
 
 #include <fcntl.h>
 #ifdef WIN32
-#include <windows.h>
-#include <io.h>
+#include <algorithm>
 #endif
 
 
@@ -34,7 +33,32 @@ SharedFiles::handlers_map::iterator SharedFiles::add_new_handler(
         return m_handlers.end();
     }
 
-    return m_handlers.insert({ fileName, {fp, 0} }).first;
+#ifdef WIN32
+    // replace invalid characters for a Win32 named object
+    auto tmp = fileName;
+    std::replace(tmp.begin(), tmp.end(), '\\', '_');
+    std::replace(tmp.begin(), tmp.end(), '/', '_');
+
+    // use named mutex for multi-process locking support
+    const auto mutexName = "Global\\ModSecurity_" + tmp;
+
+    HANDLE hMutex = CreateMutex(NULL, FALSE, mutexName.c_str());
+    if (hMutex == NULL) {
+        error->assign("Failed to create mutex for shared file: " + fileName);
+        fclose(fp);
+        return m_handlers.end();
+    }
+#endif
+
+    auto handler = handler_info {
+        fp,
+#ifdef WIN32
+        hMutex,
+#endif
+        0
+    };
+    // cppcheck-suppress resourceLeak ; false positive, fp is closed in SharedFiles::close
+    return m_handlers.insert({ fileName, handler }).first;
 }
 
 
@@ -69,6 +93,9 @@ void SharedFiles::close(const std::string& fileName) {
     if (it->second.cnt == 0)
     {
         fclose(it->second.fp);
+#ifdef WIN32
+        CloseHandle(it->second.hMutex);
+#endif
 
         m_handlers.erase(it);
     }
@@ -92,9 +119,11 @@ bool SharedFiles::write(const std::string& fileName,
     lock.l_type = F_WRLCK;
     fcntl(fileno(it->second.fp), F_SETLKW, &lock);
 #else
-    auto handle = reinterpret_cast<HANDLE>(_get_osfhandle(fileno(it->second.fp)));
-    OVERLAPPED overlapped = { 0 };
-    ::LockFileEx(handle, LOCKFILE_EXCLUSIVE_LOCK, 0, MAXDWORD, MAXDWORD, &overlapped);
+    DWORD dwWaitResult = WaitForSingleObject(it->second.hMutex, INFINITE);
+    if (dwWaitResult != WAIT_OBJECT_0) {
+        error->assign("couldn't lock shared file: " + fileName);
+        return false;
+    }
 #endif
 
     auto wrote = fwrite(msg.c_str(), 1, msg.size(), it->second.fp);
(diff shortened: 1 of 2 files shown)


eduar-hte:
An alternative fix still using relying on `LockFileEx`/`UnlockFileEx` was evaluated, which would have required opening and closing the file whenever the shared file is written to (in each call to `SharedFiles::write`) in order for each thread to have a different handle to try and lock the file, as expected by the Win32 APIs.

However, this implementation was discarded for performance reasons (in comparison to the one using the mutex) and to minimize differences between the implementation on different platforms.

airween:
Hi @eduar-hte,

> The issue is that the `LockFileEx` Win32 function works on a per-handle basis. In a multi-process context, each process will have called `SharedFiles::add_new_handler` when initializing the `SharedFile` and obtained a handle, and thus locking will work.
> 
> When running ModSecurity in a single process using multiple threads, the initialization of the `SharedFile` will happen once and the handle will be shared by all threads.

Seems you have fixed this issue only if the platform is WIN32, right?

Don't you think this could happen on other platforms too? (I never used libmodsecurity3 on MT environment so I don't have any experiences, but there are some other reported issues: #3138, #3054)

Have ever you faced other problems in MT environment? (See mentioned issues above)

eduar-hte:
> Seems you have fixed this issue only if the platform is WIN32, right?

Yes, this issue was introduced in my implementation of the Windows port. 🤦

> Don't you think this could happen on other platforms too?

I'm not sure whether a deadlock will be triggered in a MT environment, but I understand that locking using `F_SETLKW` is not suitable in that context, as the documentation states:

```
The threads in a process share locks.  In other words, a
multithreaded program can't use record locking to ensure that
threads don't simultaneously access the same region of a file.
```

If you look at the history of the `SharedFiles` implementation, it was previously using a mutex that was shared across shared-memory between processes. This was replaced with the lock `fcntl` in commit 3d20304.

> (I never used libmodsecurity3 on MT environment so I don't have any experiences, but there are some other reported issues: #3138, #3054)
> 
> Have ever you faced other problems in MT environment? (See mentioned issues above)

I'm currently working through the integration of libModSecurity in a MT web server on Windows, and have only found this issue (I introduced) at this point.

With regards to #3138, I haven't run into that issue because initialization of rules happens only once and in a single thread before the `ModSecurity` instance is used to analyze requests. 

NOTE: I think it's reasonable for the library to expect configuration changes to be synchronized, while the critical scenario should be to be able to handle transactions concurrently without synchronization for the sake of performance.

I'll keep an eye on the issue mentioned in #3054, thanks for the heads up.

I'd guess that we could say that MT support is at this point not guaranteed (or experimental).

airween:
> Yes, this issue was introduced in my implementation of the Windows port. 🤦

thanks :smiley: 
 
> I'm not sure whether a deadlock will be triggered in a MT environment, but I understand that locking using `F_SETLKW` is not suitable in that context, as the documentation states:
> 
> ```
> The threads in a process share locks.  In other words, a
> multithreaded program can't use record locking to ensure that
> threads don't simultaneously access the same region of a file.
> ```
> 
> If you look at the history of the `SharedFiles` implementation, it was previously using a mutex that was shared across shared-memory between processes. This was replaced with the lock `fcntl` in commit [3d20304](https://github.com/owasp-modsecurity/ModSecurity/commit/3d2030426c22c41a85cda92289dc284094fba5fd).

"Great". So we can state that the library cannot be used in MT environment.
 
> > (I never used libmodsecurity3 on MT environment so I don't have any experiences, but there are some other reported issues: #3138, #3054)
> > Have ever you faced other problems in MT environment? (See mentioned issues above)
> 
> I'm currently working through the integration of libModSecurity in a MT web server on Windows, and have only found this issue (I introduced) at this point.

Thanks - a good news at least.

> With regards to #3138, I haven't run into that issue because initialization of rules happens only once and in a single thread before the `ModSecurity` instance is used to analyze requests.

Yes, this is what I suggested the reporter.
 
> I'll keep an eye on the issue mentioned in #3054, thanks for the heads up.

Thanks. I created a new directory with full MT under examples, but I wasn't able to reproduce the mentioned issue. (I suspended the work on that issue and haven't added the example yet.)
 
> I'd guess that we could say that MT support is at this point not guaranteed (or experimental).

Yes, at least.


Thanks again for this great PR, I'm going to merge that.

airween (approved):

eduar-hte:
> Thanks. I created a new directory with full MT under examples, but I wasn't able to reproduce the mentioned issue. (I suspended the work on that issue and haven't added the example yet.)

There is already an example that shows some multithreading support in the library, which is `reading_logs_via_rule_message`. This example launches 100 threads to create and process 10000 transactions concurrently. The update for the example to be multithreaded was done back in 2017 in commit 2a50852.

When I was initially getting to know the library and working on the Windows port, I went through all the examples in order to validate the port, and based on this and some other references in the code to multithreading I assumed it was supported.

> > I'd guess that we could say that MT support is at this point not guaranteed (or experimental).
> 
> Yes, at least.

I think it makes sense to create an issue to keep track of MT support and link to this PR and the issues that you mentioned before. I'll write that up later.

eduar-hte:
> I think it makes sense to create an issue to keep track of MT support and link to this PR and the issues that you mentioned before. I'll write that up later.

#3215

eduar-hte:
> I'll keep an eye on the issue mentioned in #3054, thanks for the heads up.

For the sake of reference, trying to address this in PR #3216.

PR #3212, opened 2024-08-06
Do not assume ModSecurityIntervention argument to transaction::intervention has been initialized/cleaned

eduar-hte:
## what

Minor changes to transaction::intervention to guarantee to caller that all members of the `ModSecurityIntervention` have been initialized if an intervention should be made and that the return value of the function is consistent with the state of the transaction.

## why

The current version of Transaction::intervention assumes (and implicitly requires) that the `ModSecurityIntervention` argument has been initialized/cleaned before the call. 

This is not necessarily expected because the parameter is used as an *output* parameter and it's not clear that any input is provided to the function.

Additionally, in order to make the API more robust, when the function returns that an intervention should be made, initialize all members in the structure to guarantee to the caller that all of them can be safely accessed and interpreted.

## changes

- Keep `m_it->disruptive` value and use it as return value to guarantee that the value is correct.
  - If `m_it->disruptive` is `false` and the `it` argument has not been initialized/cleaned, the function may incorrectly return a non-zero value.
- When a disruptive intervention is being reported by the function, defensively initialize `log` & `url` fields to `NULL` if there's no such data to provide to the caller.
  - If the caller has not initialized/cleaned those fields in the `it` argument, after returning from `transaction::intervention`, the user can safely read the `log` & `url` fields (and call cleanup functions such as `intervention::free` or `msc_intervention_cleanup` -introduced in PR #3209).
- Minor changes to improve & fix documentation of `Transaction::intervention` & `msc_intervention`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,8 +12,8 @@ jobs:
       matrix:
         os: [ubuntu-22.04]
         platform:
-          - {label: "x64", arch: "amd64", configure: "--enable-assertions=yes"}
-          - {label: "x32", arch: "i386", configure: "PKG_CONFIG_PATH=/usr/lib/i386-linux-gnu/pkgconfig CFLAGS=-m32 CXXFLAGS=-m32 LDFLAGS=-m32 --enable-assertions=yes"}
+          - {label: "x64", arch: "amd64", configure: ""}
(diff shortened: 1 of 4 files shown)


airween (approved):
LGTM

airween:
Thanks @eduar-hte!

PR #3216, opened 2024-08-07
Prevent concurrent access to data in InMemoryPerProcess' resolveXXX methods

eduar-hte:
## what

Lock access to `InMemoryPerProcess`' data in `resolveXXX` methods.

## why

As reported in #3054, the `resolveXXX` methods in `InMemoryPerProcess` are not acquiring a lock/mutex to prevent concurrent access to the data structures that may be modified at the same time from other threads, and thus triggering undefined behaviour.

## changes

 * Replaced inheritance of `std::unordered_multimap` data structure with data member to prevent potential clients to use it without acquiring the mutex to protect concurrent access.
 * Updates to `InMemoryPerProcess::resolveFirst`
   * Updated the code to store the expired var in `expiredVars` to delete them after iterating over the range (and releasing the read lock, as `delIfExpired` needs to acquire it for exclusive access), as the current call to `delIfExpired` would invalidate the range triggering undefined behaviour on the followingiteration.
   * Noticed that in commit 118e1b3 the call to `delIfExpired` in this function is done using `it->second.getValue()` instead of `it->first`, which seems incorrect (based on similar code in other `resolveXXX` functions).
 * Updated `InMemoryPerProcess::delIfExpired` to use `std::find_if` (with a lambda that matches both the key and the `isExpired` condition) because the data structure is a multimap. The version introduced in commit 118e1b3 could find an entry (not necessarily the first, because the map is unordered) where `isExpired` is `false` and exit, while another entry could be expired.
 * Replaced pthreads lock with std C++11 `std::shared_mutex`
   * Provides exclusive/shared lock access so that multiple readers can access the data at the same time, but only one writer. this is used to favor query performance by allowing more concurrent access to the data until an update needs to be performed.
   * Simplifies acquisition and disposal of lock/mutex with std::lock_guard, which has RAII semantics.
   * NOTE: Because `std::shared_mutex` is not recursive, calls to another function that tries to acquire the lock will fail. Introduced `__store` & `__updateFirst` helper methods to workaround this.

### DIFF
--- a/build/win32/CMakeLists.txt
+++ b/build/win32/CMakeLists.txt
@@ -2,13 +2,13 @@ cmake_minimum_required(VERSION 3.24)
 
 set(BASE_DIR ${CMAKE_CURRENT_LIST_DIR}/../..)
 
-option(WITHOUT_LMDB    "Include LMDB support"    OFF)
-option(WITHOUT_LUA     "Include LUA support"     OFF)
-option(WITHOUT_LIBXML2 "Include LibXML2 support" OFF)
-option(WITHOUT_MAXMIND "Include MaxMind support" OFF)
-option(WITHOUT_CURL    "Include CURL support"    OFF)
+option(WITH_LMDB    "Include LMDB support"    OFF)
+option(WITH_LUA     "Include LUA support"     ON)
+option(WITH_LIBXML2 "Include LibXML2 support" ON)
+option(WITH_MAXMIND "Include MaxMind support" ON)
+option(WITH_CURL    "Include CURL support"    ON)
 
-option(USE_ASAN        "Build with Address Sanitizer" OFF)
+option(USE_ASAN     "Build with Address Sanitizer" OFF)
 
 # common compiler settings
 
@@ -93,17 +93,17 @@ set(HAVE_SSDEEP   0)    # should always be zero, no conan package available
 
 macro(enable_feature flag option)
   if(${option})
-    set(${flag} 0)
+    set(${flag} 1)  # ON
   else()
-    set(${flag} 1)
+    set(${flag} 0)  # OFF
   endif()
 endmacro()
 
-enable_feature(HAVE_LMDB ${WITHOUT_LMDB})
-enable_feature(HAVE_LUA ${WITHOUT_LUA})
-enable_feature(HAVE_LIBXML2 ${WITHOUT_LIBXML2})
-enable_feature(HAVE_MAXMIND ${WITHOUT_MAXMIND})
-enable_feature(HAVE_CURL ${WITHOUT_CURL})
+enable_feature(HAVE_LMDB ${WITH_LMDB})
+enable_feature(HAVE_LUA ${WITH_LUA})
+enable_feature(HAVE_LIBXML2 ${WITH_LIBXML2})
+enable_feature(HAVE_MAXMIND ${WITH_MAXMIND})
+enable_feature(HAVE_CURL ${WITH_CURL})
 
 include(${CMAKE_CURRENT_LIST_DIR}/ConfigureChecks.cmake)
 
--- a/build/win32/CMakeLists.txt
+++ b/build/win32/CMakeLists.txt
@@ -110,7 +110,6 @@ include(${CMAKE_CURRENT_LIST_DIR}/ConfigureChecks.cmake)
 configure_file(config.h.cmake ${BASE_DIR}/src/config.h)
 
 find_package(PCRE2 REQUIRED)
-find_package(PThreads4W REQUIRED)
 find_package(Poco REQUIRED)
 find_package(dirent REQUIRED)		# used only by tests (check dirent::dirent refernces)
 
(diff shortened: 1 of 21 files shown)


eduar-hte:
@airween: I took a look at the multithreaded example you mentioned [here](https://github.com/owasp-modsecurity/ModSecurity/issues/3054#issuecomment-2000320917) (and adapted it to align with the similar code in the `reading_logs_via_rule_message` example). I could run it successfully (even after adding more threads and requests).

Do you want me to upload it as part of this PR both as an example and also if you want to test some other rules to stress test this code?

airween:
@eduar-hte - please let us know if you finished the work on this PR.

Btw: have you seen the Sonarcloud's [notifications](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3216&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)?

eduar-hte:
> @eduar-hte - please let us know if you finished the work on this PR.

I think I'm done with the changes, as I tried to limit them to introduce the `std::shared_mutex` to support exclusive & shared access to the data (protecting the areas that @martinhsv alluded to in #3054.

There is still this question I tried to ask you in [this previous comment](https://github.com/owasp-modsecurity/ModSecurity/pull/3216#issuecomment-2274342581) about whether we should add the multithreaded example you were working on in the context of that issue and that I reworked to align with the changes to the `reading_logs_via_rule_message` example.

> Btw: have you seen the Sonarcloud's [notifications](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3216&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)?

Yes, I went through all of them.

Most of them (`Extract this nested code block into a separate function` & `Replace this declaration by a structured binding declaration`) are about the existing `InMemoryPerProcess` code, which I tried not to update to limit the changes in this PR to introducing the `std::shared_mutex`. However, because I had to introduce scopes to unlock the mutex before deleting expired entries, Sonarcloud looks at that as new/updated code.

Some of them (`Replace this use of "std::thread" with "std::jthread"`) are related to suggestions to use C++20 features, which we have to disregard as we're currently targeting C++17 (as I said before, it'd be nice to see how to change Sonarcloud's project configuration for ModSecurity to report issues targeting C++17 instead).

airween:
> Do you want me to upload it as part of this PR both as an example and also if you want to test some other rules to stress test this code?

Yes, I think that would be fine - thanks!

build/win32/CMakeLists.txt
```diff
@@ -2,13 +2,13 @@ cmake_minimum_required(VERSION 3.24)
 
 set(BASE_DIR ${CMAKE_CURRENT_LIST_DIR}/../..)
 
-option(WITHOUT_LMDB    "Include LMDB support"    OFF)
-option(WITHOUT_LUA     "Include LUA support"     OFF)
-option(WITHOUT_LIBXML2 "Include LibXML2 support" OFF)
-option(WITHOUT_MAXMIND "Include MaxMind support" OFF)
-option(WITHOUT_CURL    "Include CURL support"    OFF)
+option(WITH_LMDB    "Include LMDB support"    OFF)
```
airween, line 5: I'm not familiar in Cmake, but if the option's first argument is `WITH_`, why is the third is `OFF`? By the other flags it the 1st is `WITH_`, then the last one is `ON`.

airween:
> > Do you want me to upload it as part of this PR both as an example and also if you want to test some other rules to stress test this code?
> 
> Yes, I think that would be fine - thanks!

@eduar-hte sorry, seems you've already uploaded with commit 3d9027b3 - thanks.

airween (dismissed):
LGTM - thanks for this very impressive PR.

eduar-hte:
> @eduar-hte sorry, seems you've already uploaded with commit [3d9027b](https://github.com/owasp-modsecurity/ModSecurity/commit/3d9027b3c10160f8cdbd27f79058130bb7f463b2) - thanks.

No, those were updates to the existing example, `reading_logs_via_rule_message`, which is multithreaded.

I've just uploaded your test from #3054 (and rewritten it to use libModSecurity C++ API and `std::thread`, to align it with the changes to the other example).

build/win32/CMakeLists.txt
eduar-hte: I transformed the options from WITHOUT_XXX to WITH_XXX to make them easier to use (and not have to think about double negatives to include a feature or not).

At the same time, while working on this PR I noticed that the default Windows build was using LMDB instead of the in-memory backend for collections, which is the default for non-Windows platforms. So I adjusted this to align them (and thus changed the GitHub workflow too, to align the names of the configurations).

eduar-hte:
> @eduar-hte - please let us know if you finished the work on this PR.

@airween - this is now done. when this is merged, issue #3054 should be closed.

airween (approved):

airween (approved):

PR #3218, opened 2024-08-08
Remove unnecessary dynamic casts

eduar-hte:
## what

This short PR removes a number of `dynamic_cast` casts in the codebase, some of which are performed twice and could be rewritten to perform the *runtime* cast only once.

## why

`dynamic_cast` is performed at runtime and requires inspecting the object's RTTI information to perform the cast. Executing it more than once incurs a bit of a performance cost that could be avoided by rewriting the code snippet. Additionally, the rewritten code is shorter and thus easier to read and maintain.

For example (from `src/rule_with_actions.cc`):

```cpp
                    } else if (dynamic_cast<actions::Severity *>(a)) {
                        m_severity = dynamic_cast<actions::Severity *>(a);
```

can be rewritten as:

```cpp
                    } else if (auto sa = dynamic_cast<actions::Severity *>(a)) {
                        m_severity = sa;
```

## misc

This is part of a series of PRs to improve performance of the library (2/n). Previous: #3164

### DIFF
--- a/headers/modsecurity/rule_with_operator.h
+++ b/headers/modsecurity/rule_with_operator.h
@@ -50,7 +50,7 @@ class RuleWithOperator : public RuleWithActions {
     bool evaluate(Transaction *transaction,
         std::shared_ptr<RuleMessage> rm) override;
 
-    void getVariablesExceptions(Transaction *t,
+    void getVariablesExceptions(Transaction &t,
         variables::Variables *exclusion, variables::Variables *addition);
     inline void getFinalVars(variables::Variables *vars,
         variables::Variables *eclusion, Transaction *trans);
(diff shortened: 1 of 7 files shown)


airween:
Thanks - I tried to measure the performance with this patch, but didn't get any significant difference (but it might improve it in real environments).

Never mind, the patch makes the code more visible indeed, going to merge.

airween (approved):
LGTM - thanks.

eduar-hte:
> Thanks - I tried to measure the performance with this patch, but didn't get any significant difference (but it might improve it in real environments).

That's right, these changes are not going to make a noticeable difference, but the goal is to shave off as much as possible over time.

PR #3079, opened 2024-02-09
Bump the C++ version from C++11 to C++17

MirkoDziadzka:
This will allow the usage of more modern features in the future.

PR #3219, opened 2024-08-08
Simplifiy configuration to build using std C++17

eduar-hte:
## what

Leveraged autoconf again to check whether the C++ compiler supports the required standard version and build using it.

## why

Simplify std C++ used to build libModSecurity codebase to minimize configuration that depends on this setting.

This streamlines updating to C++20 in the future as it needs to be changed in a single place.

## changes

- Replaced the outdaded `ax_cxx_compile_stdcxx_11.m4` macro with the latest version of `ax_cxx_compile_stdcxx` which supports C++17.
  - https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html
  - https://raw.githubusercontent.com/autoconf-archive/autoconf-archive/e4e5269db2764b9f53d759c24750ac6ca38e02ea/m4/ax_cxx_compile_stdcxx.m4

## references

The project configuration was bumped to C++17 in PR #3079 (see commit 367a871).


eduar-hte:
> This streamlines updating to C++20 in the future as it needs to be changed in a single place.

I've already tested this on a separate branch, which requires only the change to the following line in `configure.ac`:

```
AX_CXX_COMPILE_STDCXX(20, noext, mandatory)
```

NOTE: And also a similar change in `vcbuild.bat`, as the Windows build doesn't rely on `autoconf`.

airween (approved):
LGTM - thanks!

PR #3220, opened 2024-08-08
Creating a std::string with a null pointer is undefined behaviour

eduar-hte:
## what

Minor changes to avoid initializing a `std::string` with a null pointer.

## why

cppreference mentions this about the constructor that receives a `const char *` (see [here](https://en.cppreference.com/w/cpp/string/basic_string/basic_string)):

```
9) Constructs the string with the contents initialized with a copy of the null-terminated
character string pointed to by s. The length of the string is determined by the first null
character. The behavior is undefined if [s, s + Traits::length(s)) is not a valid range (for
example, if s is a null pointer).
```

## references

C++23 introduces a deleted constructor to prevent this in static scenarios, which is how this issue was detected.

### DIFF
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -333,9 +333,9 @@ class RulesSetProperties {
         case FalseConfigBoolean:
             return "False";
         case PropertyNotSetConfigBoolean:
+        default:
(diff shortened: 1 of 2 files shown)


airween (approved):
LGTM - thanks!

Issue #3215, opened 2024-08-07
Review and document multi-threading support and limitations


eduar-hte:
The goal of this issue is to collect and centralize information about multi-threading support of the current version of modSecurity , as the topic has come up in a number of issues and there are documented guidelines about usage in this context.

# Multi-threading support

The library is expected to work in multi-threaded scenarios, as stated in #1726. This means that any issues should be related to incorrect usage or bugs in the current implementation.

"The modSecurity life cycle is divided into different stages. The stage that the rules are loaded is not threading safe by design. (...) Once the rules are loaded multiple requests can share the same `RulesSet` object, leading to parallelism while addressing different requests in different processes or threads." (Source: #2536, [here](https://github.com/owasp-modsecurity/ModSecurity/issues/2536#issuecomment-800679431))

## Notes

- `Operator` & `Action` (including `Transformation`) objects are created in the first stage, when rules are loaded and parsed, so their initialization is not synchronized. If their evaluation updates internal state, this needs to be protected to prevent issues in multithreaded contexts.

# Examples

The repository currently includes an example of usage of the library in a multi-threaded context, see [reading_logs_via_rule_message](https://github.com/owasp-modsecurity/ModSecurity/tree/v3/master/examples/reading_logs_via_rule_message) in the `examples` directory.

# Potential issues/limitations

 * Shared files (audit or debug log) *potential* deadlock or incorrect behaviour on non-Windows platforms
   * NOTE: This has not been reproduced or reported as an issue, but is documented here as discussed in the context of a similar issue in the Windows port.
   * A deadlock using shared files was found in the initial port of the library to Windows (#3132) which was addressed in #3210.
   * The PR discussion mentions the possibility of a similar issue (or other incorrect behaviour) happening in non-Unix platforms due to the use of the `F_SETLKW` `fcntl` to lock the shared files.
   * The library previously used a mutex (using shared memory to make it available to other processes in a multi-process context), but this was replaced by the `F_SETLKW` `fcntl` in commit 3d20304 due to an unlocking issue under heavy load using nginx, where an acquired lock would not be released when the process was killed.
[...]

Issue #1954, opened 2018-11-14
How to use pm operator with mutex ?


luchy0120:
Dear Modsecurity team , I want to ask when using Modsecurity with pm operator ,  whether we should add --enable-mutex-on-pm during configuration .
Is it used under multithread situation ?

zimmerle:
Hi @luchy0120,

This is not meant to be used in a regular environment. That was implemented to facilitate the testing of the library in different environments. Should be safe to use without the mutex. Do you have any specific use case that you can think in a benefit of using it?

luchy0120:
Hi @zimmerle , is acmp_process_quick thread-safe ? Could pm be used in a multi-thread environment ?

Issue #1573, opened 2017-09-26
double free in ModSecurity3 'reading_logs_via_rule_message' example and OWASP CRS


majordaw:
Hi,
when I would like to use the OWASP CRS in a multi-threaded environment (e.g: in the reading_logs_via_rule_message example) I get segfaults in the _acmp_build_binary_tree(ACMP *parser, acmp_node_t *node)_ function.

Here is how to reproduce on Ubuntu 16.04/gcc (Ubuntu 5.4.0-6ubuntu1~16.04.4):
git clone https://github.com/SpiderLabs/ModSecurity.git
cd ModSecurity
git checkout origin/v3/master
git submodule init
git submodule update
./build.sh
./configure
MAKEFLAGS=-j4 make
cp modsecurity.conf-recommended examples/reading_logs_via_rule_message/modsec.conf
cd examples/reading_logs_via_rule_message/

Comment out in modsec.conf: _SecAuditEngine RelevantOnly_

git clone https://github.com/SpiderLabs/owasp-modsecurity-crs.git
cat owasp-modsecurity-crs/crs-setup.conf.example >> modsec.conf
echo 'Include "owasp-modsecurity-crs/rules/*.conf"' >> modsec.conf
./simple_request modsec.conf

There is the gdb stack back-trace:
Thread 3 "simple_request" received signal SIGABRT, Aborted.
[Switching to Thread 0x7fffee8fd700 (LWP 31472)]
0x00007ffff60a6428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007ffff60a6428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff60a802a in __GI_abort () at abort.c:89
#2  0x00007ffff60e87ea in __libc_message (do_abort=do_abort@entry=2, fmt=fmt@entry=0x7ffff6201e98 "*** Error in `%s': %s: 0x%s ***\n") at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007ffff60f137a in malloc_printerr (ar_ptr=<optimized out>, ptr=<optimized out>, str=0x7ffff6201f60 "double free or corruption (fasttop)", action=3) at malloc.c:5006
[...]

PR #3221, opened 2024-08-09
Add support to run unit tests in a multithreaded context

eduar-hte:
## what

The `unit_test` program has been updated to accept the `mtstress` argument to run operator/transformation tests in a multithreaded context.

## why

The goal is to detect if the operator/transformation fails in this context.

## misc

- In this mode, the test will be executed 5'000 times in 50 threads concurrently.
- Allocation & initialization of the operator/transformation is performed once in the main thread, while the evaluation is executed in the threads.
  - This is consistent with the library's support for multithreading, where initialization and loading of rules is expected to run once. See issue #3215.

## references

This work is related to #3215, to check whether the optional mutex on the [Pm operator](https://github.com/eduar-hte/ModSecurity/blob/119a6fc07482096e8429399dc2d7c0d3f903a7ae/src/operators/pm.h#L31) to access `acmp` trees is needed.

This was introduced in commit 119a6fc & 7d786b3 likely because of issue #1573. This option is *off* by default and it's not clear whether it's necessary, as stated  [here](https://github.com/owasp-modsecurity/ModSecurity/issues/1954#issuecomment-438725954).

### DIFF
--- a/test/common/modsecurity_test.cc
+++ b/test/common/modsecurity_test.cc
@@ -93,13 +93,13 @@ bool ModSecurityTest<T>::load_test_json(const std::string &file) {
 
 
 template <class T>
-std::pair<std::string, std::vector<T *>>*
+void
 ModSecurityTest<T>::load_tests(const std::string &path) {
     DIR *dir;
     struct dirent *ent;
     struct stat buffer;
 
-    if ((dir = opendir(path.c_str())) == NULL) {
+    if ((dir = opendir(path.c_str())) == nullptr) {
         /* if target is a file, use it as a single test. */
         if (stat(path.c_str(), &buffer) == 0) {
             if (load_test_json(path) == false) {
@@ -107,10 +107,10 @@ ModSecurityTest<T>::load_tests(const std::string &path) {
                 std::cout << std::endl;
             }
         }
-        return NULL;
+        return;
     }
 
-    while ((ent = readdir(dir)) != NULL) {
+    while ((ent = readdir(dir)) != nullptr) {
         std::string filename = ent->d_name;
         std::string json = ".json";
         if (filename.size() < json.size()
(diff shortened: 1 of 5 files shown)


eduar-hte:
All operator & transformation unit tests were executed using this feature and no issue was detected.

NOTE: This doesn't guarantee that there's no multithreading issue lurking in there, though.

airween (approved):

airween:
Thanks, @eduar-hte!

A reminder to me: 11 tests are failed, I have to check the reason.

eduar-hte:
> A reminder to me: 11 tests are failed, I have to check the reason.

I think you're referring to these unit tests: `test-cases/secrules-language-tests/transformations/phpArgsNames.json`.

I remember looking into this while working on the Windows port (PR #3132), because when I run `unit_tests` manually they'd fail (they're not part of the `make check` test suite (`test/test-suite.in`) and that's why the builds don't fail.)

These tests were added to in owasp-modsecurity/secrules-language-tests@a3d4405 and are related to PR #2387, which I understand is a v3.1 feature.

airween:
> I think you're referring to these unit tests: `test-cases/secrules-language-tests/transformations/phpArgsNames.json`.

yes, exactly.
 
> I remember looking into this while working on the Windows port (PR #3132), because when I run `unit_tests` manually they'd fail (they're not part of the `make check` test suite (`test/test-suite.in`) and that's why the builds don't fail.)

yes, I also wondered why this haven't appeared yet

> These tests were added to in [owasp-modsecurity/secrules-language-tests@a3d4405](https://github.com/owasp-modsecurity/secrules-language-tests/commit/a3d4405) and are related to PR #2387, which I understand is a v3.1 feature.

ah, thanks - I hadn't time to review the history, but yes, this explains the behavior. Seems like the tests had already added but the transformation not. I think we can remove this test file.

Thanks for helping discovery the situation.

eduar-hte:
> ah, thanks - I hadn't time to review the history, but yes, this explains the behavior. Seems like the tests had already added but the transformation not. I think we can remove this test file.

If you don't want to remove it, you could update the submodule reference in `test/test-cases/secrules-language-tests` to reference the previous commit (owasp-modsecurity/secrules-language-tests@d03f4c1), as the commit with the `phpArgsNames` tests is the latest commit in that repository.

eduar-hte:
> ah, thanks - I hadn't time to review the history, but yes, this explains the behavior. Seems like the tests had already added but the transformation not. I think we can remove this test file.

Submitted PR owasp-modsecurity/secrules-language-tests#10 to remove the tests associated to this missing transformation.

PR #3222, opened 2024-08-09
Remove several string copies and unnecessary heap allocations

eduar-hte:
## what

- Remove several unnecessary heap allocations and copies in the library.
- *Inline* string manipulation functions in `src/utils/string.*` by moving the definition of these functions into the header file.

## why

The unnecessary heap allocations and copies take time and contribute to heap fragmentation (which also reduces performance over time).

## misc

This is part of a series of PRs to improve performance of the library (3/n). Previous: #3218


airween (approved):
LGTM, thanks!

PR #3228, opened 2024-08-13
Replace usage of std::ctime, which is not safe in multithread contexts

eduar-hte:
## what

Replace usage of `std::ctime` in the string util [ascTime](https://github.com/owasp-modsecurity/ModSecurity/blob/718d121ee3c4384f01276f36a46bdbbae916c2b0/src/utils/string.h#L62), which is not safe in multithreaded contexts. `ascTime` is used by `Transaction::toJSON`, which is used to log transactions.

## why

From IBM's [ctime](https://www.ibm.com/docs/en/i/7.5?topic=functions-ctime-convert-time-character-string) documentation:

> The asctime() and ctime() functions, and other time functions can use a common, statically allocated buffer to hold the return string. Each call to one of these functions might destroy the result of the previous call. The asctime_r(), ctime_r(), gmtime_r(), and localtime_r() functions do not use a common, statically allocated buffer to hold the return string. These functions can be used in place of asctime(), ctime(), gmtime(), and localtime() if reentrancy is desired.

This was reported by `sonarcloud` while working on PR #3222:

> Remove use of this obsolete "ctime" function. Replace it by a call to "strftime".
> Obsolete POSIX functions should not be used cpp:S1911

## changes

- Replace usage of `std::ctime` with a call to `strftime`, where the output buffer is provided as an argument and is thread-safe.
- Reviewed and adjusted buffer size for others uses of `strftime` to leverage `std::size` to reserve the space needed by the format string.
- Simplified implementation of `TimeMon::evaluate` which was unnecessarily generating a string representation of the month to then parse it as an integer value, when the value is already available in the `struct tm timeinfo` with the call to `localtime_r`.

## notes

Checked that the library doesn't currently include any call to `localtime` (which is not thread-safe either) and uses `localtime_r` instead.

### DIFF
--- a/src/audit_log/writer/parallel.cc
+++ b/src/audit_log/writer/parallel.cc
@@ -51,28 +51,27 @@ Parallel::~Parallel() {
 
 inline std::string Parallel::logFilePath(time_t *t,
     int part) {
-    struct tm timeinfo;
-    char tstr[300];
-    std::string name("");
+    std::string name;
 
+    struct tm timeinfo;
     localtime_r(t, &timeinfo);
 
     if (part & YearMonthDayDirectory) {
-        memset(tstr, '\0', 300);
-        strftime(tstr, 299, "/%Y%m%d", &timeinfo);
-        name = tstr;
+        char tstr[std::size("/yyyymmdd")];
+        strftime(tstr, std::size(tstr), "/%Y%m%d", &timeinfo);
+        name.append(tstr);
     }
 
     if (part & YearMonthDayAndTimeDirectory) {
-        memset(tstr, '\0', 300);
-        strftime(tstr, 299, "/%Y%m%d-%H%M", &timeinfo);
-        name = name + tstr;
+        char tstr[std::size("/yyyymmdd-hhmm")];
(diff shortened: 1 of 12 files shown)


eduar-hte:
The sonarcloud reported issues recommend replacing the use of `strftime` with C++20's `std::chrono` library and `std::format`. 🙄

> Replace this use of "strftime" with equivalent code using "std::chrono".
> "std::chrono" components should be used to operate on time cpp:S6229
> (...)
> The chrono library, introduced in C++20, provides support for calendars, time zones, and i/o formatting and parsing operations on time-related objects.

airween (approved):
LGTM - thanks!

airween:
> The sonarcloud reported issues recommend replacing the use of `strftime` with C++20's `std::chrono` library and `std::format`. 🙄

I saw.

I updated SonarCloud settings, now it checks the code with C++17 standards. We will see the result...

eduar-hte:
> I updated SonarCloud settings, now it checks the code with C++17 standards. We will see the result...

That's great! I think that should reduce noise (that is, we won't need to dismiss issues because they don't apply, at least yet) Thx!

PR #3231, opened 2024-08-16
Remove unnecessary heap allocated copies in Transformation actions

eduar-hte:
## what

Remove unnecessary heap allocated copies in most `Transformation` classes.

## why

The current implementation of most `Transformation` classes are copying the `std::string` values in order to modify them and then return them. This can be optimized by updating the signature of `Transformation::evaluate` to receive a reference of the value to be modified, and if the transformation allows it, modify it in-place.

This change increases performance of the `benchmark` test by about 11%-12,5%.

## changes

- Updated transformations to work inplace on the `std::string` buffer of the value where the transformation is being applied, instead of copying to a temporary (heap-allocated) string.
- Updated `Transformation::evaluate` signature to receive the value by reference and perform the transformation inplace, if possible (some transformations cannot be performed in place because they need a bigger buffer to produce the new value, such as `base64` encoding).
  - Additionally, the signature of the method has been updated to return a bool value to indicate if the value has been changed. This removes an additional string comparison in `RuleWithActions::executeTransformation` to compare the original and transformed value further improving performance.
    - This required many of the transformations to detect when a change is produced in the input value to produce this in an efficient way (look for the `changed` bool flag in their implementation). Some transformations *always* transform their input value, so they just return `true`. Notably, a few of the transformations had implementations that would compute a `changed` flag, but that was currently unused.
- Made inplace transformations methods inline and moved them out of the class interface as almost all of them are not used outside of them.
- Simplified & shared implementation of several transformations, removing duplicate code.
- Removed unnecessarily included header files.

## references

While working on updating the configuration to build on C++17 (and C++20), I found out through this [comment](https://github.com/owasp-modsecurity/ModSecurity/pull/3079#issuecomment-1938638503) that a PR to make the transformations in-place was submitted before in PR #2368 and was integrated into the v3.1 branch (though curiously the changes are not currently included in that branch).

This PR goes further in the review of the different in-place transformations to simplify and make their implementations more consistent and efficient. Additionally, because of the signature change of `Transformation::evaluate` included in this PR, an allocation and copy is avoided.

## misc

This is part of a series of PRs to improve performance of the library (4/n). Previous: #3222


eduar-hte:
> This change increases performance of the `benchmark` test by about 11%-12,5%.

 * Mac mini (1st generation)
   * v3/master: 96.15 secs
   * This PR: 84.11 secs (12.52% improvement)
   * v3/master: 143,34 secs
   * This PR: 127,37 secs (11,14% improvement)

eduar-hte:
The unit tests for each transformation were executed (using a different number of iterations, as some take more time than others).

These results were obtained in a Debian (bullseye) container running on WSL on Windows Dev Kit 2023, and are provided as a reference.

(Times in milliseconds. Averages of five runs)

Transformation | Iterations | v3/master | This PR | Difference
-- | -- | -- | -- | --
base64decode | 1500000 | 2,404.83 | 2,111.14 | -12.21%
base64decodeExt | 1250000 | 2,129.00 | 1,886.29 | -11.40%
base64encode | 2000000 | 2,205.14 | 1,890.67 | -14.26%
cmdLine | 2000000 | 2,586.29 | 2,290.14 | -11.45%
compressWhitespace | 800000 | 2,291.43 | 1,875.29 | -18.16%
cssDecode | 1600000 | 2,155.43 | 1,931.00 | -10.41%
escapeSeqDecode | 500000 | 1,899.67 | 1,704.14 | -10.29%
hexDecode | 1750000 | 2,505.14 | 2,213.50 | -11.64%
hexEncode | 600000 | 933.00 | 924.29 | -0.93%
htmlEntityDecode | 1000000 | 3066,41 | 2400,63 | -21,71%
jsDecode | 600000 | 2,249.14 | 2,009.57 | -10.65%
lowercase | 1750000 | 2,295.33 | 2,040.57 | -11.10%
md5 | 1500000 | 3,321.80 | 3,270.86 | -1.53%
normalisePath | 250000 | 2,581.00 | 2,375.17 | -7.97%
normalisePathWin | 200000 | 2,217.86 | 2,004.67 | -9.61%
parityEven7bit | 1500000 | 2,259.00 | 2,003.00 | -11.33%
parityOdd7bit | 2000000 | 2,829.00 | 2,475.00 | -12.51%
parityZero7bit | 1750000 | 2,488.83 | 2,222.33 | -10.71%
removeComments | 300000 | 2,300.67 | 2,156.86 | -6.25%
removeCommentsChar | 300000 | 2,280.67 | 2,201.86 | -3.46%
removeNulls | 1000000 | 2,444.17 | 2,361.43 | -3.39%
removeWhitespace | 1250000 | 2,448.00 | 2,335.86 | -4.58%
replaceComments | 250000 | 2,007.57 | 1,913.57 | -4.68%
replaceNulls | 1250000 | 2,963.17 | 2,794.43 | -5.69%
sha1 | 2000000 | 3,141.14 | 3,067.50 | -2.34%
sqlHexDecode | 8000000 | 3,004.86 | 2,777.60 | -7.56%
trim | 750000 | 2,637.43 | 2,498.00 | -5.29%
trimLeft | 750000 | 2,528.14 | 2,387.43 | -5.57%
trimRight | 750000 | 2,598.33 | 2,454.29 | -5.54%
urlDecode | 300000 | 2,657.33 | 2,448.00 | -7.88%
urlDecodeUni | 150000 | 2,850.83 | 2,535.80 | -11.05%
urlEncode | 250000 | 2,469.86 | 2,378.00 | -3.72%
utf8toUnicode | 30000 | 4,194.29 | 4,061.29 | -3.17%

eduar-hte:
Regenerated PR to divide changes to make transformations in-place into multiple commits to simplify review. The original commit can be found in eduar-hte/ModSecurity@0a2dd1a.

NOTE: This required making the first commit in the PR include the changes to the signature of `Transformation::evaluate` and temporarily adjust each derived class for the code to still compile and work.

airween:
Hi @eduar-hte,

first of all, many thanks for this very impressive PR.

I was able to measure about a 10% performance increase with the modifications, which is amazing of course.

I am concerned about two things:
* Sonarcloud reports 50 new issues. I see those are very trivial issues (for eg. [this](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3231&id=owasp-modsecurity_ModSecurity&open=AZF2A9QV4s7qX3LZavzE) one or any other), but we have an intention to decrease and eliminate all (possible) issues sooner or later. If we are increasing this number then it will be harder to fix them later
* this PR contains many other modifications than restructuring of Transformation actions, for eg. the [transaction.cc](https://github.com/owasp-modsecurity/ModSecurity/pull/3231/files#diff-974db259ad96aebcb69f17e473d6dcde871731ca153572295e15b932073b3af4L351) is strictly unaffected. This is not a problem but it makes harder the review process. Also it will also be difficult to understand the intent and context of the amendments in the future.

So in the future please consider to split these modifications and send only the strictly affected ones. (But now you don't need to split them up.)

For the Sonarcloud: I don't know how to handle this now. As I explained above we want to decrease the number of issues, not increment :smile:. If you don't want or can't fix them, I'm going to do that later, but I think it's more easier to finish them now, than later.

What do you think?

eduar-hte:
> * this PR contains many other modifications than restructuring of Transformation actions, for eg. the [transaction.cc](https://github.com/owasp-modsecurity/ModSecurity/pull/3231/files#diff-974db259ad96aebcb69f17e473d6dcde871731ca153572295e15b932073b3af4L351) is strictly unaffected. This is not a problem but it makes harder the review process. Also it will also be difficult to understand the intent and context of the amendments in the future.

As I mentioned in #3233, this PR, PR #3222 & PR #3233 were originally on the same branch about removing unnecessary heap-allocations and copies (mostly of `std::string` objects). I ended up splitting that branch into these separate PRs for them to be more specific.

I'd argue that all the commits in this PR are specific to the `Transformation` related changes.

 * The first commit (commit 9d773cf)
   * Sets up the change to the signature of `Transformation::evaluate` that allows for in-place modification of the value.
   * This requires updating `RuleWithActions::executeTransformation`, which is where this method is called.
     * I opportunistically updated the signature of this method to receive `const` paramters (and, if possible, references), which is aligned with some of the issues reported by a newer version of `cppcheck` (which I've been running on the codebase).
     * I also moved the declaration of this function to the `private` section because it was not used outside this class.
   * As I was going through all the headers in `src/actions/transformations`, I removed unnecessarily included header files (to improve build/rebuild times) and simplified the namespace definition. For the sake of consistency, I took the time to go and to the same for the rest of the `actions`, which goes beyond the scope of this PR. I wonder whether I'd have submitted a separate PR with these changes otherwise.
 * The rest of the commits are pretty much limited to updating, if possible, each `Transformation` to make them in-place.
   * There are a couple of opportunistic changes to refactor and remove code duplication in transformations that were clearly copy & pasted and where integrating the implementation favors their maintenance.

PS: The change in `transaction.cc` is necessary because `utils::urldecode_nonstrict_inplace` was updated in the context of the in-place transformation updates (see `src/actions/transformations/url_decode.cc`), so this other call had to be updated as well.

eduar-hte:
> * Sonarcloud reports 50 new issues. I see those are very trivial issues (for eg. [this](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3231&id=owasp-modsecurity_ModSecurity&open=AZF2A9QV4s7qX3LZavzE) one or any other), but we have an intention to decrease and eliminate all (possible) issues sooner or later. If we are increasing this number then it will be harder to fix them later

I actually did take some time and did multiple iterations to address issues reported by Sonarcloud (see the multiple `force-pushed` entries in the PR history).

After addressing many of the things reported by Sonarcloud, I think (almost) all the current ones exist in the current codebase, and are only flagged now because the PR updates some of the code.

The following are my notes on the currently reported issues. Half of the issues (24) are related to the first three categories.

 * `Add a using-declaration to this derived class to inherit the constructors of "Action", and remove the ones you manually duplicated. Note that this may add other constructors to your derived class.` (13)
   * This is a nice suggestion, and I applied it in *all* the `Transformation` classes. This can be seen with the line `using Transformation::Transformation;` to simplify the definition of each `Transformation` derived class, instead of having each class write a constructor such as:
```explicit RemoveNulls(const std::string &action) : Transformation(action) { }```
   * However, I don't think that we should address the rest of the reported issues because they introduce additional constructors in the derived classes, which is something the issue itself mentions (see `Note that this may add other constructors to your derived class`). This could be applied to `Transformation` derived classes, because `Transformation` has a single constructor with the same signature that was used in all derived classes, but it's not true of `Action`.
   * NOTE: It could be argued that something like a `cppcheck` inline suppression should be used here, though even that would pollute most of the rest `Action` derived classes because of an inaccurate Sonarcloud report. ¯\_(ツ)_/¯
 * `Refactor this function to reduce its Cognitive Complexity from 59 to the 25 allowed.` (7)
   * There are a few of these reported in this PR, and all of them are related to how some of the transformation algorithms are (currently) implemented.
   * I don't think we should try and address them, with the risk of introducing bugs, just for the sake of removing Sonarcloud issues (I'd understand and favor keeping this bar high for new code, though)
   * Even though I updated most of the transformations in the context of this PR, I tried to minimize and limit my changes to the transformations' algorithms to how the buffers were handled and referenced, without changing the existing/functioning algorithm because of this reason (especially in large/complex algorithms that could not be trivially rewritten/simplified).
 * `Refactor this code to not nest more than 3 if|for|do|while|switch statements.` (3)
   * The same as above.
 * `Reduce the number of nested "goto" statements from 5 to 1 authorized.` (1)
   * The same as above.
 * `Use the init-statement to declare "t" inside the if statement.` (2)
   * This is reported in `Action::set_name_and_payload`, where I replaced an unnecessary heap-allocated `std::string` everytime the function is executed for a constant string (`t:`). I looked at the report and even though I like the feature of declaring variables in `if` statements, I don't think it's idiomatic (and readable) to replace this:
```cpp
        const char t[] = "t:";

        if (data.compare(0, std::size(t) - 1, t) == 0) {
```
with:
```cpp
        if (const char t[] = "t:"; data.compare(0, std::size(t) - 1, t) == 0) {
```
 * `Use "std::string" instead of a C-style char array.` (4)
   * Sonarcloud reports this a couple of times in `src/actions/transformations/html_entity_decode.cc` where using a stack allocated fixed-size buffer is clearly a better option than creating a heap-allocated `std::string`.
 * `Replace "reinterpret_cast" with a safer operation` (2)
   * Sonarcloud reports this a couple of times where I introduced an explicit `reinterpret_cast` instead of a naked cast in the current codebase to cast `char` to `unsigned char`, which is something most of the `Transformation` algorithms do.
 * `Use "std::byte" for byte-oriented data manipulation.` (7)
   * I briefly looked into this and I'm not sure the issue is right in the suggestion, as you can see the `parity` algorithms actually extend the `unsigned char` byte to an `unsigned int` in order to implement its logic.
 * `Modify the macro definition so that it needs to be followed by a semicolon, or remove this empty statement.`
   * This is reported in one of the many uses of the macro `ms_dbg_a` in the codebase.
 * `Replace this use of "push_back" with "emplace_back".` (3)
   * These two are present in the current codebase, I'll try and address them.
 * `Either add a parameter list or the "&" operator to this use of "mbedtls_md5".` (2)
   * This is reported a couple of times when using the function name to instantiate a template introduced to remove code duplication for the implementation of the md5 & sha1. I'll take a look into whether I can address them.
 * `Define each identifier in a dedicated statement.` (5)
   * These are reported on multiple variables used to implement some of the transformation algorithms in the current codebase. I think introducing each of them in a separate line unnecessarily increases the function length just for the sake of keeping Sonarcloud happy.

airween:
>     * The first commit (commit [9d773cf](https://github.com/owasp-modsecurity/ModSecurity/commit/9d773cfbc8cee5ce9f76738c160caa550430095b))

that's okay.

> PS: The change in `transaction.cc` is necessary because `utils::urldecode_nonstrict_inplace` was updated in the context of the in-place transformation updates (see `src/actions/transformations/url_decode.cc`), so this other call had to be updated as well.

ah, you're right, I get it wrong, sorry.

eduar-hte:
> * `Replace this use of "push_back" with "emplace_back".` (3)
> * `Either add a parameter list or the "&" operator to this use of "mbedtls_md5".` (2)

I've just pushed an update (integrating these changes into the corresponding existing commits) to address these.

NOTE: I'm updating the comment to reflect that Sonarcloud quickly updated its findings, but still reports issues in code that is no longer there (!). This happened to me a couple of times last week while I was in the process of addressing other issues. It seems that the scan is sometimes done in *stale* code, so it's not accurate. I had to wait for a while and then generate a new push for the scan to be accurate. If you could re-run the scan in an hour (or two), I think we should see the 5 issues above go away.

airween:
> After addressing many of the things reported by Sonarcloud, I think (almost) all the current ones exist in the current codebase, and are only flagged now because the PR updates some of the code.

Yes, I knew that - I just asked you about this to know what do you think in long term.

Thank you for all explanations - we discussed the Sonarcloud issues with @gberkes and we agreed that we have to review all issues (not just these ones) later, and fix them by category.

Anyway, these summaries are awesome here and everyone can see that these are not new issues. Thanks again, and thanks for the great PR.

I'm going to merge this now.

airween (approved):

eduar-hte:
Submitted PR owasp-modsecurity/secrules-language-tests#10 to add more tests for the `sqlHexDecode` transformation updated in this PR.

PR #3239, opened 2024-08-25
fix: Handle "filename*" field in MP header

airween:
## what

Add handling of Multipart Header's "filename*" (asterisk at the end!) field.

## why

mod_security2 (v2) does not handle MULTIPART header's "filename*" field, eg:

```
Content-Disposition: form-data; name="file"; filename*=UTF-8''r%C3%A9sum%C3%A9.pdf
```
where the filename is UTF8 encoded string with value "résumé.pdf".

## references

[RFC 7578](https://www.rfc-editor.org/rfc/rfc7578.html)

## additional notes

v3 handles as well this header, I almost copied that [part](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/request_body_processor/multipart.cc#L342-L375) of [code](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/request_body_processor/multipart.cc#L463-L469).

Thanks @fzipi for bringing it to our attention.

### DIFF
--- a/apache2/msc_multipart.c
+++ b/apache2/msc_multipart.c
@@ -20,6 +20,9 @@
 #include "msc_util.h"
 #include "msc_parsers.h"
 
+static const char* mime_charset_special = "!#$%&+-^_`{}~";
+static const char* attr_char_special = "!#$&+-.^_`~";
+
 void validate_quotes(modsec_rec *msr, char *data, char quote)  {
     assert(msr != NULL);
     int i, len;
@@ -85,6 +88,7 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
     assert(msr != NULL);
     assert(c_d_value != NULL);
     char *p = NULL, *t = NULL;
+    char *filenameStar = NULL;
 
     /* accept only what we understand */
     if (strncmp(c_d_value, "form-data", 9) != 0) {
@@ -130,7 +134,41 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
          */
 
         char quote = '\0';
-        if ((*p == '"') || (*p == '\'')) {
+        if (name == "filename*") {
+            /* filename*=charset'[optional-language]'filename */
+            /* Read beyond the charset and the optional language*/
+            const char* start_of_charset = p;
+            while ((*p != '\0') && (isalnum(*p) || (strchr(mime_charset_special, *p)))) {
+                p++;
+            }
+            if ((*p != '\'') || (p == start_of_charset)) {
+                return -16; // Must be at least one legit char before ' for start of language
+            }
+            p++;
+            while ((*p != '\0') && (*p != '\'')) {
+                p++;
+            }
+            if (*p != '\'') {
+                return -17; // Single quote for end-of-language not found
+            }
+            p++;
+
+            /* Now read what should be the actual filename */
+            const char* start_of_filename = p;
+            while ((*p != '\0') && (*p != ';')) {
+                if (*p == '%') {
+                    if ((*(p+1) == '\0') || (!isxdigit(*(p+1))) || (!isxdigit(*(p+2)))) {
+                        return -18;
+                    }
+                    p += 3;
+                } else if (isalnum(*p) || strchr(attr_char_special, *p)) {
+                    p++;
+                } else {
+                    return -19;
+                }
+            }
+            value = apr_pmemdup(msr->mp, start_of_filename, p - start_of_filename);
+        } else if ((*p == '"') || (*p == '\'')) {
             /* quoted */
             quote = *p; // remember which quote character was used for the value
 
@@ -205,8 +243,7 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
                     log_escape_nq(msr->mp, value));
             }
         }
-        else
-        if (strcmp(name, "filename") == 0) {
+        else if (strcmp(name, "filename") == 0) {
 
             validate_quotes(msr, value, quote);
 
@@ -221,6 +258,17 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
 
             if (msr->txcfg->debuglog_level >= 9) {
                 msr_log(msr, 9, "Multipart: Content-Disposition filename: %s",
+                    log_escape_nq(msr->mp, value));
+            }
+        } else if (strcmp(name, "filename*") == 0) {
+            if (filenameStar != NULL && strlen(filenameStar) != 0) {
+                msr_log(msr, 4,
+                    "Multipart: Warning: Duplicate Content-Disposition filename*: %s.",  log_escape_nq(msr->mp, value));
+                return -20;
+            }
+            filenameStar = apr_pstrdup(msr->mp, value);
+            if (msr->txcfg->debuglog_level >= 9) {
+                msr_log(msr, 9, "Multipart: Content-Disposition filename*: %s.",
                     log_escape_nq(msr->mp, value));
             }
         }
--- a/apache2/msc_multipart.c
+++ b/apache2/msc_multipart.c
@@ -134,7 +134,7 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
          */
 
         char quote = '\0';
-        if (name == "filename*") {
+        if (strcmp(name, "filename*") == 0) {
             /* filename*=charset'[optional-language]'filename */
             /* Read beyond the charset and the optional language*/
             const char* start_of_charset = p;
--- a/apache2/msc_multipart.c
+++ b/apache2/msc_multipart.c
@@ -149,6 +149,7 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
                 p++;
             }
             if (*p != '\'') {
+                msr->mpd->flag_invalid_quoting = 1;
                 return -17; // Single quote for end-of-language not found
             }
             p++;


apache2/msc_multipart.c
```diff
@@ -130,7 +134,41 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
          */
 
         char quote = '\0';
-        if ((*p == '"') || (*p == '\'')) {
+        if (name == "filename*") {
```
fzipi:
```suggestion
        if (strcmp(name, "filename*") == 0) {
```
This is the way it is used in the other cases, so I guess it makes sense to maintain the same format. Another (simpler) option would be to start using `switch` instead.

apache2/msc_multipart.c
```diff
@@ -223,6 +260,17 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
                 msr_log(msr, 9, "Multipart: Content-Disposition filename: %s",
                     log_escape_nq(msr->mp, value));
             }
+        } else if (strcmp(name, "filename*") == 0) {
```
fzipi, line 264: Why is this comparison being used twice?

apache2/msc_multipart.c
```diff
@@ -130,7 +134,41 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
...
+            }
+            if ((*p != '\'') || (p == start_of_charset)) {
+                return -16; // Must be at least one legit char before ' for start of language
+            }
+            p++;
+            while ((*p != '\0') && (*p != '\'')) {
+                p++;
+            }
+            if (*p != '\'') {
+                return -17; // Single quote for end-of-language not found
```
theseion, line 153: Shouldn't you use something like this here, as below?
```
msr->mpd->flag_invalid_quoting = 1;
```

apache2/msc_multipart.c
```diff
@@ -130,7 +134,41 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
...
+            }
+            p++;
+
+            /* Now read what should be the actual filename */
+            const char* start_of_filename = p;
```
theseion, line 158: Looking at the parsing of `filename` below, I expect the corresponding logic here to include the same use of flags and logging, e.g., `msr_log(msr, 4, "Multipart: Warning: Duplicate Content-Disposition name: %s",`.

apache2/msc_multipart.c
theseion: Agreed. This stuff needs to go in the first `filename*` branch.

theseion (changes requested):

marcstern:
https://www.rfc-editor.org/rfc/rfc5987#section-3.2
This RFC says that only UTF-8 & ISO-8859-1 are standardized:

> Producers MUST use either the "UTF-8" ([RFC3629]) or the "ISO-8859-1" ([ISO-8859-1]) character set.  Extension character sets (mime-charset) are reserved for future use.

We're restricting for years the languages to these two and never found a false positive.
Should we allow all of them?
If the code allows it, we should add a new collection (FILES_LANG?) allowing checking this syntax.

apache2/msc_multipart.c
```diff
@@ -130,7 +134,41 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
...
+            /* Read beyond the charset and the optional language*/
+            const char* start_of_charset = p;
+            while ((*p != '\0') && (isalnum(*p) || (strchr(mime_charset_special, *p)))) {
+                p++;
+            }
+            if ((*p != '\'') || (p == start_of_charset)) {
+                return -16; // Must be at least one legit char before ' for start of language
+            }
+            p++;
+            while ((*p != '\0') && (*p != '\'')) {
```
marcstern, line 148: Languages can only contain alphanum & '-'
`while (isalnum(*p) || *p == '-') p++;`

apache2/msc_multipart.c
```diff
@@ -223,6 +260,17 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
                 msr_log(msr, 9, "Multipart: Content-Disposition filename: %s",
                     log_escape_nq(msr->mp, value));
             }
+        } else if (strcmp(name, "filename*") == 0) {
+            if (filenameStar != NULL && strlen(filenameStar) != 0) {
```
marcstern, line 265: Should we allow "filename*=; ... ;filename*=UTF-8''fname"?
I don't think so.
We don't allow it for filename.

apache2/msc_multipart.c
```diff
@@ -20,6 +20,9 @@
 #include "msc_util.h"
 #include "msc_parsers.h"
 
+static const char* mime_charset_special = "!#$%&+-^_`{}~";
```
marcstern, line 23: In case we allow something else than UTF-8 & ISO-8859-1, these characters look extremely permissive.
The standard allows them obviously, but did you ever encounter such a charset?

apache2/msc_multipart.c
airween: Thanks for letting this know. It's my fault, in C language we can't use direct comparison in case of strings. I tried to replace all occurrences but I forget this one - thanks again.

Also in case of C language, we can't use `switch`-`case` structure for strings. The reason is that `case` expects a constant, not an array, and strings in C are arrays.

I replaced this line in 524a799.

apache2/msc_multipart.c
airween: If we move this stuff into the first block then this block will be empty. But I'm afraid then the logic will change too much. We **MUST** check the three cases there, namely the variable `name` value is `name`, `filename` or `filename*`.

I think we should leave now as it is.

apache2/msc_multipart.c
airween: Do you think about [this](https://github.com/owasp-modsecurity/ModSecurity/blob/524a7996b010ebae64f51b1c5d36dbb150c5ef08/apache2/msc_multipart.c#L266) line? (Line 266. in current implementation):

```C
                msr_log(msr, 4,
                    "Multipart: Warning: Duplicate Content-Disposition filename*: %s.",  log_escape_nq(msr->mp, value));
                return -20;
```

apache2/msc_multipart.c
airween: It's a good idea - thanks.

@fzipi, @marcstern - what do you think about this guys?

I already added this in 2b22261, but I can remove that.

If we keep this feature, we should add that to v3 too.

airween:
> https://www.rfc-editor.org/rfc/rfc5987#section-3.2 This RFC says that only UTF-8 & ISO-8859-1 are standardized:
> 
> > Producers MUST use either the "UTF-8" ([RFC3629]) or the "ISO-8859-1" ([ISO-8859-1]) character set.  Extension character sets (mime-charset) are reserved for future use.
> 
> We're restricting for years the languages to these two and never found a false positive. Should we allow all of them? If the code allows it, we should add a new collection (FILES_LANG?) allowing checking this syntax.

Well, you are right, but as it stands in RFC: _"Extension character sets (mime-charset) are reserved for future use."_ - if someone introduce a new feature in the future we should align the code again.

Introducing a new settings would be confused (in this case).

And I think first we should investigate is it possible to check those languages with rules.

apache2/msc_multipart.c
airween: You're right, and here - as you suggested - we can control this.

The question is: is it allowed to send the request without charset? I mean:

```
Content-Disposition: form-data; name="post"; filename*=resume.pdf
```

I can't find any relevant information about that.

apache2/msc_multipart.c
airween: I'm not sure I understand this question, because actually we don't allow this.

```
Content-Disposition: form-data; name="file"; filename*=; filename*=UTF-8''r%C3%A9sum%C3%A9.pdf
```

and I get:
```
ModSecurity: Multipart parsing error: Multipart: Invalid Content-Disposition header (-16): form-data; name="file"; filename*=; filename*=UTF-8''r%C3%A9sum%C3%A9.pdf. [hostname "localhost"] [uri "/post"] [unique_id "ZsyOgg3BHX5k_DYroV3swwAAAAA"]
ModSecurity: Warning. Match of "eq 0" against "REQBODY_ERROR" required. [file "/etc/modsecurity/modsecurity.conf"] [line "72"] [id "200002"] [msg "Failed to parse request body."] [data "Multipart parsing error: Multipart: Invalid Content-Disposition header (-16): form-data; name=\\x22file\\x22; filename*=; filename*=UTF-8''r%C3%A9sum%C3%A9.pdf."] [severity "CRITICAL"] [hostname "localhost"] [uri "/post"] [unique_id "ZsyOgg3BHX5k_DYroV3swwAAAAA"]
```

apache2/msc_multipart.c
airween: That's a good question we want to handle the (currently) allowed charset (as you wrote) or we permit anything.

Moreover as I wrote [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3239#issuecomment-2309893202) we should examine the possibility of applying the rules.

apache2/msc_multipart.c
theseion: https://www.rfc-editor.org/rfc/rfc7578#section-4.2 also says:

>    Some commonly deployed systems use multipart/form-data with file
   names directly encoded including octets outside the US-ASCII range.
   The encoding used for the file names is typically UTF-8, although
   HTML forms will use the charset associated with the form.

Which sounds to me that any charset is valid, regardless of what the other RFC says.

`charset` is required, AFAICT, only `language` is optional.

apache2/msc_multipart.c
theseion: Right. I didn't see that there are _two_ if/else structures. It looked to me like the last branch of the same if/else.

apache2/msc_multipart.c
theseion: Never mind. I was confused here as well about the two if/else.

apache2/msc_multipart.c
airween: So, do you think we do not need to check the charset and we should allow anything? Or restrict the charset content only to alnum chars (+ `-`)?

theseion: IMO, the charset should be restricted to the ABNF from the RFC.

apache2/msc_multipart.c
```diff
@@ -130,7 +134,42 @@ static int multipart_parse_content_disposition(modsec_rec *msr, char *c_d_value)
...
+                return -16; // Must be at least one legit char before ' for start of language
+            }
+            p++;
+            while ((*p != '\0') && (*p != '\'')) {
+                p++;
+            }
+            if (*p != '\'') {
+                msr->mpd->flag_invalid_quoting = 1;
+                return -17; // Single quote for end-of-language not found
+            }
```
fzipi, line 154: Picking this up from the discussion at https://github.com/owasp-modsecurity/ModSecurity/pull/3239#discussion_r1730758250 and https://github.com/owasp-modsecurity/ModSecurity/pull/3239#discussion_r1732154153 (@marcstern, @theseion) — the charset segment above already restricts to the RFC 5987 `mime-charsetc` ABNF via `mime_charset_special`, but the language segment right after it has no character validation at all, it just scans to the next `'`. Restricting it to `isalnum`/`-` (RFC 5646 language subtags) as @marcstern originally suggested closes that gap and matches "restricted to the ABNF" for the whole `charset'language'` prefix, not just the charset part:

```suggestion
            p++;
            while ((*p != '\0') && (*p != '\'')) {
                if (!(isalnum(*p) || (*p == '-'))) {
                    msr->mpd->flag_invalid_quoting = 1;
                    return -21; // Invalid character in language tag
                }
                p++;
            }
            if (*p != '\'') {
                msr->mpd->flag_invalid_quoting = 1;
                return -17; // Single quote for end-of-language not found
            }
            p++;
```

(`-21` is the next free negative return code in this function — `-1`..`-20` are all already used.)

Separately, re: "is charset optional" (your question above) — it isn't per the `ext-value` ABNF (`charset "'" [language] "'" value-chars`, charset is mandatory), and the current code already enforces that correctly (line 144 rejects when no `'` is found after zero-or-more charset chars), so no change needed there.

apache2/msc_multipart.c
fzipi: +1 on keeping this — matches how the other quoting/format issues in this function already set `flag_invalid_quoting` for callers to act on via rules, and dropping it would be inconsistent with that pattern. Since v3's equivalent parser doesn't set it in this branch, might be worth a follow-up issue on v3/master for parity rather than blocking this PR on it.

apache2/msc_multipart.c
airween: > Picking this up from the discussion at [#3239 (comment)](https://github.com/owasp-modsecurity/ModSecurity/pull/3239#discussion_r1730758250) and [#3239 (comment)](https://github.com/owasp-modsecurity/ModSecurity/pull/3239#discussion_r1732154153) (@marcstern, @theseion) — the charset segment above already restricts to the RFC 5987 `mime-charsetc` ABNF via `mime_charset_special`

Unfortunately the `mime-charsetc` describes the possible charsets this way:
```
charset       = "UTF-8" / "ISO-8859-1" / mime-charset

     mime-charset  = 1*mime-charsetc
     mime-charsetc = ALPHA / DIGIT
                   / "!" / "#" / "$" / "%" / "&"
                   / "+" / "-" / "^" / "_" / "`"
                   / "{" / "}" / "~"
                   ; as <mime-charset> in [Section 2.3 of [RFC2978]](https://datatracker.ietf.org/doc/html/rfc2978#section-2.3)
                   ; except that the single quote is not included
                   ; SHOULD be registered in the IANA charset registry
```

See the comments in last three lines.

Here are the registered IANA charsets:

https://www.iana.org/assignments/character-sets/character-sets.xhtml

> but the language segment right after it has no character validation at all, it just scans to the next `'`. Restricting it to `isalnum`/`-` (RFC 5646 language subtags) as @marcstern originally suggested closes that gap and matches "restricted to the ABNF" for the whole `charset'language'` prefix, not just the charset part:

yes, processing `language` (which is optional) is more easy then the `charset`.

What if we introduce a new configuration directive, for eg. `SecMultipartAllowedCharsets`?

Eg:

```
SecMultipartAllowedCharsets utf-8 iso-8859-1
```
and this would be the default (if someone do not want to set it explicit).

theseion: Sounds reasonable.

PR #3243, opened 2024-08-28
Adjust reference to modsecurity::utils::string::VALID_HEX

eduar-hte:
## what

Fix build after integrating PR #3231 & #3233 because of reference to `modsecurity::utils::string::VALID_HEX` in `pm.cc`'s `parse_pm_content`.

## why

These two PRs were independent of each other, but when we added a reference to `VALID_HEX` to simplify the code, we introduced an incompatibility between both PRs because PR #3231 moved `VALID_HEX` into the namespace `modsecurity::utils::string` (alongside all other helper functions in `src/utils/string.h`) and adjusted all call sites to this function (previously a macro). However, in that branch the call added to `parse_pm_content` that we introduced later in PR #3233 was not present and thus was not adjusted.

### DIFF
--- a/src/operators/pm.cc
+++ b/src/operators/pm.cc
@@ -22,6 +22,7 @@
 #include "src/utils/acmp.h"
 #include "src/utils/string.h"
 
+using namespace modsecurity::utils::string;
 
 static inline std::string parse_pm_content(const std::string &op_parm) {
     auto offset = op_parm.find_first_not_of(" \t");
--- a/test/test-cases/regression/operator-pm.json
+++ b/test/test-cases/regression/operator-pm.json
@@ -3,7 +3,7 @@
     "enabled": 1,
     "version_min": 300000,
     "version_max": 0,
-    "title": "pm operator test 1/4",
+    "title": "pm operator test 1/6",
     "client": {
       "ip": "200.249.12.31",
       "port": 2313
@@ -77,7 +77,7 @@
     "enabled": 1,
     "version_min": 300000,
     "version_max": 0,
-    "title": "pm operater test 3/4",
+    "title": "pm operater test 3/6",
     "client": {
       "ip": "200.249.12.31",
       "port": 2313
@@ -114,7 +114,7 @@
     "enabled": 1,
     "version_min": 300000,
     "version_max": 0,
-    "title": "pm operater test 4/4",
+    "title": "pm operater test 4/6",
     "client": {
       "ip": "200.249.12.31",
       "port": 2313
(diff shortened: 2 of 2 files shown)


airween:
Ahm, thanks - meanwhile I also started to work on this [fix](https://github.com/airween/ModSecurity/tree/v3/fixvalidhex) :smile:

Btw I don't understand the situation, because the PR's check were [successful](https://github.com/owasp-modsecurity/ModSecurity/pull/3233/checks), but later I get reports about failed [checks](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10596803612/job/29365572574).

Seems like now everything is fine.

I also added two new [tests](https://github.com/airween/ModSecurity/commit/019d4194e058d75f8e5e0f8baf752c9d51379bc8) (a positive and a negative) - do you think that would be useful to check the `@pm`'s correct behavior? If yes, could you pick up those modifications?

eduar-hte:
> Btw I don't understand the situation, because the PR's check were [successful](https://github.com/owasp-modsecurity/ModSecurity/pull/3233/checks), but later I get reports about failed [checks](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/10596803612/job/29365572574).

`VALID_HEX` used to live in the global namespace because it was a macro (`#define`), and in PR #3231 I converted it to a function (in order to continue removing preprocessor dependencies), and the function was included in the namespace `modsecurity::utils::string` (alongside the rest of the functions in `src/utils/string.h`, which is where `VALID_HEX` was defined).

You can see in commit a6d64bf that I adjusted the source files that were using this macro because the macro/function was no longer in the global namespace.

When we introduced the use of `VALID_HEX` in PR #3233 to simplify the code in `parse_pm_content`, a new reference to this macro/function was introduced, which needed to be adjusted to the namespace change.

> I also added two new [tests](https://github.com/airween/ModSecurity/commit/019d4194e058d75f8e5e0f8baf752c9d51379bc8) (a positive and a negative) - do you think that would be useful to check the `@pm`'s correct behavior? If yes, could you pick up those modifications?

ok, I'll do that and push an update to this PR.

eduar-hte:
> ok, I'll do that and push an update to this PR.

done. I added your repository as a `remote` in my local repository in order to `cherry-pick` your commit and preserve your authorship. :-)

airween:
> in PR #3231 I converted it to a function

Oh, thanks - I didn't noticed that. I looked up for that and found as macro so that was confused.

Never mind, thanks for all again. If the tests will passed I'm going to merge this PR.

airween (approved):

airween:
~Hmm... after this PR was merged, I receive notifications about failed tests, especially on macOS, with `SecRemoteRules` - like [this](https://github.com/airween/ModSecurity/actions/runs/10598379516/job/29370831532#step:8:74) one.~

Never mind, may be I forget to update that branch where I get from the notifications.

PR #3248, opened 2024-09-03
Simplified constructors, copy constructors & assignment operators

eduar-hte:
## what

Several constructors, copy constructors & assignment operators in the codebase can be simplified or even removed because they're not used.

## why

Simplify the codebase making it easier to maintain.

NOTE: In the case of `RuleMessage`, it addresses one of the `FIXME` in the codebase, see [here](https://github.com/owasp-modsecurity/ModSecurity/blob/1a8c96a1cd5e58e00401ad392ab1344104afd283/headers/modsecurity/rule_message.h#L47)

## misc

This is part of a series of PRs to improve performance of the library (5/n). Previous: #3231

### DIFF
--- a/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
+++ b/examples/reading_logs_via_rule_message/reading_logs_via_rule_message.h
@@ -155,8 +155,8 @@ class ReadingLogsViaRuleMessage {
         const modsecurity::RuleMessage *ruleMessage = \
             reinterpret_cast<const modsecurity::RuleMessage *>(ruleMessagev);
 
-        std::cout << "Rule Id: " << std::to_string(ruleMessage->m_ruleId);
-        std::cout << " phase: " << std::to_string(ruleMessage->m_phase);
+        std::cout << "Rule Id: " << std::to_string(ruleMessage->m_rule.m_ruleId);
+        std::cout << " phase: " << std::to_string(ruleMessage->getPhase());
         std::cout << std::endl;
         if (ruleMessage->m_isDisruptive) {
             std::cout << " * Disruptive action: ";
--- a/examples/using_bodies_in_chunks/simple_request.cc
+++ b/examples/using_bodies_in_chunks/simple_request.cc
@@ -76,8 +76,8 @@ static void logCb(void *data, const void *ruleMessagev) {
     const modsecurity::RuleMessage *ruleMessage = \
         reinterpret_cast<const modsecurity::RuleMessage *>(ruleMessagev);
 
-    std::cout << "Rule Id: " << std::to_string(ruleMessage->m_ruleId);
-    std::cout << " phase: " << std::to_string(ruleMessage->m_phase);
+    std::cout << "Rule Id: " << std::to_string(ruleMessage->m_rule.m_ruleId);
+    std::cout << " phase: " << std::to_string(ruleMessage->getPhase());
     std::cout << std::endl;
     if (ruleMessage->m_isDisruptive) {
         std::cout << " * Disruptive action: ";
--- a/headers/modsecurity/actions/action.h
+++ b/headers/modsecurity/actions/action.h
@@ -81,21 +81,9 @@ class Action {
             set_name_and_payload(_action);
         }
 
-    Action(const Action &a)
-        : m_isNone(a.m_isNone),
-        temporaryAction(a.temporaryAction),
-        action_kind(a.action_kind),
-        m_name(a.m_name),
-        m_parser_payload(a.m_parser_payload) { }
+    Action(const Action &a) = delete;
 
-    Action &operator=(const Action& a) {
-        m_isNone = a.m_isNone;
-        temporaryAction = a.temporaryAction;
-        action_kind = a.action_kind;
-        m_name = a.m_name;
-        m_parser_payload = a.m_parser_payload;
-        return *this;
-    }
+    Action &operator=(const Action& a) = delete;
 
     virtual ~Action() { }
 
--- a/headers/modsecurity/rule.h
+++ b/headers/modsecurity/rule.h
@@ -64,8 +64,8 @@ using MatchActions = std::vector<actions::Action *>;
 
 class Rule {
  public:
-    Rule(std::unique_ptr<std::string> fileName, int lineNumber)
-        : m_fileName(*fileName),
+    Rule(const std::string &fileName, int lineNumber)
+        : m_fileName(fileName),
         m_lineNumber(lineNumber),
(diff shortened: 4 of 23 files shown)


eduar-hte:
> Failed conditions  [8.8% Duplication on New Code](https://sonarcloud.io/component_measures?id=owasp-modsecurity_ModSecurity&pullRequest=3248&metric=new_duplicated_lines_density&view=list) (required ≤ 3%)

The tool is reporting duplicate code on changes to data the initialization of data members that were simplified (in `src/transaction.cc`).

The other two files mentioned are `examples` that use the same code to log rules (this PR only adjusts the existing lines to update how the data is retrieved). In library code (let's say, `src` dir) I'd obviously favor refactoring the code, but in the case of examples, it makes sense that they're independent to make it easier to use them as a simple reference and starting point for library users.

This PR is actually about *removing* code, but Sonarcloud will Sonarcloud...😊

eduar-hte:
> The tool is reporting duplicate code on changes to data the initialization of data members that were simplified (in `src/transaction.cc`).

I didn't look into updating the `Transaction` constructors in this PR, because these changes originated in those to simplify `RuleMessage`. These commits were part of another branch where use of `RuleMessage` is simplified, but I wanted to send them in a different PR to limit the scope of changes, as we've previously discussed.

PS: I'll take a look at the `Transaction` constructors and see if they can be simplified as well. They actually look identical, other than the `id` parameter, so they could be refactor to use a [delegating constructor](https://www.ibm.com/docs/en/zos/2.4.0?topic=only-delegating-constructors-c11).

eduar-hte:
> PS: I'll take a look at the `Transaction` constructors and see if they can be simplified as well. They actually look identical, other than the `id` parameter, so they could be refactor to use a [delegating constructor](https://www.ibm.com/docs/en/zos/2.4.0?topic=only-delegating-constructors-c11).

Done

PS: Notice that Sonarcloud once again run on stale code after the push, and is still showing the old version of `transaction.cc` and thus still flags duplicate code (in code that no longer exists). ¯\_(ツ)_/¯

eduar-hte:
> Issues  [17 New issues](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3248&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true)  [0 Accepted issues](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3248&issueStatuses=ACCEPTED)

Edited commits to address the following issues reported by Sonarcloud:

 * `headers/modsecurity/rule_marker.h`
   * [Merge this "if" statement with the enclosing one](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2JMH3hiFLGRloh9&id=owasp-modsecurity_ModSecurity)
     * Existing code, not introduced in this PR.
 * `headers/modsecurity/rule_message.h`
   * [Do not use the constructor's initializer list for data member "m_severity". Use the in-class initializer instead](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2G5H3hiFLGRloh5&id=owasp-modsecurity_ModSecurity) (8)
     * Existing code, not introduced in this PR.
 * `headers/modsecurity/rule_unconditional.h`
   * [Add a using-declaration to this derived class to inherit the constructors of "RuleWithActions", and remove the ones you manually duplicated](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2J8H3hiFLGRloh_&id=owasp-modsecurity_ModSecurity)
 * `headers/modsecurity/rule_with_actions.h`
   * [Use "override" or "final" instead of "virtual"](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2JqH3hiFLGRloh-&id=owasp-modsecurity_ModSecurity)
 * `src/rule_with_actions.cc`
   * [This function should be declared "const"](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2OIH3hiFLGRloiC&id=owasp-modsecurity_ModSecurity)
     * Existing code, not introduced in this PR.
 * `src/rule_with_operator.cc`
   * ["std::move" should not be called on a const object.](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2NmH3hiFLGRloiB&id=owasp-modsecurity_ModSecurity)
     * Nice catch!
 * `src/transaction.cc`
   * [Make the type of this parameter a pointer-to-const. The current type of "id" is "char *"](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG6yQPq1oQi0_rGKa9d&id=owasp-modsecurity_ModSecurity)
     * Existing code, not introduced in this PR.

Did not address:

 * `src/transaction.cc`
   * [Replace this use of "void *" with a more meaningful type](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG6yQPq1oQi0_rGKa9e&id=owasp-modsecurity_ModSecurity) (2)
     * `logCbData` is a C-style opaque pointer required by the C callback interface.
   * [Replace this const reference to "std::string" by a "std::string_view"](https://sonarcloud.io/project/issues?pullRequest=3248&open=AZG4I2M-H3hiFLGRloiA&id=owasp-modsecurity_ModSecurity)
     * I wanted to address this, but using a `std::string_view` (which are passed by value) introduces a `cppcheck` warning, which would require a suppression.

eduar-hte:
> This is part of a series of PRs to improve performance of the library (5/n). Previous: #3231

Even though the goal of the PR is to simplify the codebase, commit 2ad87f640f4445e2b6a62dc7bf68a94d0c382d98 (Reference RuleWithActions & Transaction object instead of copying values in RuleMessage) provides some performance improvements because it removes a number of variables copied from the associated rule & transaction.

 * Mac mini (1st generation)
   * v3/master: 83.44 secs
   * This PR: 81.44 secs (2.4% improvement)
   * v3/master: 121.83 secs
   * This PR: 117.45 secs (3.59% improvement)

airween (approved):

airween:
Hi @eduar-hte 

Just one reminder to myself:

> PS: I'll take a look at the `Transaction` constructors and see if they can be simplified as well. They actually look identical, other than the `id` parameter, so they could be refactor to use a [delegating constructor](https://www.ibm.com/docs/en/zos/2.4.0?topic=only-delegating-constructors-c11).

Seems like the `Transaction` class public constructors are (not) a bit different that it documented in [README.md](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/README.md#library-utilization):

```C++
Transaction *modsecTransaction = new Transaction(modsec, rules);
```
or in [source documentation](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/transaction.cc#L91), so we need to align them.

PR #3253, opened 2024-09-10
Simplified handling of RuleMessage by removing usage of std::shared_ptr

eduar-hte:
## what

Replace usage of `std::shared_ptr` to pass a `RuleMessage` instance with a reference.

## why

A `RuleMessage` instance is created in `bool RuleWithActions::evaluate(Transaction *transaction)` (see  `src/rule_with_actions.cc`), and is then passed around many times in the evaluation of the rule with a `std::shared_ptr` copy in every call. Notice that this doesn't copy the `RuleMessage` instance, which could be costly, but copying a `std::shared_ptr` is not cheaper than a reference because its control block is guaranteed to be thread-safe, and thus these copies use synchronization primitives (such as atomics) which decrease performance.

Because the lifetime of the `RuleMessage` is limited to the `RuleWithActions::evaluate` method where it's created, an instance can be created in the stack (instead of allocating it in the heap and creating a `std::shared_ptr`) and then simply pass the instance by reference in each function call.

NOTE: Additionally, the current implementation doesn't validate that the received `std::shared_ptr` stores a valid instance, and just dereferences it. The current version of the code guarantees that the pointer is not going to be `nullptr` but this is not checked nor asserted. Using a reference explicitly states in the code that each function expects a valid instance, and thus the compiler can enforce it in every method call and the code is explicit in its requirements.

`RuleWithActions::performLogging` currently creates a new `RuleMessage` instance on the heap and switches it with the one held in the `std::shared_ptr` (after storing the current `RuleMessage` in the list of messages in the `Transaction`) in a specific condition (when `lastLog` is `false` and the following condition holds: `hasMultimatch() && isItToBeLogged`. This code has been adjusted to leverage the existing (and unused) `clean` method in `RuleMessage`, which has been renamed as `reset`, and just reset the fields in the `RuleMessage` instance being used.

## misc

This is part of a series of PRs to improve performance of the library (6/n). Previous: #3248


eduar-hte:
> This is part of a series of PRs to improve performance of the library (6/n). Previous: #3248

In addition to simplifying the codebase, this change introduces some minor performance improvements by replacing the copies of `std::shared_ptr` with a faster/cheaper reference.

 * Mac mini (1st generation)
   * v3/master: 81.33 secs
   * This PR: 79.94 secs (1.71% improvement)
   * v3/master: 117.98 secs
   * This PR: 114.19 secs (3.22% improvement)

eduar-hte:
> Issues  [25 New issues](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3253&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true) 

Addressed most of the simple issues (unrelated to this PR) reported by Sonarcloud to reduce 'technical debt'.

The following issues were not addressed to avoid introducing potential bugs in code that was not implemented in this PR:

 * `Refactor this code to not nest more than 3 if|for|do|while|switch statements`
 * `Refactor this function to reduce its Cognitive Complexity from 30 to the 25 allowed.`

airween (dismissed):

airween:
Hi @eduar-hte,

many thanks for this great PR. Basically everything is fine, I just have one "side" question: you moved two files from `test/regression` directory to `test/common/` (`custom_debug_log.(cc|h)`) (which is not a problem, you're right, they are in a better place there), but changed the syntax (I mean format/indentation). These files used to be [OTR](https://en.wikipedia.org/wiki/Indentation_style#One_True_Brace) style, but it seems like now they have [Allman](https://en.wikipedia.org/wiki/Indentation_style#Allman_style).

This is no problem at all, you don't need to modify them, but let me ask: was that intentional or accidentally? I already thinking we should introduce a style convention (for both versions) which should be checked in CI workflow. If once we get there (at this stage), this should be reformatted, I guess.

eduar-hte:
> This is no problem at all, you don't need to modify them, but let me ask: was that intentional or accidentally? I already thinking we should introduce a style convention (for both versions) which should be checked in CI workflow. If once we get there (at this stage), this should be reformatted, I guess.

You're right, this was an oversight in the 'Introduce ModSecurityTestContext to encapsulate setup of objects required to execute transactions' commit. My editor sometimes does update the formatting when saving files, and I usually catch this and revert those changes.

I've edited the commits to update the formatting to align with the rest of the codebase, as I don't like introducing formatting changes (that overwrite 'blame' history). Other than the moved `custom_debug_log.*` files, I think the most significant (unnecessary) reformatting was in `test/regression/regression.cc`.

eduar-hte:
> (...) I already thinking we should introduce a style convention (for both versions) which should be checked in CI workflow. If once we get there (at this stage), this should be reformatted, I guess.

I haven't used it, but `clang-format` (and introducing a `.clang-format` file to the project) could be useful with this.

Did you notice the `check-style` (and `check-coding-style`) configs in `Makefile.am` (see [here](https://github.com/owasp-modsecurity/ModSecurity/blob/9a1155ca26a841be981417a8c6df06c9c54d24da/Makefile.am#L76))? I think this is a mix of static code checks but also formatting (it follows Google's C++ Style Guide).

airween:
> > This is no problem at all, you don't need to modify them, but let me ask: was that intentional or accidentally? I already thinking we should introduce a style convention (for both versions) which should be checked in CI workflow. If once we get there (at this stage), this should be reformatted, I guess.
> 
> You're right, this was an oversight in the 'Introduce ModSecurityTestContext to encapsulate setup of objects required to execute transactions' commit. My editor sometimes does update the formatting when saving files, and I usually catch this and revert those changes.
> 
> I've edited the commits to update the formatting to align with the rest of the codebase, as I don't like introducing formatting changes (that overwrite 'blame' history). Other than the moved `custom_debug_log.*` files, I think the most significant (unnecessary) reformatting was in `test/regression/regression.cc`.

Oh, many thanks!

airween:
> > (...) I already thinking we should introduce a style convention (for both versions) which should be checked in CI workflow. If once we get there (at this stage), this should be reformatted, I guess.
> 
> I haven't used it, but `clang-format` (and introducing a `.clang-format` file to the project) could be useful with this.
> 
> Did you notice the `check-style` (and `check-coding-style`) configs in `Makefile.am` (see [here](https://github.com/owasp-modsecurity/ModSecurity/blob/9a1155ca26a841be981417a8c6df06c9c54d24da/Makefile.am#L76))? I think this is a mix of static code checks but also formatting (it follows Google's C++ Style Guide).

I still haven't used `.clang-format` yet, but thanks for bringing this up, I'll check that.

Also haven't noticed that part of `Makefile.am`, now I just checked and realized it does nothing (perhaps because I don't have `cpplint.py`).

In another project where I participate as developer we use [astyle](https://github.com/Tlf/tlf/blob/master/tools/astylerc), which is easy to integrate into the CI workflow. We can take a look at that.

Thanks again.

Please let me know if you finished the work on this PR, I will merge that.

eduar-hte:
> Please let me know if you finished the work on this PR, I will merge that.

Yes, I'm done with this PR.

sonarcloud is now reporting a couple of issues about using `std::make_shared` & `std::make_unique` but those were already addressed in #3254.

airween (approved):

airween:
Thanks again, @eduar-hte!

PR #3254, opened 2024-09-10
Leverage std::make_unique & std::make_shared to create objects in the heap

eduar-hte:
## what

Replace naked `new` calls to create object instances in the heap that are then used to create `std::unique_ptr` & `std::shared_ptr` instances.

## why

Simplifies code and follows C++ Core Guidelines, see [R.22: Use make_shared() to make shared_ptrs](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r22-use-make_shared-to-make-shared_ptrs) & [R.23: Use make_unique() to make unique_ptrs](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r23-use-make_unique-to-make-unique_ptrs)

Additionally, the use of `std::make_shared` allows for an optimization by which the shared pointer's control block and the object can be allocated in the same memory block, which increasees memory locality (and thus performance).

### DIFF
--- a/headers/modsecurity/actions/action.h
+++ b/headers/modsecurity/actions/action.h
@@ -105,11 +105,11 @@ class Action {
         }
 
         if (pos == std::string::npos) {
-            m_name = std::shared_ptr<std::string>(new std::string(data));
+            m_name = std::make_shared<std::string>(data);
             return;
         }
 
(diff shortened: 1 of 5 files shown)


airween (approved):

airween:
@eduar-hte - thanks again for this PR!

Issue #3255, opened 2024-09-11
SIGSEGV writing log on FreeBSD


einsibjarni:
**Describe the bug**

I'm trying to update the [FreeBSD port](https://www.freshports.org/www/mod_security/) to 2.9.8 (current version in ports is 2.9.6) and to link with pcre2 instead of pcre.

I can build it and apache runs, but when modsecurity tries to write an audit log, the httpd process dies with SIGSEGV.
Setting `SecAuditLogType Concurrent` stops httpd from dying.

**Logs and dumps**

Output of:

[httpd.zip](https://github.com/user-attachments/files/16962634/httpd.zip)

_Notice:_ Be carefully to not leak any confidential information.

**To Reproduce**

Anything that causes mod_security to write an audit log.

**Expected behavior**

**Server (please complete the following information):**
 - ModSecurity version (and connector): [e.g. ModSecurity v3.0.1 with nginx-connector v1.0.0]
 - 2.9.8
 - WebServer: [e.g. nginx-1.15.5]
 - apache 2.4.62
 - OS (and distro): [e.g. Linux, archlinux]
 - FreeBSD 14.1-RELEASE

**Rule Set (please complete the following information):**
 - Running any public or commercial rule set? [e.g. SpiderLabs commercial rules]
 - What is the version number? [e.g. 2018-08-11]

**Additional context**

The stacktrace indicates that the problem is in apr_global_mutex_lock().

The configure command that the ports build generates and uses to build is:
`./configure --with-apxs=/usr/local/sbin/apxs --with-pcre2=/usr/local --with-yajl=/usr/local --with-curl=/usr/local --without-ssdeep --without-lua --disable-mlogc --prefix=/usr/local --localstatedir=/var --mandir=/usr/local/share/man --disable-silent-rules --infodir=/usr/local/share/info/ --build=amd64-portbld-freebsd14.1`

I'm trying to get the port in FreeBSD ports updated, so even though I can workaround the problem, we need to address this issue if it is to be merged upstream.

airween:
@einsibjarni thanks for reporting.

Could you share your stacktrace output? Eg. `bt full`.

@marcstern could you take a look at this? I assume this can be related to PR #3191.

einsibjarni:
> Could you share your stacktrace output? Eg. `bt full`.

The core dump is attached to the post as httpd.zip, but the output from `bt full` is:

```
[...]

PR #3257, opened 2024-09-12
msr_global_mutex_lock: handle errors from apr_global_mutex_lock

marcstern:
`apr_global_mutex_lock` is sometimes called with a lock that wasn't created (for any reason).
In this case, the pointer is null and `apr_global_mutex_lock` dereferences a null pointer, leading to a crash.
This PR creates a wrapper around `apr_global_mutex_lock` to handle checking that and correct logging.
Same for `msr_global_mutex_unlock`.

### DIFF
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -126,7 +126,7 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
     apr_status_t rc;
     apr_file_t *lock_name;
     const char *temp_dir;
-    const char *filename;
+    const char *filename = NULL;
 
     // get platform temp dir
     rc = apr_temp_dir_get(&temp_dir, mp);
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -166,6 +166,24 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
 #endif /* MSC_TEST */
     return APR_SUCCESS;
 }
+
+/**
+ * handle errors from apr_global_mutex_lock
+ */
+int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
+        return -1;
+    }
+
+    int rc = apr_global_mutex_lock(msr->modsecurity->auditlog_lock);
+    if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to lock global mutex: %s", get_apr_error(msr->mp, rc));
+    return rc;
+}
+
 /**
  * Initialise the modsecurity engine. This function must be invoked
  * after configuration processing is complete as Apache needs to know the
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -183,6 +183,22 @@ int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char*
     if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to lock global mutex: %s", get_apr_error(msr->mp, rc));
     return rc;
 }
+/**
+ * handle errors from apr_global_mutex_unlock
+ */
+int msr_global_mutex_unlock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
+        return -1;
+    }
+
+    int rc = apr_global_mutex_unlock(msr->modsecurity->auditlog_lock);
+    if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to lock global mutex: %s", get_apr_error(msr->mp, rc));
+    return rc;
+}
 
 /**
  * Initialise the modsecurity engine. This function must be invoked
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -196,7 +196,7 @@ int msr_global_mutex_unlock(modsec_rec* msr, apr_global_mutex_t* lock, const cha
     }
 
     int rc = apr_global_mutex_unlock(msr->modsecurity->auditlog_lock);
-    // We should have get the warning at lock time, so ignore it here
+    // We should have been warnend during lock acquisition already, so don't log the same warning here
     // if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to unlock global mutex: %s", get_apr_error(msr->mp, rc));
     return rc;
 }
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -179,7 +179,7 @@ int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char*
         return -1;
     }
 
-    int rc = apr_global_mutex_lock(msr->modsecurity->auditlog_lock);
+    int rc = apr_global_mutex_lock(lock);
     if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to lock global mutex: %s", get_apr_error(msr->mp, rc));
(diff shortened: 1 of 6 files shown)


marcstern:
https://github.com/owasp-modsecurity/ModSecurity/issues/3255 acknowledged to be solved by the PR

fzipi (approved):
LGTM

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
+        return -1;
+    }
+
+    int rc = apr_global_mutex_unlock(msr->modsecurity->auditlog_lock);
+    // We should have get the warning at lock time, so ignore it here
```
theseion, line 199:
```suggestion
    // We should have been warnend during lock acquisition already, so don't log the same warning here
```

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+    // if (rc != APR_SUCCESS) msr_log(msr, 1, "Audit log: Failed to unlock global mutex: %s", get_apr_error(msr->mp, rc));
```
theseion, line 200:
```suggestion
```

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+
+/**
+ * handle errors from apr_global_mutex_lock
+ */
+int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
```
theseion, line 178:
```suggestion
        msr_log(msr, 1, "%s: Global mutex not initialised", fct);
```

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+}
+/**
+ * handle errors from apr_global_mutex_unlock
+ */
+int msr_global_mutex_unlock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
```
theseion, line 194:
```suggestion
        msr_log(msr, 1, "%s: Global mutex not initialised", fct);
```

theseion (dismissed):

airween:
Before we merge, I want to check why the lock acquire was failed (see #3255 - the user reported that this issue came with the new version where we introduced this method).

marcstern:
The problem was already happening in the previous versions. Looks to me linked to concurrency.

apache2/modsecurity.c
marcstern: I know leaving de-activated code is not a bad practice, but I wonder if it's not a good idea to let it for a while to let a chance to future reviewers to chanllenge this.

airween:
> The problem was already happening in the previous versions. Looks to me linked to concurrency.

May be - but now the audit.log is empty (with this patch too).

I think we should investigate this issue. I already installed FreeBSD and could reproduce the behavior, so I'm on it.

marcstern:
the audit.log is empty, but you should have an entry in the error log (about the problem at creation time)

airween:
> the audit.log is empty, but you should have an entry in the error log (about the problem at creation time)

Indeed, but I think we should investigate the issue.

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex not initialised", fct);
+        return -1;
+    }
+
+    int rc = apr_global_mutex_lock(msr->modsecurity->auditlog_lock);
```
airween: Here you always use the `auditlog_lock` even though the argument can be different. I assume the correct form here is
```C
    int rc = apr_global_mutex_lock(lock);
```

apache2/modsecurity.c
```diff
@@ -166,6 +166,41 @@ int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
...
+int msr_global_mutex_unlock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct) {
+    assert(msr);
+    assert(msr->modsecurity); // lock is msr->modsecurity->..._lock
+    assert(msr->mp);
+    if (!lock) {
+        msr_log(msr, 1, "%s: Global mutex was not created", fct);
+        return -1;
+    }
+
+    int rc = apr_global_mutex_unlock(msr->modsecurity->auditlog_lock);
```
airween: Same as above, the mutex is always the same, use the argument:
```C
    int rc = apr_global_mutex_unlock(lock);
```

apache2/msc_logging.c
```diff
@@ -1471,15 +1464,8 @@ void sec_audit_logger_json(modsec_rec *msr) {
...
-
         /* Unlock the mutex we used to serialise access to the audit log file. */
-        rc = apr_global_mutex_unlock(msr->modsecurity->auditlog_lock);
-        if (rc != APR_SUCCESS) {
-            msr_log(msr, 1, "Audit log: Failed to unlock global mutex '%s': %s",
-                    apr_global_mutex_lockfile(msr->modsecurity->auditlog_lock),
-                    get_apr_error(msr->mp, rc));
-        }
-
+        msr_global_mutex_lock(msr, msr->modsecurity->auditlog_lock, "Audit log");
```
airween: Didn't you want to use the unlock function here?
```C
    msr_global_mutex_unlock(msr, msr->modsecurity->auditlog_lock, "Audit log");
```

apache2/msc_logging.c
```diff
@@ -2253,15 +2235,8 @@ void sec_audit_logger_native(modsec_rec *msr) {
...
```
airween: Same as above, use the unlock function.
```C
    msr_global_mutex_unlock(msr, msr->modsecurity->auditlog_lock, "Audit log");
```

airween (changes requested):
Please see my comments.

apache2/msc_logging.c
marcstern: Indeed, fixed

apache2/msc_logging.c
marcstern: Indeed, fixed

marcstern:
> > the audit.log is empty, but you should have an entry in the error log (about the problem at creation time)
> 
> Indeed, but I think we should investigate the issue.
Correct. We should open a new ticket for this I guess (or continue https://github.com/owasp-modsecurity/ModSecurity/issues/3255)

airween:
> Indeed, but I think we should investigate the issue.
> Correct. We should open a new ticket for this I guess (or continue #3255)

yes, I made some investigation - see my [comment](https://github.com/owasp-modsecurity/ModSecurity/issues/3255#issuecomment-2382897497) there.

marcstern:
I propose to accept this PR first, then create one that doesn't use the temp filename, as it's not the right way to create a mutx, see https://lists.apache.org/thread/ykb26kg4lgcqnldvxwd9p6hv16fy4z9l

airween (approved):

apache2/modsecurity.c
airween: I think we can resolve this conversation at the moment. Later we can remove that old code.

airween (approved):

airween:
> I propose to accept this PR first, then create one that doesn't use the temp filename, as it's not the right way to create a mutx, see https://lists.apache.org/thread/ykb26kg4lgcqnldvxwd9p6hv16fy4z9l

Okay, I approved this PR and resolved all conversation. You can merge this PR now.

PR #3262, opened 2024-09-25
Missing #include <time.h>

marcstern:
"struct tm" defined in time.h is used in re_variables.c.
time.h is not explicitely included, but is included via apr/apr-util.
In latest version of apr/apr-util, the inclusion is no more present, leading to a compilation failure.
This PR just adds the missing include.


airween (approved):
LGTM

PR #3268, opened 2024-10-03
Fix apr_global_mutex_create() usage

marcstern:
Fix for https://github.com/owasp-modsecurity/ModSecurity/issues/3255
We don't have to generate a temp name ourselves, it'll be done in apr_global_mutex_create().
We don't have to provide a filename, apr_global_mutex_create() generates one automatically.
Moreover, under Unix & Windows, the preferred mechanism won't use a file at all.
apr_file_mktemp() cannot be used as it creates the file (at least on FreeBSD).
Discussion in Apache mailing list: https://lists.apache.org/thread/ykb26kg4lgcqnldvxwd9p6hv16fy4z9l

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -1,3 +1,9 @@
+(to be released) - 2.9.x 
+------------------------
+
+ * handle errors from apr_global_mutex_lock
+   [PR #3257 - @marcstern]
+
 03 Sep 2024 - 2.9.8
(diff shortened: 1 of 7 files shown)


marcstern:
Replaced by https://github.com/owasp-modsecurity/ModSecurity/pull/3269

PR #3279, opened 2024-10-17
Fixed PCRE2 error message

marcstern:
In case an error occurs in PCRE2 compilation, the error message is not displayed.
Example: "Error creating rule: Error compiling pattern (offset 6): "

With the fix, the PCRE2 error message is retrieved:
Example: "Error creating rule: Error compiling pattern (offset 6): length of lookbehind assertion is not limited"

The code was mimicing PCRE behaviour; PCRE2 error message handling changed.


airween (approved):
LGTM

PR #3280, opened 2024-10-17
Replace usage of range-checked 'at' method when vector/string has already been size checked

eduar-hte:
## what

Replace places in the codebase where the `at` method of a `std::vector` or `std::string` instance is used to access an element instead of the subscript operator (`[]`), and the index of the item is guaranteed to be valid.

Additionally, simplify some of these uses where an old-style C loop is used to iterate over all elements and replace it with a range-based loops (which also implicitly guarantees no out-of-bound access).

## why

The `at` method is not equivalent to the subscript operator (`[]`) operator, as it's range-checked and can thus throw a `std::out_of_range` exception which makes the code slower not only because of the additional check but also due to the generation of exception handling code at these call sites.

Because the updated instances are guaranteed to access valid indexes in the container, we can replace the usage of `at` to improve performance.

And where possible, simplify the code by adopting ranged-based loops.

## misc

This is part of a series of PRs to improve performance of the library (7/n). Previous: #3253

### DIFF
--- a/headers/modsecurity/anchored_set_variable_translation_proxy.h
+++ b/headers/modsecurity/anchored_set_variable_translation_proxy.h
@@ -100,13 +100,9 @@ class AnchoredSetVariableTranslationProxy {
             return nullptr;
         }
 
-        std::unique_ptr<std::string> ret(new std::string(""));
+        auto ret = std::make_unique<std::string>(l[0]->getValue());
 
-        ret->assign(l.at(0)->getValue());
-
-        while (!l.empty()) {
-            auto &a = l.back();
-            l.pop_back();
+        for(auto a : l) {
             delete a;
         }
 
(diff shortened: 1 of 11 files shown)


eduar-hte:
> This is part of a series of PRs to improve performance of the library (7/n). Previous: #3253

These changes provide minor performance improvements to the library.

These results were obtained running the `benchmark` test with 100'000 iterations with the OWASP CRS v4 and are presented as a reference (average of three runs):

 * Mac mini (1st generation)
   * v3/master: 80.74 secs
   * This PR: 80.39 secs (0.44% improvement)
   * v3/master: 119.01 secs
   * This PR: 114.87 secs (3.5% improvement)

airween (approved):
LGTM

airween:
Thanks @eduar-hte!

Issue #3129, opened 2024-04-23
Incorrect utf8toUnicode transformation for 00xx


marcstern:
In utf8_unicode_inplace_ex(), we have the following code:
```
c = *utf;
/* If first byte begins with binary 0 it is single byte encoding */
if ((c & 0x80) == 0) {
    /* single byte unicode (7 bit ASCII equivilent) has no validation */
    count++;
    if (count <= len) {
        if (c == 0) *data = x2c(&c);
        else *data++ = c;
    }
}
```
The code `if (c == 0) *data = x2c(&c);` is wrong.
utf the input string.
If input is "x", we copy the character => correct.
If input is "\x00\xA0", we call the x2c function which expects an hexadecimal string as input (like "20" for a space).
Furthermore, the output buffer (data) is not increased, so the character is overwritten on the next loop.
The correct code is
```
if (c == 0) {
    sprintf(data, "%%u%04x", utf[1]);
    count += 4;
    data += 6;
    i++;
    *changed = 1;
}
```

Remarks:

1. Using "%04x" replaces a lot of useless code used in other parts of the code
2. I didn't add a check `if (count <= len)` after `count += 4;` because it cannot overflow if we fix the code: `len = input_len * 6 + 1;` instead of `len = input_len * 4 + 1;` (4 bytes -> %u1234). The check (and the variable count) could thus be removed everywhere
3. After filling data, we have several checks (/* invalid UTF-8 character number range (RFC 3629) */, /* check for overlong */, ...) with this code `count++; *data++ = c;` What's the point? This copies the character after the ones we already copied. This looks wrong to me. Any reason to keep this code?

airween:
Hi @marcstern, is this sill open? I mean haven't you add something relevant code which fixed this?

marcstern:
I'll make a PR for this

marcstern:
Fixed in https://github.com/owasp-modsecurity/ModSecurity/pull/3284

PR #3284, opened 2024-10-22
Incorrect utf8toUnicode transformation for 00xx

marcstern:
Fix issue and restructure handling
https://github.com/owasp-modsecurity/ModSecurity/issues/3129

Perform all the checks and calculate the multibute length, then write the bytes at a single place (the correct way).

### DIFF
--- a/apache2/msc_util.c
+++ b/apache2/msc_util.c
@@ -112,7 +112,7 @@ char *utf8_unicode_inplace_ex(apr_pool_t *mp, unsigned char *input, long int inp
     unsigned int bytes_left = input_len;
     unsigned char *unicode = NULL;
 
-    if (input == NULL) return NULL;
+    assert(input != NULL);
 
     *changed = 0;
     /* RFC3629 states that UTF-8 are encoded using sequences of 1 to 4 octets. */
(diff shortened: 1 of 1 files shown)


apache2/msc_util.c
```diff
@@ -105,225 +105,101 @@ int swap_int32(int x) {
...
     int unicode_len = 0, length = 0;
-    unsigned int d = 0, count = 0;
+    unsigned int d = 0;
     unsigned char c, *utf;
     char *rval, *data;
     unsigned int i, len, j;
     unsigned int bytes_left = input_len;
     unsigned char *unicode = NULL;
 
+    if (input == NULL) return NULL;
```
airween: Is `NULL` allowed here? If not, we could use `assert()`.

airween:
Hi @marcstern,

I added one comment above, and would like to ask if we modify a function that used by a transformation (`utf8toUnicode`) then probably we should add a test case to our CI. What do you think?

apache2/msc_util.c
marcstern: I changed it

airween (approved):

airween:
@marcstern do you think we can add a test case to check this transformation's behavior? Or it's not necessary?

marcstern:
We should ideally add  test for this, indeed

PR #3285, opened 2024-10-22
Don't explicitly use files for mutex creation

marcstern:
We don't have to generate a temp name ourselves, it'll be done in apr_global_mutex_create().
We don't have to provide a filename, apr_global_mutex_create() generates one automatically.
Moreover, under Unix & Windows, the preferred mechanism won't use a file at all.
apr_file_mktemp() cannot be used as it creates the file (at least on FreeBSD).
Discussion in Apache mailing list: https://lists.apache.org/thread/ykb26kg4lgcqnldvxwd9p6hv16fy4z9l

### DIFF
--- a/apache2/modsecurity.c
+++ b/apache2/modsecurity.c
@@ -123,30 +123,7 @@ msc_engine *modsecurity_create(apr_pool_t *mp, int processing_mode) {
 }
 
 int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp) {
-    apr_status_t rc;
-    apr_file_t *lock_name;
-    const char *temp_dir;
-    const char *filename;
-
-    // get platform temp dir
-    rc = apr_temp_dir_get(&temp_dir, mp);
-    if (rc != APR_SUCCESS) {
-        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, mp, "ModSecurity: Could not get temp dir");
-        return -1;
-    }
-
-    // use temp path template for lock files
-    char *path = apr_pstrcat(mp, temp_dir, GLOBAL_LOCK_TEMPLATE, NULL);
-
-    rc = apr_file_mktemp(&lock_name, path, 0, mp);
-    if (rc != APR_SUCCESS) {
-        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, mp, " ModSecurity: Could not create temporary file for global lock");
-        return -1;
-    }
-    // below func always return APR_SUCCESS
-    apr_file_name_get(&filename, lock_name);
-
-    rc = apr_global_mutex_create(lock, filename, APR_LOCK_DEFAULT, mp);
+    apr_status_t rc = apr_global_mutex_create(lock, NULL, APR_LOCK_DEFAULT, mp);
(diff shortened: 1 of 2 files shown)


apache2/modsecurity.h
```diff
@@ -707,6 +705,8 @@ struct msc_parm {
 
 /* Reusable functions */
 int acquire_global_lock(apr_global_mutex_t **lock, apr_pool_t *mp);
+int msr_global_mutex_lock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct);
+int msr_global_mutex_unlock(modsec_rec* msr, apr_global_mutex_t* lock, const char* fct);
```
airween: Why are these declarations here? I can't find any implementation in any other places.

airween (changes requested):

airween:
@marcstern ping.

apache2/modsecurity.h
marcstern: Indeed, it's fixed

fzipi (approved):
There is no better feeling than removing unneeded code.

apache2/modsecurity.c
```diff
@@ -123,30 +123,7 @@ msc_engine *modsecurity_create(apr_pool_t *mp, int processing_mode) {
...
-        ap_log_perror(APLOG_MARK, APLOG_ERR, 0, mp, " ModSecurity: Could not create temporary file for global lock");
-        return -1;
-    }
-    // below func always return APR_SUCCESS
-    apr_file_name_get(&filename, lock_name);
-
-    rc = apr_global_mutex_create(lock, filename, APR_LOCK_DEFAULT, mp);
+    apr_status_t rc = apr_global_mutex_create(lock, NULL, APR_LOCK_DEFAULT, mp);
     if (rc != APR_SUCCESS) {
         ap_log_perror(APLOG_MARK, APLOG_ERR, 0, mp, " ModSecurity: Could not create global mutex");
```
notroj, line 128: Passing `rc` as the `apr_status_t` argument rather than `0`  would be better.

notroj (approved):

airween:
@notroj,

could you help us to review this PR and finish it (if it's necessary to do anything). Thanks!

notroj:
I can't push to this PR branch but I'd add:

https://github.com/notroj/ModSecurity/commit/aac61ae4263521bae01e222599241d7874b65d58

and merge this

airween:
> I can't push to this PR branch but I'd add:
> 
> [notroj@aac61ae](https://github.com/notroj/ModSecurity/commit/aac61ae4263521bae01e222599241d7874b65d58)
> 
> and merge this

Ah, you're right - thanks. Probably we should open a new PR, based on this one and then odd yours.

I can do that later or feel free to do it.

Thanks anyway.

notroj:
Actually it looks like the first commits from this PR were merged already in 1121ef0bed4f083eb6de1e7c49daf49ab554a1da and the third one looks wrong. So I just made a new PR #3385 with the suggested change. I think this PR can be closed.

airween:
Seems like this PR is orphan, I mean the modifications are the part of an another PR (#3269) and already merged. Closing as the issue is solved and PR is abandoned.

PR #3298, opened 2024-11-12
Fix m_requestHostName variable behavior

airween:
## what

This PR fixes the wrong behavior of `m_requestHostName` transaction variable.

## why

Previously, this variable was set when the `setRequestHostName()` was called, but previously it was initialized in `processConnection()`. So the last function overwritten the variable, and the previously set value disappeared.

If the connector calls if after `processConnection()`, then the `host` field in log entries which generated during `processConnection()` phase does not contain the necessary value.

The solution was to add a condition which sets the `m_requestHostName` only if it's empty. Also the correct call to the function `setRequestHostName()` can be seen in the `regression.cc`.

This modification does not modify the API.

### DIFF
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -227,7 +227,9 @@ int Transaction::processConnection(const char *client, int cPort,
     const char *server, int sPort) {
     m_clientIpAddress = client;
     m_serverIpAddress = server;
-    m_requestHostName = server;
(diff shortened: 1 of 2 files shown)


marcstern (approved):

airween:
No other notes, going to merge this.

PR #3306, opened 2024-11-22
fix: align TIME_MON variable's behavior

### LINKED ISSUE #3305: `TIME_MON` variable: inconsistent ranges between `2.x` and `3.x` engines.
**Describe the bug**
Both `v2` and `v3` docs are reporting the following line about `TIME_MON` variable:
```
This variable holds the current month value (0–11). The following rule matches if the month is either November (value 10) or December (value 11)
```
But [Apache docs](https://httpd.apache.org/docs/2.4/zh-cn/expr.html) provide the following:
```
TIME_MON	The current month (01, ..., 12)
```
And [v2 implementation](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/re_variables.c#L2383) seems that it is indeed following it (Thanks @airween for the pointer): 
```
rvar->value = apr_psprintf(mptmp, "%02d", tm->tm_mon + 1);
```

**Logs and dumps**
Same rule and same request:
v2:
```
Executing operator "eq" with param "11" against TIME_MON.
Target value: "11"
```
v3:
```
(Rule: 800) Executing operator "Eq" with param "11" against TIME_MON.
Target value: "10" (Variable: TIME_MON)
```
**Expected behavior**
Consistency across engines, and/or documentation aligned.

**Additional context**
1) Coraza is implementing `TIME` variables, and it would be great to come up with a unified behavior across all the engines. I'm personally in favour of ranging between `1-12` as it sounds very natural and as far as I know it is the standard way. E.g. [ISO_8601](https://www.loc.gov/standards/datetime/iso-tc154-wg5_n0039_iso_wd_8601-2_2016-02-16.pdf) reports: `[MM] indicates a two-digit month of the year, 01 through 12`
2) It might be worth another issue, but looks like also `TIME_WDAY` is not consistent. This time docs are reporting different behaviours:
`This variable holds the current weekday value (1–7), where Monday is 1`
`This variable holds the current weekday value (0–6). The following rule triggers only on Satur- day and Sunday`

airween:
## what

This PR changes `TIME_MON` variable's behavior and fixes #3305.

Also fixes the regex in changed variable's test file (the old one allows any number, eg. 0 or 1111 which makes no sense; the new one allows between 1 and 12.).

## why

As the issue describes in libmodsecurity3 the `TIME_MON` variable can be between 0 and 11. mod_security2 uses values between 1 and 12 - which is much more natural.

## references

See issue #3305.

### DIFF
--- a/test/test-cases/regression/variable-TIME_MON.json
+++ b/test/test-cases/regression/variable-TIME_MON.json
@@ -33,7 +33,8 @@
       ]
     },
     "expected":{
-      "debug_log":"Target value: \"([0-9]+)\" \\(Variable: TIME_MON\\)"
+      "http_code": 200,
+      "debug_log":"Target value: \"[1-9][012]?\" \\(Variable: TIME_MON\\)"
     },
     "rules":[
       "SecRuleEngine On",
--- a/test/test-cases/regression/variable-TIME_MON.json
+++ b/test/test-cases/regression/variable-TIME_MON.json
@@ -34,7 +34,7 @@
     },
     "expected":{
       "http_code": 200,
-      "debug_log":"Target value: \"[1-9][012]?\" \\(Variable: TIME_MON\\)"
+      "debug_log":"Target value: \"([1-9]|1[012])\" \\(Variable: TIME_MON\\)"
     },
     "rules":[
       "SecRuleEngine On",
--- a/src/variables/time_mon.cc
+++ b/src/variables/time_mon.cc
@@ -46,7 +46,7 @@ void TimeMon::evaluate(Transaction *transaction,
     struct tm timeinfo;
     localtime_r(&timer, &timeinfo);
 
-    transaction->m_variableTimeMin.assign(std::to_string(timeinfo.tm_mon));
+    transaction->m_variableTimeMin.assign(std::to_string(timeinfo.tm_mon + 1));
 
     l->push_back(new VariableValue(&m_retName,
         &transaction->m_variableTimeMin));


airween:
@M4tteoP could you review this PR? Thanks!

airween:
There is a very interesting check result:
* in first attempt the `cppcheck` step showed an error [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33372492045#step:5:4039)
* I re-run that check and then the error disappeared - see [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33374020597#step:5:4038)

I tried this on my local environment (with a newer `cppcheck` (2.16.0)), and that also reports this issue:
```
$ cppcheck -U YYSTYPE -U MBEDTLS_MD5_ALT -U MBEDTLS_SHA1_ALT -D MS_CPPCHECK_DISABLED_FOR_PARSER -U YY_USER_INIT --suppressions-list=./test/cppcheck_suppressions.txt --inline-suppr --enable=warning,style,performance,portability,unusedFunction,missingInclude --inconclusive --template="warning: {file},{line},{severity},{id},{message}" -I headers -I . -I others -I src -I others/mbedtls/include --error-exitcode=1 -i "src/parser/seclang-parser.cc" -i "src/parser/seclang-scanner.cc" -i others --std=c++17 --force --verbose src/utils/system.cc 
Checking src/utils/system.cc ...
Defines:MS_CPPCHECK_DISABLED_FOR_PARSER=1
Undefines: MBEDTLS_MD5_ALT; MBEDTLS_SHA1_ALT; YYSTYPE; YY_USER_INIT
Includes: -Iheaders/ -I./ -Iothers/ -Isrc/ -Iothers/mbedtls/include/
Platform:native
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1...
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;NO_LOGS...
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;S_IFMT;S_IFREG;S_ISREG...
warning: src/utils/system.cc,213,error,syntaxError,syntax error
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;WIN32...
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;_MSC_VER...
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;__GNUC__...
Checking src/utils/system.cc: MS_CPPCHECK_DISABLED_FOR_PARSER=1;__OpenBSD__...
```

@eduar-hte do you have any idea?

M4tteoP:
> @M4tteoP could you review this PR? Thanks!

It looks good to me! I just don't know how this change will be handled regarding breaking changes. Rules relying on `TIME_MON` might require changes after this fix.

fzipi (approved):
LGTM.

test/test-cases/regression/variable-TIME_MON.json
```diff
@@ -33,7 +33,8 @@
       ]
     },
     "expected":{
-      "debug_log":"Target value: \"([0-9]+)\" \\(Variable: TIME_MON\\)"
+      "http_code": 200,
+      "debug_log":"Target value: \"[1-9][012]?\" \\(Variable: TIME_MON\\)"
```
theseion:
```suggestion
      "debug_log":"Target value: \"[1-9]|1[012]\" \\(Variable: TIME_MON\\)"
```

theseion (changes requested):

airween:
> There is a very interesting check result:
> 
>     * in first attempt the `cppcheck` step showed an error [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33372492045#step:5:4039)
> 
>     * I re-run that check and then the error disappeared - see [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33374020597#step:5:4038)

Seems like @gberkes solved the myth: `cppcheck` 2.16 detects this as FP - see #3307.

The reason is why those checks have different results above that they have two different image:
* the first one has [14.7.1](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33372492045#step:1:4), which is a newer
* the second has [14.7](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/11970218824/job/33374020597#step:1:4) which probably has an older `cppcheck`

Thanks @gberkes.

test/test-cases/regression/variable-TIME_MON.json
airween: Hmm... Your suggested pattern matches only with `"1`, the rest part is out of scope. I mean if the line contains this:
`"13"`
then the `"1` will match which is not enough for us.

Or em I wrong?

theseion: You're right. It should be `^(?:1-9|1[012])$)`.

airween: I'm afraid we can't use anchors (`^` and `$`) (and probably your last suggestion contains a typo: the last `)` is unnecessary).

The test program compiles the regex and matches it with the whole log line.

Eg. with your suggestion the whole regex will be this:
```
Target value: \"^(?:1-9|1[012])$\" \\(Variable: TIME_MON\\)
```
and the subject is something like this (if it's correct):
```
Target value: "11" (Variable: TIME_MON)
```
But this does not match.

theseion: Hm. But switching the order should do the trick: `1[012]|[0-9]`.

airween: No.

Subject:
```
Target value: "1" (Variable: TIME_MON)
```
does not match at all.
```
Target value: "10" (Variable: TIME_MON)
```
matches partially:
```
Target value: "10
```
What's wrong with the "original" one (which part of original PR)?

theseion: I don't understand what the matcher is doing. Your original regex allows arbitrary numbers like `92`.

theseion: TBH, I also don't know what that JSON thing does. If it's just a test, then maybe it doesn't matter and we'll just go with what you had originally.

airween: > I don't understand what the matcher is doing. Your original regex allows arbitrary numbers like `92`.

Ah, I see, you're right. As you figured out, the main goal is to cover the test as it accepts only numbers in range between 1 and 12. I'll try to figure out something.

airween: > TBH, I also don't know what that JSON thing does. If it's just a test, then maybe it doesn't matter and we'll just go with what you had originally.

Yes, it's just a test. Unfortunately there is no any solution to create a "dynamic" test, I mean use a request where the argument value is the current month... So we have to check the range in these cases. (Note, I think the current test is completely unusable.)

airween:
@theseion could you take a look at it again? I changed the [regex](https://github.com/owasp-modsecurity/ModSecurity/pull/3306/files#diff-38c18c914ebe6d4a0a6e8381b69fb058d50ba84d93a5b32993f9b4afc96e76a6R37), I think now it fills the expectations.

theseion (approved):

dune73:
Thanks for this fix (and @M4tteoP's report in the first case).

It's a rare, but apparently a breaking change. Please report with care. And glad you settled on the ModSec 2 behavior.

I see the attractiveness of Jan being "0", but it's really counterintuitive since we write dates differently.

PR #3319, opened 2025-01-04
fix: build library on OSX without GeoIP brew package

airween:
## what

Build GeoIP library on OSX from source in GH workflow.

## why

On OSX the GeoIP package is obsolete and could fetch through brew, therefore the GH workflow on OSX was unsuccessful, see [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/12270940148/job/34610196548?pr=3314#step:2:45).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
           fetch-depth: 0
       - name: Build GeoIP
         run: |
-          git clone --depth 0 https://github.com/maxmind/geoip-api-c.git
+          git clone --depth 1 https://github.com/maxmind/geoip-api-c.git
           cd geoip-api-c
           autoreconf --install
           ./configure --disable-dependency-tracking --disable-silent-rules --prefix=/opt/homebrew
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,7 +112,7 @@ jobs:
           fetch-depth: 0
       - name: Build GeoIP
         run: |
-          git clone https://github.com/maxmind/geoip-api-c.git
+          git clone --depth 0 https://github.com/maxmind/geoip-api-c.git
           cd geoip-api-c
           autoreconf --install
           ./configure --disable-dependency-tracking --disable-silent-rules --prefix=/opt/homebrew
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,10 +113,10 @@ jobs:
       - name: Build GeoIP
         run: |
           git clone --depth 1 --no-checkout https://github.com/maxmind/geoip-api-c.git
+          cd geoip-api-c
           git fetch --tags
           # Check out the last release, v1.6.12
           git checkout 4b526e7331ca1d692b74a0509ddcc725622ed31a
-          cd geoip-api-c
           autoreconf --install
           ./configure --disable-dependency-tracking --disable-silent-rules --prefix=/opt/homebrew
           make install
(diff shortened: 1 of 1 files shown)


airween:
Thanks for help, @theseion!

.github/workflows/ci.yml
```diff
@@ -111,6 +110,13 @@ jobs:
         with:
           submodules: true
           fetch-depth: 0
+      - name: Build GeoIP
+        run: |
+          git clone https://github.com/maxmind/geoip-api-c.git
```
```suggestion
          git clone --depth 0 https://github.com/maxmind/geoip-api-c.git
```

theseion (changes requested):

.github/workflows/ci.yml
airween: Hmm, [seems](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/12613242971/job/35150863548?pr=3319#step:4:9) like `--depth 0` isn't a good value.

Even though I couldn't find any information in [official](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt-code--depthcodeemltdepthgtem) documentation, I found an [article](https://medium.com/@thriving_chiffon_gnu_714/exploring-git-clone-depth-828eb4bac6c4) when they said the minimum value is `1`.

What do you think?

theseion: Yeah, I'm never sure :) Use `1` then.

airween: Done.

.github/workflows/ci.yml
```diff
@@ -111,6 +110,13 @@ jobs:
         with:
           submodules: true
           fetch-depth: 0
+      - name: Build GeoIP
+        run: |
+          git clone --depth 1 https://github.com/maxmind/geoip-api-c.git
```
```suggestion
          git clone --depth 1 --no-checkout https://github.com/maxmind/geoip-api-c.git
          git fetch --tags
          # Check out the last release, v1.6.12
          git checkout 4b526e7331ca1d692b74a0509ddcc725622ed31a
```

theseion (changes requested):
Sorry, had forgotten this.

.github/workflows/ci.yml
airween: I think this wold be more clear and we don't need to follow the releases:
```
git clone https://github.com/maxmind/geoip-api-c.git
cd geoip-api-c/
latest=$(git describe --tag | cut -d"-" -f1)
git checkout tags/$latest
```
Now `git status` shows me
```
$ git status
HEAD detached at v1.6.12
nothing to commit, working tree clean
```
What do you think?

theseion: I wanted to pin the commit hash, a tag can be modified. You want to be sure that you always build with the same version. The alternative would be to download the release and verify the checksum. Also, this is the last release. The repository was archived and no new releases will be published anyway, so you will never have to update the version anyway.

airween: > Also, this is the last release. The repository was archived and no new releases will be published anyway, so you will never have to update the version anyway.

Ah, you're right, this make sense - thanks.

Btw I think the `--no-checkout` option and `git fetch tags` (and `--depth 1`) aren't necessary.

This would be enough:
```
git clone https://github.com/maxmind/geoip-api-c.git
git checkout 4b526e73
```
Or why are they necessary there?

theseion: That is true, but then you clone the full history. `--no-checkout` is just to save some CPU cycles.

airween: Thanks, I added your suggestion.

theseion (approved):

PR #3321, opened 2025-01-06
Refactor/default pcre2

gberkes:
## What

Refactor the build system and the relevant source files to make the PCRE2 library the default build target. If you wish to continue building with the legacy PCRE library, you must explicitly specify it using the following command:
configure --with-pcre.

Additional build system cleanups include:

Removed unnecessary AM_CONDITIONAL lines from configure.ac.

## Why

The legacy PCRE library will be deprecated and eventually removed. The first step in this process is to make PCRE2 the default build dependency, replacing the old PCRE library.


theseion:
Yay! :)

gberkes:
> Yay! :)

Legacy code. :)

mmmmkjjui3:
I think we should have keep legacy pcre as there are many users who use for example  nas systems, routers with 3rd party software which are partialy close software and where you cannot update things or stop people building pcre3 and pcre2 withing 1 build for example nginx. Where you can compile security with pcre2 and nginx with pcre3. Chaos...

Its great as it is already, blazing fast with millions of happy users with multi platform. So kick others out  ?? 

Keep option legacy or pcre2, just set default pcre2 to conclude tests, dont need to remove excellent compatibility with 99,9% devices that can run nginx on the market.

airween:
@mmmmkjjui3,

> I think we should have keep legacy pcre as there are many users who use for example nas systems, routers with 3rd party software which are partialy close software and where you cannot update things or stop people building pcre3 and pcre2 withing 1 build for example nginx. Where you can compile security with pcre2 and nginx with pcre3. Chaos...

Well, we definitely didn't say we will finish old PCRE support of any engines.

> Its great as it is already, blazing fast with millions of happy users with multi platform. So kick others out ??

No. Where did you see that?
 
> Keep option legacy or pcre2, just set default pcre2 to conclude tests,

This is what this PR does.

> dont need to remove excellent compatibility with 99,9% devices that can run nginx on the market.

Again: we didn't say we will remove this legacy code.

airween (approved):
LGTM

airween:
Seems good to me, thank you.

My only note is the output of the configure script: if I chose specifically the `--with-pcre` option, then I get this result:
```
   + PCRE                                          ....found 
     using pcre v8.39
      -lpcre -lpcre, -DWITH_PCRE  -DPCRE_HAVE_JIT
```
But when I don't choose anything (or choose `--with-pcre2` - which is the default so it's equal with nothing), then I get:
```
   + PCRE                                          ....found 
     using pcre v
      , 
   + PCRE2                                          ....found v10.45
```
It would be nice to get the same syntax.

Now I merge this PR, but please take a look at the `configure.ac` script.

airween:
Hmm... it's very interesting that the merge workflow's cppcheck was failed:

https://github.com/owasp-modsecurity/ModSecurity/actions/runs/13817500116/job/38654442042#step:5:4377

But this one's was success:

https://github.com/owasp-modsecurity/ModSecurity/actions/runs/13435232244/job/37535916649?pr=3321

airween:
> Hmm... it's very interesting that the merge workflow's cppcheck was failed:

Okay, the reason is simple: `cppcheck` has a newer version in last workflow.

The old one: [2.16.0](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/13435232244/job/37535916649?pr=3321#step:2:62)

The new: [2.17.1](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/13817500116/job/38654442042#step:2:62)

build/pcre.m4
```diff
@@ -21,8 +21,8 @@ AC_ARG_WITH(
     [test_paths="${with_pcre}"],
     [test_paths="/usr/local/libpcre /usr/local/pcre /usr/local /opt/libpcre /opt/pcre /opt /usr /opt/local"])
 
-if test "x${with_pcre2}" != "x" && test "x${with_pcre2}" != "xno"; then
-    AC_MSG_NOTICE([pcre2 specified; omitting check for pcre])
+if test "x${with_pcre}" == "x" && test "x${with_pcre}" != "xno"; then
```
theseion, line 24: @airween the `&&` doesn't make sense here. When the first condition is true, the second condition will also always be true.
```suggestion
if test "x${with_pcre}" == "x" || test "x${with_pcre}" == "xno"; then
```

build/pcre2.m4
```diff
@@ -29,10 +29,12 @@ if test "x${with_pcre2}" == "xno"; then
     AC_MSG_NOTICE([Support for PCRE2 was disabled by the utilization of --without-pcre2 or --with-pcre2=no])
     PCRE2_DISABLED=yes
 else
-    if test "x${with_pcre2}" == "xyes"; then
-        PCRE2_MANDATORY=yes
-        AC_MSG_NOTICE([PCRE2 support was marked as mandatory by the utilization of --with-pcre2=yes])
-    fi
+     PCRE2_MANDATORY=yes
+     AC_MSG_NOTICE([PCRE2 is enabled by default.])
+#    if test "x${with_pcre2}" == "xyes"; then
```
theseion, line 34: Remove commented code.

theseion: Also remove the closing (commented) `fi`.

configure.ac
```diff
@@ -146,16 +141,16 @@ CHECK_LIBXML2
...
 #
-CHECK_PCRE
-
-
-#
-# Check for pcre2
-#
-PROG_PCRE2
-AM_CONDITIONAL([PCRE2_CFLAGS], [test "PCRE2_CFLAGS" != ""])
+if test "x${with_pcre}" != "x" && test "x${with_pcre}" != "xno"; then
```
theseion, line 146: While this test is correct, it omits the possibility of `with_pcre2` having been set to `no`. It might not make much sense, but if the user sets both `with_pcre` and `with_pcre2` to `no` then I would expect that to be honoured. As it is now, PCRE2 would be used, regardless.

PR #3322, opened 2025-01-06
fix: add value checking to @validateByteRange

airween:
## what

This patch adds a value checking for operator `@validateByteRange`. Normally this operator allows values between 0 and 255.

The necessary check was already implemented but it was wrong place. The other problem was that in case of any error during the checks code never returned with `false` value - therefore the appended errors never seemed.

Unfortunately I can't add any test cases to regression tests, because the test framework does not have any option to handle these kind of errors.

## why

If the admin gives an extra high value (eg 999999) then the engine exists with a segfault. [@Wiz_io](https://x.com/wiz_io) team reported few issues in e-mail (modsecurity at owasp dot org), the first one was this.

### DIFF
--- a/src/operators/validate_byte_range.cc
+++ b/src/operators/validate_byte_range.cc
@@ -37,6 +37,11 @@ bool ValidateByteRange::getRange(const std::string &rangeRepresentation,
                 "' into a number");
             return false;
         }
+        if ((start < 0) || (start > 255)) {
+            error->assign("Invalid range start value: " +
+                std::to_string(start));
+            return false;
+        }
         table[start >> 3] = (table[start >> 3] | (1 << (start & 0x7)));
         return true;
     }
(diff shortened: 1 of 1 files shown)


src/operators/validate_byte_range.cc
```diff
@@ -37,6 +37,11 @@ bool ValidateByteRange::getRange(const std::string &rangeRepresentation,
                 "' into a number");
             return false;
         }
+        if ((start < 0) || (start > 255)) {
```
theseion, line 40: I think you need the same check after line 50, since the start of the range could also be invalid. The check of `end` together with the check of `start > end` will technically do the same, but the user will just see `Invalid range...`, instead of `Invalid range start value:...`.

src/operators/validate_byte_range.cc
```diff
@@ -37,6 +37,11 @@ bool ValidateByteRange::getRange(const std::string &rangeRepresentation,
...
+            error->assign("Invalid range start value: " +
```
theseion: In this case we aren't checking a range but a single byte.
```suggestion
            error->assign("Invalid byte value: " +
```

theseion (changes requested):

theseion (approved):

Issue #3334, opened 2025-01-29
libyajl not found. Build script not searching correct directories.


RooHTaylor:
**Describe the bug**

When trying to `./configure --with-yajl` YAJL can never be found. If I specify `--with-yajl="/usr/include/yajl"` it will find the YAJL headers, but still report that YAJL was not found. If I include multiple paths in `--with-yajl=""` it continues to find nothing.

I tried to uninstall libyajl-dev from my package manager and manually compile yajl from sources. I experieced the same results with any combination of `/usr/local/{include,lib}` in the path.

After some digging I found that `./build/yajl.m4` tries to use pkg-config to find YAJL first, and then falls back to directory scanning. However, it appears the list of directories to check is stored in `YAJL_POSSIBLE_PATHS` while the script is passing `YAJL_POSSIBLE_LIB_NAMES` to `CHECK_FOR_YAJL_AT()` on line 65.

```
# If pkg-config did not find anything useful, go over file lookup.
for x in ${YAJL_POSSIBLE_LIB_NAMES}; do
        CHECK_FOR_YAJL_AT(${x})
        if test -n "${YAJL_VERSION}"; then
            break
    fi
done
```

Reviewing the definition of `CHECK_FOR_YAJL_AT()` it expects 'path' as the first argumant, and then loops through the `YAJL_POSSIBLE_EXTENSIONS` and `YAJL_POSSIBLE_LIB_NAMES` inside the function to search that path.

```
AC_DEFUN([CHECK_FOR_YAJL_AT], [
    path=$1
    for y in ${YAJL_POSSIBLE_EXTENSIONS}; do
        for z in ${YAJL_POSSIBLE_LIB_NAMES}; do
...
```

By modifying the loop defined on line 65 to loop over `YAJL_POSSIBLE_PATHS` instead of `YAJL_POSSIBLE_LIB_NAMES` and running ./build.sh again, YAJL is found correctly.

Additionally, if multiple directories are provided, for example `--with-yajl="<dir1> <dir2>"`, they are not all searched. Instead they are treated as a single directory. In the yajl.m4 file on line 73 the script checks for `--with-yajl` with an argument and passes the argument to `CHECK_FOR_YAJL_AT()` as a path.  Instead, it would be optimal to loop over provided paths and search each one.

**Logs and dumps**

[...]

PR #3335, opened 2025-01-30
Fix for issue #3334: build not finding YAJL

RooHTaylor:
When searching for YAJL during ./configure, pkg-config is checked first, and then a list of directories is searched if pkg-config bears no fruit. The previous version of yajl.m4 was looping over YAJL_POSSIBLE_LIB_NAMES instead of YAJL_POSSIBLE_PATHS and passing the lib name to the CHECK_FOR_YAJL_AT() function instead of the path. The would lead to YAJL never being found if pkg-config could not find it.

## what

- if pkg-config cannot locate libyajl it will not be located via path search, as the loop is passing lib names as path instead of possible paths.
- change fixes loop definition to loop over paths instead of lib names.

## why

- so that libyajl may still be located if pkg-config fails

## references

https://github.com/owasp-modsecurity/ModSecurity/issues/3334

### DIFF
--- a/build/yajl.m4
+++ b/build/yajl.m4
@@ -62,7 +62,7 @@ else
                 YAJL_DISPLAY="${YAJL_LDADD}, ${YAJL_CFLAGS}"
             else
                 # If pkg-config did not find anything useful, go over file lookup.
-                for x in ${YAJL_POSSIBLE_LIB_NAMES}; do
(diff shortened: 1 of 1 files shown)


airween (approved):
LGTM - thanks, @RooHTaylor!

PR #3345, opened 2025-03-03
V2/refactor/default pcre2

gberkes:
## What

Refactor the build system and the relevant source files to make the PCRE2 library the default build target. If you wish to continue building with the legacy PCRE library, you must explicitly specify it using the following command:
configure --with-pcre.

Additional build system cleanups include:

Removed unnecessary AM_CONDITIONAL lines from configure.ac.

## Why

The legacy PCRE library will be deprecated and eventually removed. The first step in this process is to make PCRE2 the default build dependency, replacing the old PCRE library.


airween:
Closed as completed in #3383. Thanks @gberkes!

PR #3350, opened 2025-03-12
fix: align code to fix cppcheck errors

airween:
## what

This PR aligns some part of code and fix `cppcheck` issues.

## why

`cppcheck` released a new version: 2.17.1. This versions is more strict and reported few new issues.

## references

See PR merge #3321, especially [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3321#issuecomment-2718914899) comment.

### DIFF
--- a/src/actions/transformations/utf8_to_unicode.cc
+++ b/src/actions/transformations/utf8_to_unicode.cc
@@ -46,7 +46,7 @@ static inline bool encode(std::string &value) {
         int unicode_len = 0;
         unsigned int d = 0;
         unsigned char c;
-        auto utf = &input[i];
+        const auto* utf = &input[i];
 
         c = *utf;
 
--- a/examples/multiprocess_c/multi.c
+++ b/examples/multiprocess_c/multi.c
@@ -38,7 +38,7 @@ RulesSet *rules = NULL;
 ModSecurity *modsec = NULL;
 
 
-void process_special_request (int j) {
+static void process_special_request (int j) {
     Transaction *transaction;
     transaction = msc_new_transaction(modsec, rules, NULL);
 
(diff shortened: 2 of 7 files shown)


src/actions/transformations/utf8_to_unicode.cc
```diff
@@ -46,7 +46,7 @@ static inline bool encode(std::string &value) {
         int unicode_len = 0;
         unsigned int d = 0;
         unsigned char c;
-        auto utf = &input[i];
+        const auto* utf = &input[i];
```
theseion, line 49: Does this change the semantics?

theseion (changes requested):
I can't gauge the correctness of your changes, but they look good.

src/actions/transformations/utf8_to_unicode.cc
airween: > Does this change the semantics?

Yes, but that was the intention. Let me explain the reason of the change.
[Here](https://github.com/owasp-modsecurity/ModSecurity/blob/8f00f4700f4a00e4269a73cc26a909557a20eff0/src/actions/transformations/utf8_to_unicode.cc#L31) you can see the type of `input`:
```C++
    auto input = reinterpret_cast<unsigned char*>(value.data());
```
So `value` is an `std::string` reference, and this cast makes `input` as `unsigned char*`, which is a pointer.

In the previous implementation:
```C++
        auto utf = &input[i];
```
`auto` made `utf` as `unsigned char*`, a pointer which points to the `i`-th item's address in `input`.

We don't want to modify the address' of `utf` so we can make it `const`.

If we use this syntax:
```C++
        const auto utf = &str[i];
```
then we get an `unsigned char* const` type, which means we **can change the address but not the value**.

With this syntax:
```C++
        const auto* utf = &str[i];
````
we get an `const unsigned char*` type where we **can't change pointer's address but can change the value**.

And this is what we want (and `cppcheck` reported).

theseion: Thanks!

theseion (approved):

Issue #3178, opened 2024-06-28
Discussion of the new XML processing feature


airween:
**Describe the bug**

It's not a bug but a discussion about a new feature, how can we extend the XML processing.

There is a feature request from a customer that we should extend the engines' XML parsing capability. Of course, we should add this request to both engine with same behavior.

## Current behavior

Consider this payload:

```XML
<?xml version="1.0" encoding="UTF-8"?>
<root>
  <level1>
    <level2>
      <node>foo1</node>
      <node>bar1</node>
    </level2>
    <level2>
      <node>foo2</node>
      <node>bar2</node>
    </level2>
  </level1>
</root>
```

This payload will appear in current state in the engines:

(mod_security2)
```
[/post][9] Target value: "  foo1  bar1  foo2  bar2"
```
(libmodsecurity3)
```
[/post] [9] Target value: "  foo1  bar1  foo2  bar2" (Variable: XML:/*)
```
(lines from debug.logs)

## Problem

The problem is that exclusions for sub-parts and specific nodes does not work. See the example:

```
SecRule XML:/* "@rx ^foo.*" \
	"id:10001,\
	phase:2,\
	t:none,\
	log,\
	pass,\
	ctl:ruleRemoveTargetById=930120;XML:/level1/level2/node"
```

because the XML variable holds the **concatenated node values**, not a key:value pairs like JSON. Therefore it's impossible to create any exclusion against any rules.

## Possible solution

Consider this converted strcture (XML to JSON):

```JSON
{
  "root": {
    "level1": {
      "level2": [
        {
          "node": [
            "foo1",
            "bar1"
          ]
        },
        {
          "node": [
            "foo2",
            "bar2"
          ]
        }
      ]
    }
  }
}
```

This payload will expanded like this:

(mod_security2)
```
[/post][9] Adding JSON argument 'root.level1.level2.level2.node.node' with value 'foo1'
[/post][9] Adding JSON argument 'root.level1.level2.level2.node.node' with value 'bar1'
[/post][9] Adding JSON argument 'root.level1.level2.level2.node.node' with value 'foo2'
[/post][9] Adding JSON argument 'root.level1.level2.level2.node.node' with value 'bar2'
```
(libmodsecurity3)
```
[/post] [4] Adding request argument (JSON): name "json.root.level1.level2.array_0.node.array_0", value "foo1"
[/post] [4] Adding request argument (JSON): name "json.root.level1.level2.array_0.node.array_1", value "bar1"
[...]

PR #3358, opened 2025-04-07
feat: improved XMLArgs processing

airween:
## what

This PR adds a new feature within XML processing.

**Old (current) behavior**: in case of `XML:/*` target the body processor expands the node values from the XML payload. Eg.:
```XML
<?xml version="1.0" encoding="UTF-8"?>
<root>
  <level1>
    <level2>
      <node>foo1</node>
      <node>bar1</node>
    </level2>
    <level2>
      <node>foo2</node>
      <node>bar2</node>
    </level2>
  </level1>
</root>
```
will produce this value:
```
[/post][9] Target value: "  foo1  bar1  foo2  bar2"
```

In this case, there is no option to exclude any node. For example, if a node contains a term that a rule is looking for, the administrator could not create an exclusion. The only solution is to exclude the whole rule.

**New behavior**: there is a new configuration keyword, `SecParseXMLintoArgs` with possible values `On`, `Off` and `OnlyArgs`. The default value is `Off`. This won't change anything. If the administrator set this to `On`, then the engine will parse the XML into `ARGS` **AND** the `XML:/*` target will still contain the only text content as before. If the value is `OnlyArgs` then only the parsed content will appear in `ARGS` target; the `XML:/*` target won't contain the parsed content anymore.

If administrator sets it to `On`, then the node values will appear in `ARGS`, and it's easy to make any exclusion against the named target.

## why

A customer request has been received to solve this.

## references

See #3178.

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -3709,8 +3709,8 @@ static const char *cmd_cache_transformations(cmd_parms *cmd, void *_dcfg,
 * \param _dcfg Pointer to directory configuration
 * \param p1 Pointer to configuration option
 *
-* \retval NULL On failure
-* \retval apr_psprintf On Success
+* \retval NULL On Success
+* \retval apr_psprintf On error
 */
 static const char *cmd_parse_xml_into_args(cmd_parms *cmd, void *_dcfg, const char *p1)
 {
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -166,6 +166,7 @@ void *create_directory_config(apr_pool_t *mp, char *path)
 
     /* xml external entity */
     dcfg->xml_external_entity = NOT_SET;
+    dcfg->parse_xml_into_args = NOT_SET;
 
     return dcfg;
 }
@@ -640,6 +641,8 @@ void *merge_directory_configs(apr_pool_t *mp, void *_parent, void *_child)
     /* xml external entity */
     merged->xml_external_entity = (child->xml_external_entity == NOT_SET
         ? parent->xml_external_entity : child->xml_external_entity);
+    merged->parse_xml_into_args = (child->parse_xml_into_args == NOT_SET
+        ? parent->parse_xml_into_args : child->parse_xml_into_args);
 
     return merged;
 }
@@ -773,6 +776,7 @@ void init_directory_config(directory_config *dcfg)
 
     /* xml external entity */
     if (dcfg->xml_external_entity == NOT_SET) dcfg->xml_external_entity = 0;
+    if (dcfg->parse_xml_into_args == NOT_SET) dcfg->parse_xml_into_args = 0;
 
 }
 
@@ -3698,6 +3702,34 @@ static const char *cmd_cache_transformations(cmd_parms *cmd, void *_dcfg,
     return NULL;
 }
 
+/**
+* \brief Add SecParseXMLIntoArgs configuration option
+*
+* \param cmd Pointer to configuration data
+* \param _dcfg Pointer to directory configuration
+* \param p1 Pointer to configuration option
+*
+* \retval NULL On failure
+* \retval apr_psprintf On Success
+*/
+static const char *cmd_parse_xml_into_args(cmd_parms *cmd, void *_dcfg, const char *p1)
+{
+    assert(cmd != NULL);
+    assert(_dcfg != NULL);
+    assert(p1 != NULL);
+    // Normally useless code, left to be safe for the moment
+    if (_dcfg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_hash_engine: _dcfg is NULL");
+        return NULL;
+    }
+    directory_config *dcfg = (directory_config *)_dcfg;
+    if (strcasecmp(p1, "on") == 0)       { dcfg->parse_xml_into_args = MSC_XML_ARGS_ON; }
+    else if (strcasecmp(p1, "off") == 0) { dcfg->parse_xml_into_args = MSC_XML_ARGS_OFF; }
+    else if (strcasecmp(p1, "onlyargs") == 0) { dcfg->parse_xml_into_args = MSC_XML_ARGS_ONLYARGS; }
+    else return apr_psprintf(cmd->pool, "ModSecurity: Invalid value for SecParseXMLIntoArgs: %s", p1);
+
+    return NULL;
+}
 
 /* -- Configuration directives definitions -- */
 
@@ -4464,6 +4496,14 @@ const command_rec module_directives[] = {
         NULL,
         CMD_SCOPE_ANY,
         "Set Hash parameter"
+    ),
+
+    AP_INIT_TAKE1 (
+        "SecParseXMLintoArgs",
+        cmd_parse_xml_into_args,
+        NULL,
+        CMD_SCOPE_ANY,
+        "On, Off or OnlyArgs"
     ),
 
     { NULL }
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -3719,7 +3719,7 @@ static const char *cmd_parse_xml_into_args(cmd_parms *cmd, void *_dcfg, const ch
     assert(p1 != NULL);
     // Normally useless code, left to be safe for the moment
     if (_dcfg == NULL) {
-        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_hash_engine: _dcfg is NULL");
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_parse_xml_into_args: _dcfg is NULL");
         return NULL;
     }
     directory_config *dcfg = (directory_config *)_dcfg;
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -3703,7 +3703,7 @@ static const char *cmd_cache_transformations(cmd_parms *cmd, void *_dcfg,
 }
 
 /**
-* \brief Add SecParseXMLIntoArgs configuration option
+* \brief Add SecParseXmlIntoArgs configuration option
 *
 * \param cmd Pointer to configuration data
 * \param _dcfg Pointer to directory configuration
@@ -3726,7 +3726,7 @@ static const char *cmd_parse_xml_into_args(cmd_parms *cmd, void *_dcfg, const ch
     if (strcasecmp(p1, "on") == 0)       { dcfg->parse_xml_into_args = MSC_XML_ARGS_ON; }
     else if (strcasecmp(p1, "off") == 0) { dcfg->parse_xml_into_args = MSC_XML_ARGS_OFF; }
     else if (strcasecmp(p1, "onlyargs") == 0) { dcfg->parse_xml_into_args = MSC_XML_ARGS_ONLYARGS; }
-    else return apr_psprintf(cmd->pool, "ModSecurity: Invalid value for SecParseXMLIntoArgs: %s", p1);
+    else return apr_psprintf(cmd->pool, "ModSecurity: Invalid value for SecParseXmlIntoArgs: %s", p1);
 
     return NULL;
 }
@@ -4499,7 +4499,7 @@ const command_rec module_directives[] = {
     ),
 
     AP_INIT_TAKE1 (
-        "SecParseXMLintoArgs",
+        "SecParseXmlIntoArgs",
         cmd_parse_xml_into_args,
         NULL,
         CMD_SCOPE_ANY,
--- a/apache2/modsecurity.h
+++ b/apache2/modsecurity.h
@@ -243,6 +243,10 @@ extern DSOLOCAL int *unicode_map_table;
 #define RULE_EXCEPTION_REMOVE_MSG       4
 #define RULE_EXCEPTION_REMOVE_TAG       5
 
+#define MSC_XML_ARGS_OFF                0
+#define MSC_XML_ARGS_ON                 1
+#define MSC_XML_ARGS_ONLYARGS           2
+
 #define NBSP                            160
 
 struct rule_exception {
@@ -643,6 +647,7 @@ struct directory_config {
 
     /* xml */
     int                 xml_external_entity;
+    int                 parse_xml_into_args;
 
     /* This will be used whenever ModSecurity will be ready
      * to ask the server for newer rules.
--- a/apache2/msc_xml.c
+++ b/apache2/msc_xml.c
@@ -14,6 +14,108 @@
 
 #include "msc_xml.h"
 
+static void msc_xml_on_start_elementns(
+    void *ctx,
+    const xmlChar *localname,
+    const xmlChar *prefix,
+    const xmlChar *URI,
+    int nb_namespaces,
+    const xmlChar **namespaces,
+    int nb_attributes,
+    int nb_defaulted,
+    const xmlChar **attributes
+) {
+
+    // get the length of XML tag (localname)
+    size_t taglen = strlen((const char *)localname);
+    modsec_rec * msr = (modsec_rec *)ctx;
+    msc_xml_parser_state * xml_parser_state = msr->xml->xml_parser_state;
+
+    // pathlen contains the concatenated strings of tags with '.'
+    // eg xml.root.level1.leaf
+    xml_parser_state->pathlen += (taglen + 1);
+    char *newpath = apr_pstrcat(msr->mp, xml_parser_state->currpath, ".", (char *)localname, NULL);
+    xml_parser_state->currpath = newpath;
+
+    int *new_stack_item = (int *)apr_array_push(xml_parser_state->has_child_stack);
+    *new_stack_item = 0;
+    xml_parser_state->depth++;
+
+    // if there is an item before the current one we set that has a child
+    if (xml_parser_state->depth > 1) {
+        int *parent_stack_item = &((int *)xml_parser_state->has_child_stack->elts)[xml_parser_state->has_child_stack->nelts - 2];
+        *parent_stack_item = 1;
+    }
+
+}
+
+static void msc_xml_on_end_elementns(
+    void* ctx,
+    const xmlChar* localname,
+    const xmlChar* prefix,
+    const xmlChar* URI
+) {
+
+    size_t taglen = strlen((const char *)localname);
+    modsec_rec * msr = (modsec_rec *)ctx;
+    msc_xml_parser_state * xml_parser_state = msr->xml->xml_parser_state;
+
+    // if the node is a leaf we add it as argument
+    // get the top item from the stack which tells this info
+    int * top_stack_item = apr_array_pop(xml_parser_state->has_child_stack);
+    if (*top_stack_item == 0) {
+
+        if (apr_table_elts(msr->arguments)->nelts >= msr->txcfg->arguments_limit) {
+            if (msr->txcfg->debuglog_level >= 4) {
+                msr_log(msr, 4, "Skipping request argument, over limit (XML): name \"%s\", value \"%s\"",
+                        log_escape_ex(msr->mp, xml_parser_state->currpath, strlen(xml_parser_state->currpath)),
+                        log_escape_ex(msr->mp, xml_parser_state->currval, strlen(xml_parser_state->currval)));
+            }
+            msr->msc_reqbody_error = 1;
+            msr->xml->xml_error = apr_psprintf(msr->mp, "More than %ld XML keys", msr->txcfg->arguments_limit);
+            xmlStopParser((xmlParserCtxtPtr)msr->xml->parsing_ctx_arg);
+        }
+        else {
+
+            msc_arg * arg = (msc_arg *) apr_pcalloc(msr->mp, sizeof(msc_arg));
+
+            arg->name = xml_parser_state->currpath;
+            arg->name_len = strlen(arg->name);
+            arg->value = xml_parser_state->currval;
+            arg->value_len = strlen(xml_parser_state->currval);
+            arg->value_origin_len = arg->value_len;
+            //arg->value_origin_offset = value-base_offset;
+            arg->origin = "XML";
+
+            if (msr->txcfg->debuglog_level >= 9) {
+                msr_log(msr, 9, "Adding XML argument '%s' with value '%s'",
+                    xml_parser_state->currpath, xml_parser_state->currval);
+            }
+
+            apr_table_addn(msr->arguments,
+                log_escape_nq_ex(msr->mp, arg->name, arg->name_len), (void *) arg);
+        } // end else
+    } // end top_stack_item == 0
+
+    // decrease the length of current path length - +1 because of the '\0'
+    xml_parser_state->pathlen -= (taglen + 1);
+
+    // -1 need because we don't need the '.'
+    char * newpath = apr_pstrndup(msr->mp, xml_parser_state->currpath, xml_parser_state->pathlen - 1);
+    xml_parser_state->currpath = newpath;
+
+    xml_parser_state->depth--;
+}
+
+static void msc_xml_on_characters(void *ctx, const xmlChar *ch, int len) {
+
+    modsec_rec * msr = (modsec_rec *)ctx;
+    msc_xml_parser_state * xml_parser_state = msr->xml->xml_parser_state;
+
+    xml_parser_state->currval = apr_pstrndup(msr->mp, (const char *)ch, len);
+}
+
+
 static xmlParserInputBufferPtr
 xml_unload_external_entity(const char *URI, xmlCharEncoding enc)    {
     return NULL;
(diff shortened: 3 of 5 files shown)


RedXanadu:
This is a great new feature. This will open up ModSecurity to anyone who needs to do serious processing of XML APIs (lots of legacy and current applications!). Especially with pre-written rule sets like CRS, this makes the task of handling false positives possible.

Thank you for the work that has gone into this :rocket:

dune73:
@airween Could you share how the new option parses / advertises multi-level documents with multiple leaves carrying the same name? Is the hierarchy part of the name or is that hidden?

Needless to say, that I really like this option.

airween:
@dune73,

I hope I understand your question as well :smile:, so consider this file:
```bash
cat test.xml
<?xml version="1.0" encoding="UTF-8"?>
<root>
  <level1>
    <level2>
      <node>foo1</node>
      <node>bar1</node>
    </level2>
    <level2>
      <node>foo2</node>
      <node>bar2</node>
    </level2>
  </level1>
  <level1>
    <level2>
      <node>foo1</node>
      <node>bar1</node>
    </level2>
    <level2>
      <node>foo2</node>
      <node>bar2</node>
    </level2>
  </level1>
</root>
```
and this request:
```bash
curl -v -H "Content-Type: application/xml" -X POST -d @test.xml http://localhost/post.php
```

This will generates these arguments (it's totally the same as in case of JSON):
```
Adding XML argument 'xml.root.level1.level2.node' with value 'foo1'
Adding XML argument 'xml.root.level1.level2.node' with value 'bar1'
Adding XML argument 'xml.root.level1.level2.node' with value 'foo2'
Adding XML argument 'xml.root.level1.level2.node' with value 'bar2'
Adding XML argument 'xml.root.level1.level2.node' with value 'foo1'
Adding XML argument 'xml.root.level1.level2.node' with value 'bar1'
Adding XML argument 'xml.root.level1.level2.node' with value 'foo2'
Adding XML argument 'xml.root.level1.level2.node' with value 'bar2'
Expanded "REQUEST_URI_RAW|REQUEST_HEADERS|ARGS|ARGS_NAMES" to "REQUEST_URI_RAW|REQUEST_HEADERS:Host|REQUEST_HEADERS:User-Agent|REQUEST_HEADERS:Accept|REQUEST_HEADERS:Content-Type|REQUEST_HEADERS:Content-Length|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node|ARGS_NAMES:xml.root.level1.level2.node".
```

dune73:
This was what I expected. Thanks for the confirmation. Very good.

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+    void *ctx,
+    const xmlChar *localname,
+    const xmlChar *prefix,
+    const xmlChar *URI,
+    int nb_namespaces,
+    const xmlChar **namespaces,
+    int nb_attributes,
+    int nb_defaulted,
+    const xmlChar **attributes
+) {
```
fzipi, line 27: Is this new formatting for the code? Or was adopted as standard already?

Having mixed format in parameters, in general makes more difficult reading the code. So my suggestion will be:
- use the same format as all other files
- propose a new standard
- once accepted, apply to all files once and for all
- enforce the new format in the pipeline.

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+    }
+
+}
+
+static void msc_xml_on_end_elementns(
+    void* ctx,
+    const xmlChar* localname,
+    const xmlChar* prefix,
+    const xmlChar* URI
+) {
```
fzipi, line 61: Same here.

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+        else {
+
+            msc_arg * arg = (msc_arg *) apr_pcalloc(msr->mp, sizeof(msc_arg));
+
+            arg->name = xml_parser_state->currpath;
+            arg->name_len = strlen(arg->name);
+            arg->value = xml_parser_state->currval;
+            arg->value_len = strlen(xml_parser_state->currval);
+            arg->value_origin_len = arg->value_len;
+            //arg->value_origin_offset = value-base_offset;
```
fzipi:
```suggestion
```

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+
+            apr_table_addn(msr->arguments,
+                log_escape_nq_ex(msr->mp, arg->name, arg->name_len), (void *) arg);
+        } // end else
+    } // end top_stack_item == 0
+
+    // decrease the length of current path length - +1 because of the '\0'
+    xml_parser_state->pathlen -= (taglen + 1);
+
+    // -1 need because we don't need the '.'
```
fzipi:
```suggestion
    // -1 is needed because we don't need the last '.'
```

apache2/msc_xml.c
```diff
@@ -86,18 +215,52 @@ int xml_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
...
+                NULL);
+            if (msr->xml->parsing_ctx_arg == NULL) {
+                *error_msg = apr_psprintf(msr->mp, "XML: Failed to create parsing context for ARGS.");
+                return -1;
+            }
         }
     } else {
 
         /* Not a first invocation. */
+        msr_log(msr, 4, "XML: Continue parsing.");
```
fzipi, line 256: What does even "Continue parsing" means?

apache2/msc_xml.h
```diff
@@ -20,15 +20,36 @@ typedef struct xml_data xml_data;
...
     xmlSAXHandler           *sax_handler;
     xmlParserCtxtPtr        parsing_ctx;
     xmlDocPtr               doc;
 
     unsigned int            well_formed;
+
+    /* error reporting and XML array flag */
+    char                   *xml_error;
+
+    /* another parser context for arguments */
```
fzipi:
```suggestion
    /* additional parser context for arguments */
```

fzipi (commented):
No tests?

apache2/apache2_config.c
```diff
@@ -3698,6 +3702,34 @@ static const char *cmd_cache_transformations(cmd_parms *cmd, void *_dcfg,
...
+* \retval apr_psprintf On Success
+*/
+static const char *cmd_parse_xml_into_args(cmd_parms *cmd, void *_dcfg, const char *p1)
+{
+    assert(cmd != NULL);
+    assert(_dcfg != NULL);
+    assert(p1 != NULL);
+    // Normally useless code, left to be safe for the moment
+    if (_dcfg == NULL) {
+        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_hash_engine: _dcfg is NULL");
```
theseion:
```suggestion
        ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_parse_xml_into_args: _dcfg is NULL");
```

apache2/apache2_config.c
```diff
@@ -3698,6 +3702,34 @@ static const char *cmd_cache_transformations(cmd_parms *cmd, void *_dcfg,
...
 
+/**
+* \brief Add SecParseXMLIntoArgs configuration option
+*
+* \param cmd Pointer to configuration data
+* \param _dcfg Pointer to directory configuration
+* \param p1 Pointer to configuration option
+*
+* \retval NULL On failure
+* \retval apr_psprintf On Success
```
theseion: I don't think this is correct. In one error case you return `NULL`, in the other `apr_psprintf`. In the second case `NULL` is the success value, if I'm not mistaken.

apache2/msc_xml.c
theseion: What do you mean by "new"? AFAICT, there are already multiple different formats for long function signatures (e.g., `sec_guarstatic void dian_logger`, `static void copy_rules_phase`).

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+    size_t taglen = strlen((const char *)localname);
+    modsec_rec * msr = (modsec_rec *)ctx;
+    msc_xml_parser_state * xml_parser_state = msr->xml->xml_parser_state;
+
+    // if the node is a leaf we add it as argument
+    // get the top item from the stack which tells this info
+    int * top_stack_item = apr_array_pop(xml_parser_state->has_child_stack);
+    if (*top_stack_item == 0) {
+
+        if (apr_table_elts(msr->arguments)->nelts >= msr->txcfg->arguments_limit) {
```
theseion, line 72: `arugments_limit` is probably the limit for all args. The number of args will be `POST + GET + XML`. Another reason to put XML into a separate target variable, IMO.

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+            if (msr->txcfg->debuglog_level >= 4) {
+                msr_log(msr, 4, "Skipping request argument, over limit (XML): name \"%s\", value \"%s\"",
+                        log_escape_ex(msr->mp, xml_parser_state->currpath, strlen(xml_parser_state->currpath)),
+                        log_escape_ex(msr->mp, xml_parser_state->currval, strlen(xml_parser_state->currval)));
+            }
+            msr->msc_reqbody_error = 1;
+            msr->xml->xml_error = apr_psprintf(msr->mp, "More than %ld XML keys", msr->txcfg->arguments_limit);
```
theseion: Does the limit apply to all args? Or just the values for the XML target? In the former case, this sentence would be wrong, since the number would be made up of POST + GET + XML.

apache2/msc_xml.c
```diff
@@ -14,6 +14,108 @@
...
+                    xml_parser_state->currpath, xml_parser_state->currval);
+            }
+
+            apr_table_addn(msr->arguments,
+                log_escape_nq_ex(msr->mp, arg->name, arg->name_len), (void *) arg);
+        } // end else
+    } // end top_stack_item == 0
+
+    // decrease the length of current path length - +1 because of the '\0'
+    xml_parser_state->pathlen -= (taglen + 1);
```
theseion, line 104: Wait... wouldn't that leave a `.` at the end? Which you are then stripping again below? Couldn't you just do `... -= taglen - 2`?

apache2/msc_xml.c
```diff
@@ -86,18 +215,52 @@ int xml_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
 
         */
 
-        msr->xml->parsing_ctx = xmlCreatePushParserCtxt(NULL, NULL, buf, size, "body.xml");
-        if (msr->xml->parsing_ctx == NULL) {
-            *error_msg = apr_psprintf(msr->mp, "XML: Failed to create parsing context.");
-            return -1;
+        if (msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_ONLYARGS) {
```
theseion, line 234: This will also apply to `MSC_XML_ARGS_OFF`. Is that correct?

apache2/msc_xml.c
```diff
@@ -86,18 +215,52 @@ int xml_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
...
+                *error_msg = apr_psprintf(msr->mp, "XML: Failed to create parsing context for ARGS.");
+                return -1;
+            }
         }
     } else {
 
         /* Not a first invocation. */
+        msr_log(msr, 4, "XML: Continue parsing.");
+        if (msr->xml->parsing_ctx != NULL &&
+            msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_ONLYARGS) {
```
theseion, line 258: This will also apply to `MSC_XML_ARGS_OFF`. Is that correct?

apache2/msc_xml.c
```diff
@@ -86,18 +215,52 @@ int xml_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
...
+            xmlParseChunk(msr->xml->parsing_ctx, buf, size, 0);
+            if (msr->xml->parsing_ctx->wellFormed != 1) {
+                *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document.");
```
theseion:
```suggestion
                *error_msg = apr_psprintf(msr->mp, "XML: Failed to parse document.");
```

apache2/msc_xml.c
```diff
@@ -86,18 +215,52 @@ int xml_process_chunk(modsec_rec *msr, const char *buf, unsigned int size, char
...
-        if (msr->xml->parsing_ctx->wellFormed != 1) {
-            *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document.");
+        if (msr->xml->parsing_ctx_arg != NULL &&
+            msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_OFF) {
+            if (xmlParseChunk(msr->xml->parsing_ctx_arg, buf, size, 0) != 0) {
+                if (msr->xml->xml_error) {
+                    *error_msg = msr->xml->xml_error;
+                }
+                else {
+                    *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document for ARGS.");
```
theseion:
```suggestion
                    *error_msg = apr_psprintf(msr->mp, "XML: Failed to parse document for ARGS.");
```

apache2/msc_xml.c
```diff
@@ -114,23 +277,44 @@ int xml_complete(modsec_rec *msr, char **error_msg) {
     *error_msg = NULL;
 
     /* Only if we have a context, meaning we've done some work. */
-    if (msr->xml->parsing_ctx != NULL) {
-        /* This is how we signalise the end of parsing to libxml. */
-        xmlParseChunk(msr->xml->parsing_ctx, NULL, 0, 1);
+    if (msr->xml->parsing_ctx != NULL || msr->xml->parsing_ctx_arg != NULL) {
+        if (msr->xml->parsing_ctx != NULL &&
+            msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_ONLYARGS) {
+            /* This is how we signalise the end of parsing to libxml. */
```
theseion:
```suggestion
            /* This is how we signal the end of parsing to libxml. */
```

apache2/msc_xml.c
```diff
@@ -114,23 +277,44 @@ int xml_complete(modsec_rec *msr, char **error_msg) {
...
+            /* Clean up everything else. */
+            xmlFreeParserCtxt(msr->xml->parsing_ctx);
+            msr->xml->parsing_ctx = NULL;
+            msr_log(msr, 4, "XML: Parsing complete (well_formed %u).", msr->xml->well_formed);
 
-        if (msr->xml->well_formed != 1) {
-            *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document.");
-            return -1;
+            if (msr->xml->well_formed != 1) {
+                *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document.");
```
theseion:
```suggestion
                *error_msg = apr_psprintf(msr->mp, "XML: Failed to parse document.");
```

apache2/msc_xml.c
```diff
@@ -114,23 +277,44 @@ int xml_complete(modsec_rec *msr, char **error_msg) {
     *error_msg = NULL;
 
     /* Only if we have a context, meaning we've done some work. */
-    if (msr->xml->parsing_ctx != NULL) {
-        /* This is how we signalise the end of parsing to libxml. */
-        xmlParseChunk(msr->xml->parsing_ctx, NULL, 0, 1);
+    if (msr->xml->parsing_ctx != NULL || msr->xml->parsing_ctx_arg != NULL) {
+        if (msr->xml->parsing_ctx != NULL &&
+            msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_ONLYARGS) {
```
theseion, line 298: This will also apply to `MSC_XML_ARGS_OFF`. Is that correct?

apache2/msc_xml.c
```diff
@@ -114,23 +277,44 @@ int xml_complete(modsec_rec *msr, char **error_msg) {
...
+        }
+
+        if (msr->xml->parsing_ctx_arg != NULL &&
+            msr->txcfg->parse_xml_into_args != MSC_XML_ARGS_OFF) {
+            if (xmlParseChunk(msr->xml->parsing_ctx_arg, NULL, 0, 1) != 0) {
+                if (msr->xml->xml_error) {
+                    *error_msg = msr->xml->xml_error;
+                }
+                else {
+                    *error_msg = apr_psprintf(msr->mp, "XML: Failed parsing document for ARGS.");
```
theseion:
```suggestion
                    *error_msg = apr_psprintf(msr->mp, "XML: Failed to parse document for ARGS.");
```

apache2/re_actions.c
```diff
@@ -1022,6 +1022,13 @@ static char *msre_action_ctl_validate(msre_engine *engine, apr_pool_t *mp, msre_
         if (strcasecmp(value, "on") == 0) return NULL;
         if (strcasecmp(value, "off") == 0) return NULL;
         return apr_psprintf(mp, "Invalid setting for ctl name HashEngine: %s", value);
+     }
+     else
+        if (strcasecmp(name, "parseXMLintoArgs") == 0) {
```
theseion:
```suggestion
        if (strcasecmp(name, "parseXmlIntoArgs") == 0) {
```

apache2/re_actions.c
```diff
@@ -1022,6 +1022,13 @@ static char *msre_action_ctl_validate(msre_engine *engine, apr_pool_t *mp, msre_
...
+        if (strcasecmp(value, "on") == 0) return NULL;
+        if (strcasecmp(value, "off") == 0) return NULL;
+        if (strcasecmp(value, "onlyargs") == 0) return NULL;
+        return apr_psprintf(mp, "Invalid setting for ctl name parseXMLintoArgs: %s", value);
```
theseion:
```suggestion
        return apr_psprintf(mp, "Invalid setting for ctl name parseXmlIntoArgs: %s", value);
```

apache2/re_actions.c
```diff
@@ -1377,6 +1384,29 @@ static apr_status_t msre_action_ctl_execute(modsec_rec *msr, apr_pool_t *mptmp,
             return -1;
         }
         apr_table_addn(msr->removed_targets, apr_pstrdup(msr->mp, p2), (void *)re);
+        return 1;
+    } else
+    if (strcasecmp(name, "parseXMLintoArgs") == 0) {
```
theseion:
```suggestion
    if (strcasecmp(name, "parseXmlIntoArgs") == 0) {
```

apache2/modsecurity.h
```diff
@@ -243,6 +243,10 @@ extern DSOLOCAL int *unicode_map_table;
 #define RULE_EXCEPTION_REMOVE_MSG       4
 #define RULE_EXCEPTION_REMOVE_TAG       5
 
+#define MSC_XML_ARGS_OFF                0
+#define MSC_XML_ARGS_ON                 1
+#define MSC_XML_ARGS_ONLYARGS           2
```
theseion, line 248: I don't understand why you need three states here. AFAICT from the code, "onlyargs" is equivalent to "on".

theseion (changes requested):

airween:
> No tests?

I know we should start at some point to add tests, but unfortunately almost all of directives and features do not have tests. In project [MRTS](https://github.com/owasp-modsecurity/MRTS) I really hope once we will arrive that point where we can cover all directives, arguments, operators and actions. But not yet.

NB: in case of libmodsecurity3 implementation there are many new tests, but it's more easier.

apache2/msc_xml.c
airween: Sorry, could you show me an example what can I assume as "standard" already?

apache2/msc_xml.c
airween: This means the engine called this function but not first and we continue the parsing.

apache2/apache2_config.c
airween: You're right. I assume you think about [this](https://github.com/airween/ModSecurity/blob/v2/xmlargsfeat/apache2/apache2_config.c#L3720-L3724) condition (with its comment). Please check that comment: `// Normally useless code, left to be safe for the moment`. It exists in all directive handle fn, I didn't want to remove that.

So yes, you're right, but if once we will remove those conditions, then the comment will be true.

apache2/msc_xml.c
airween: We can do that later, IMO. Now the plan in #3178 was this, and I explained the intention about this in my first comment. In case of JSON the JSON nodes increase this counter. Therefore I followed this behavior.

Btw in case of XML payload I'm not sure there can be any POST arguments, probably from GET (from query string).

apache2/msc_xml.c
airween: Probably you're right. But as I wrote above, if the request is an XML, there won't be any POST argument, only GET. I wouldn't touch it at this moment.

apache2/msc_xml.c
airween: Yes, exactly. That's how it must behave.

If the user set the `SecParseXMLintoArgs` to `Off` (which is the default), then we definitely **need** this context.
Also, if the user set the directive to `On`, then it means user wants to parse XML nodes into ARGS **plus** he wants to see them under `XML` target too - therefore we definitely need this context too.

If the user set the directive to `OnlyArg`, then we **don't need** this context.

apache2/msc_xml.c
airween: Yes, same as I explained it [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3358#discussion_r2061536318).

apache2/modsecurity.h
airween: No, it's definitely not equivalent.

`Off` means nothing changes, the feature is turned off.
`On` means XML nodes will be parsed into `ARGS` **AND** also will appear under `XML`.
`OnlyArgs` means XML nodes will be parsed into `ARGS`, **AND** `XML` will be empty.

apache2/msc_xml.c
airween: Yes, same as [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3358#discussion_r2061536318) comment.

apache2/msc_xml.c
airween: If I used `... -= taglen - 2`, then the `-1` wouldn't be needed on the next line, but `pathlen` would have to be recalculated later, because the necessary length is what we get with the current calculation. This only sets the exact length of the string once, which I think is more transparent.

apache2/modsecurity.h
theseion: Ok. What is the idea behind the third option? Does it reduce the overhead of populating XML? Is there another reason?

Maybe add comments to the three variables that explain this. At least to me, the meanings and consequences of those variables are not at all obvious.

apache2/apache2_config.c
theseion: No, even if you remove it, the function will return `NULL` on _success_, and `apr_psprintf` on _error_, which is the opposite of what the comment says.

apache2/msc_xml.c
theseion: Ok. But the comment would still be wrong. How about 
```suggestion
            msr->xml->xml_error = apr_psprintf(msr->mp, "More than %ld ARGS (GET + XML)", msr->txcfg->arguments_limit);
```

theseion:
> I know we should start at some point to add tests, but unfortunately almost all of directives and features do not have tests. In project [MRTS](https://github.com/owasp-modsecurity/MRTS) I really hope once we will arrive that point where we can cover all directives, arguments, operators and actions. But not yet.
> 
> NB: in case of libmodsecurity3 implementation there are many new tests, but it's more easier.

I have only a rough working knowledge of C/C++, so it's very likely that I missed stuff during the review. Knowing how many mistakes I make in languages that I'm proficient in I feel rather uncomfortable saying that code is good. Tests would give me much more confidence. I do understand your situation though. Ultimately, it's up to you (you would probably have to carry the weight of bug fixes and user inquiries).

airween:
> I have only a rough working knowledge of C/C++, so it's very likely that I missed stuff during the review. Knowing how many mistakes I make in languages that I'm proficient in I feel rather uncomfortable saying that code is good. Tests would give me much more confidence. I do understand your situation though. Ultimately, it's up to you (you would probably have to carry the weight of bug fixes and user inquiries).

I completely agree with you. Making a review without tests is very hard.

Fortunately the two codes (this and #3363) is _almost_ the same, but yes, we can't declare that they are the same, and the other PR's test cases cover this PR too.

My problem is in this case that there are many-many tests are missing. Adding a unique test against this PR just makes the CI more complicated. The solution is that we find out how can we make a framework when we can write tests against directives and their arguments. (This is why I started MRTS project.)

apache2/apache2_config.c
airween: Ah, now I see - sorry and thank you. Fixed in c11bd6c.

apache2/msc_xml.c
airween: Agree.

apache2/modsecurity.h
airween: > What is the idea behind the third option? Does it reduce the overhead of populating XML? Is there another reason?

In the discussion thread (#3358) I explained the situation: `XML` holds values of XML nodes (see the example). Consider a rule which checks a pattern in given target:
```
SecRule ARGS|ARGS_NAMES|XML:/ "@rx attack" ...
```
and one of the nodes value contains this pattern. What you can do in this case?

You will make an exclusion against the whole rule, or rule's XML target. But then you will exclude the whole check, so the other nodes will be excluded too. The main problem with the current XML examination is that the target holds all of the nodes, you can't examine the nodes one by one. The customer who asked this feature uses XML API, and the login method works this way too. So imagine the XML:
```xml
<?xml version="1.0" encoding="UTF-8"?>
<request>
    <transaction>changepasswd</transaction>
    <username>USERNAME</username>
    <password>PASSWORD</password>
    <node01>USERNAME</node01>
    <node03>NEWPASSWORD</node03>
    <node04>NEWPASSWORD</node04>
    <node05>0</node05>
</request>
```
Now if the `password` or the `newpassword` contains an injection, then the WAF will block the request.

With the third option you can completely turn off the populating of `XML`, and you don't need to exclude it at every rules which uses `XML` target.

> Maybe add comments to the three variables that explain this. At least to me, the meanings and consequences of those variables are not at all obvious.

Here I also tried to follow the conventions. There aren't any comments which explain the meanings of macro's. I think if anyone understands the point of this feature, the names are clear. But of course, if the PR will be merged, I'm going to add the whole feature documentation into the [Wiki](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29) page.

apache2/msc_xml.c
fzipi: 🤷 If it is clear to you then ✅

apache2/modsecurity.h
theseion: I see. Thanks for the explanation.

theseion (dismissed):

airween:
I added two new commits:
* 4c043a0 - changed the directive `SecXMLintoArgs` to `SecXmlIntoArgs`, as we discussed [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3363#discussion_r2062688492)
* 19b7e98 - there is a small change in the node value's parsing: if the node value contains a multi-byte character the SAX calls the function multiple times, so we need to concatenate those sub-strings, not creating a new string every time

apache2/msc_xml.c
```diff
@@ -40,6 +40,10 @@ static void msc_xml_on_start_elementns(
     int *new_stack_item = (int *)apr_array_push(xml_parser_state->has_child_stack);
     *new_stack_item = 0;
     xml_parser_state->depth++;
+    // set null to the current value
```
theseion:
```suggestion
    // set the current value to null
```

theseion (changes requested):

theseion (approved):

apache2/msc_xml.c
fzipi: Ok, we discussed this in v2. It is just a matter of creating the work.

fzipi (approved):
Let's go with the flow.

PR #3359, opened 2025-04-15
build/*.m4: Remove Bashisms/POSIX-ify the M4 helper macros

BalkanMadman:
This commit makes it possible to build ModSecurity on systems where /bin/sh is a POSIX-compliant shell that is not Bash. Debian, Alpine Linux, and Gentoo Linux with the system shell set to not Bash, are examples of such systems.

Previously, the helper macros contained two types of Bashisms:
 

- The `==` comparison operator. Very easy to change, as the proper POSIX-compliant form is `=`. For example:

```
if test "${var}" == "myvalue"  ->  if test "${var}" = "myvalue"
```

- The '-a' (and) operator in the 'test' builtin. The '-a' and '-o' operators were removed in POSIX 2024 (Issue 8). The correct form is to use the '&&' and '||' operators respectively. For instance:
```
if test -d "${var}" -a -r "${var}/file"  ->  if test -d "${var}" && test -r "${var}/file"
```

Bug: https://bugs.gentoo.org/887135

## what

- The Bash `==` operator changed to the POSIX-compliant `=`.
- The obsolete and removed in POSIX 2024 `-a` operator changed to the compliant `&&`.

## why

- Unable to build ModSecurity on systems where `/bin/sh` is not Bash (is dash, ash, etc.). See the Gentoo bug below.

## references

- The Gentoo bug: https://bugs.gentoo.org/887135

### DIFF
--- a/build/pcre.m4
+++ b/build/pcre.m4
@@ -21,7 +21,7 @@ AC_ARG_WITH(
     [test_paths="${with_pcre}"],
     [test_paths="/usr/local/libpcre /usr/local/pcre /usr/local /opt/libpcre /opt/pcre /opt /usr /opt/local"])
 
-if test "x${with_pcre}" == "x" && test "x${with_pcre}" != "xno"; then
+if test "x${with_pcre}" = "x" && test "x${with_pcre}" != "xno"; then
     AC_MSG_NOTICE([Support for pcre not requested; omitting check for pcre])
 else
 
@@ -39,7 +39,7 @@ else
 
         for x in ${test_paths}; do
             dnl # Determine if the script was specified and use it directly
-            if test ! -d "$x" -a -e "$x"; then
+            if test ! -d "$x" && test -e "$x"; then
                 PCRE_CONFIG=$x
                 pcre_path="no"
                 break
--- a/build/curl.m4
+++ b/build/curl.m4
@@ -25,7 +25,7 @@ AC_MSG_CHECKING([for libcurl config script])
 
 for x in ${test_paths}; do
     dnl # Determine if the script was specified and use it directly
-    if test ! -d "$x" -a -e "$x"; then
+    if test ! -d "$x" && test -e "$x"; then
         CURL_CONFIG=$x
         curl_path="no"
         break
@@ -100,7 +100,7 @@ AC_SUBST(CURL_LDFLAGS)
 AC_SUBST(CURL_LDADD)
 AC_SUBST(CURL_USES_GNUTLS)
 
-if test "x${with_curl}" == "xno"; then
+if test "x${with_curl}" = "xno"; then
     CURL_DISABLED=yes
 else
     if test "x${with_curl}" != "x"; then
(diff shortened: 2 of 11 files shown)


build/pcre.m4
```diff
@@ -21,7 +21,7 @@ AC_ARG_WITH(
     [test_paths="${with_pcre}"],
     [test_paths="/usr/local/libpcre /usr/local/pcre /usr/local /opt/libpcre /opt/pcre /opt /usr /opt/local"])
 
-if test "x${with_pcre}" == "x" && test "x${with_pcre}" != "xno"; then
+if test "x${with_pcre}" = "x" && test "x${with_pcre}" != "xno"; then
```
airween, line 24: Please take a look at [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3321#discussion_r2076758023) comment from @theseion. I think he is right, and (unfortunately) we should review all complex conditions.

Moreover, I don't know: do we want to handle the `yes` value in case of all `--with...` options?

Here are the output of different scenarios (tested with these expressions):
```
if test "x${with_pcre}" = "x" && test "x${with_pcre}" != "xno"; then echo "TRUE"; else echo "FALSE";
if test "x${with_pcre}" = "x" || test "x${with_pcre}" != "xno"; then echo "TRUE"; else echo "FALSE";
```

| Value of `with_pcre`  | expected output | real output | meanings                                                                          |
| ----------------------------- | ------------------------ | -----------------|---------------------------------------------------------------------------|
| **condition:** `&&`  |                                |                      |                                                                                             |
| `""` (empty)                 | TRUE                     | TRUE            | `--with-pcre` - we want to use PCRE                             |
| `yes`                             | TRUE                     | **FALSE!** | `--with-pcre=yes` - we want to use PCRE                     |
| `no`                              | FALSE                    | FALSE           | `--with-pcre=no` - we **do not** want to use PCRE |
|                                      |                                |                      |                                                                                             |
| **condition:** `\|\|` |                                |                      |                                                                                             |
| `""` (empty)                 | TRUE                     | TRUE            | `--with-pcre` - we want to use PCRE                             |
| `yes`                             | TRUE                     | TRUE            | `--with-pcre=yes` - we want to use PCRE                     |
| `no`                              | FALSE                    | FALSE           | `--with-pcre=no` - we do not want to use PCRE         |

airween:
Hi @BalkanMadman,

first of all, many thanks for this PR. I think this is a really important patch, we must apply this.

Please take a look at my comment above - I don't expect you to review all conditions, but that would be a huge help for the project.

Also, could you take a look at the [v2/master](https://github.com/owasp-modsecurity/ModSecurity/tree/v2/master) tree too, and make this modifications?

build/pcre.m4
BalkanMadman: So, essentially, the only form of this `--with-pcre` option that should disable pcre is `--with-pcre=no` (`--without-pcre` desugars to `--with-pcre=no`), am I right? Any other forms of `--with-pcre=blabla` should be treated the same way as plain `--with-pcre`, correct? If so, I will change this PR accordingly.

PR #3363, opened 2025-04-20
feat: improved XMLArgs processing

airween:
## what

This PR adds a new feature within XML processing. It's same as the #3358 but for v3.

## why

See v2 patch.

## references

See #3178 and #3358.


src/actions/ctl/parse_xml_into_args.cc
```diff
@@ -0,0 +1,63 @@
...
+ * ModSecurity, http://www.modsecurity.org/
+ * Copyright (c) 2025 OWASP ModSecurity project
+ *
+ * You may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * If any of the files related to licensing are missing or if you have any
+ * other questions related to licensing please contact Trustwave Holdings, Inc.
```
fzipi: No, I don't think they need to contact Trustwave Holdings Inc., right?

src/actions/ctl/parse_xml_into_args.cc
```diff
@@ -0,0 +1,63 @@
...
+#include "modsecurity/rules_set.h"
+#include "modsecurity/transaction.h"
+
+namespace modsecurity {
+namespace actions {
+namespace ctl {
+
+
+bool ParseXmlIntoArgs::init(std::string *error) {
+    std::string what(m_parser_payload, 17, m_parser_payload.size() - 17);
```
fzipi, line 31: 17? 🤦 It looks like the lenght of the `SecParseXMLIntoArgs`, but alone is so arbitrary.

src/actions/ctl/parse_xml_into_args.h
```diff
@@ -0,0 +1,48 @@
...
+ * ModSecurity, http://www.modsecurity.org/
+ * Copyright (c) 2025 OWASP ModSecurity Project
+ *
+ * You may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * If any of the files related to licensing are missing or if you have any
+ * other questions related to licensing please contact Trustwave Holdings, Inc.
```
fzipi: Same here. Trustwave?

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,122 @@ namespace RequestBodyProcessor {
...
+extern "C" {
+    void MSC_startElement(void *userData,
+        const xmlChar *name,
+        const xmlChar *prefix,
+        const xmlChar *URI,
+        int nb_namespaces,
+        const xmlChar **namespaces,
+        int nb_attributes,
+        int nb_defaulted,
+        const xmlChar **attributes) {
```
fzipi, line 130: Same here on this declaration. Is this the new standard?

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,122 @@ namespace RequestBodyProcessor {
...
+
+            MSCSAXHandler* handler = static_cast<MSCSAXHandler*>(userData);
+            handler->onStartElement(userData, name);
+    }
+
+    void MSC_endElement(
+        void *userData,
+        const xmlChar *name,
+        const xmlChar* prefix,
+        const xmlChar* URI) {
```
fzipi, line 140: Same here.

fzipi (changes requested):
~I don't see this being the same patch as v2. Here you implemented the extra `ctl:parseXMLintoArgs`: it is not in v2.~ My bad, this is indeed implemented.

airween:
> I don't see this being the same patch as v2. Here you implemented the extra `ctl:parseXMLintoArgs`: it is not in v2.

It's [there](https://github.com/owasp-modsecurity/ModSecurity/pull/3358/files#diff-932be5a644786245b37df0cd6e69d7a28eaa5f1e08cec566e0cea0b81e505f07).

src/actions/ctl/parse_xml_into_args.cc
airween: Yes, you're definitely right. See 22fee12, it's fixed.

src/actions/ctl/parse_xml_into_args.h
airween: Same as above, 22fee12 fixed it.

src/actions/ctl/parse_xml_into_args.cc
airween: You're definitely right, but I just wanted to follow the conventions (see other files) (even if it's not the prettiest solution). If we want to clean up the code, then we should do that as a sub project.

src/request_body_processor/xml.cc
airween: Sorry, could you show a "reference"? What is the "standard"? What is the "old" standard?

src/request_body_processor/xml.cc
airween: Same as above.

src/request_body_processor/xml.cc
fzipi: The signature of the function is just in the same line, normally.

theseion: As I had pointed out in the other PR, there are already several different ways of formatting long function signatures in the code base. This is one of them.

src/actions/ctl/parse_xml_into_args.cc
fzipi: Does the project want this? It is worth tracking then (e.g. creating a new issue with the subproject?)

airween: > Does the project want this? 

Not necessarily. But we can live together with this.

Just a few example:

* [audit_engine.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/actions/ctl/audit_engine.cc#L31)
* [audit_log_parts.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/actions/ctl/audit_log_parts.cc#L31)
* [request_body_access.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/actions/ctl/request_body_access.cc#L30)
* [rule_egine.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/actions/ctl/rule_engine.cc#L31)
* ...

> It is worth tracking then (e.g. creating a new issue with the subproject?)

Sure, it's been like this for almost ten years. Nobody reported, you are the first, so I can't say this has a priority.

fzipi: Sounds perfect. But (simple) tickets like this might get additional people into the dev side ;)

airween: > But (simple) tickets like this might get additional people into the dev side ;)

Amen! :)

headers/modsecurity/rules_set_properties.h
```diff
@@ -218,14 +225,28 @@ class RulesSetProperties {
 
     /**
      *
-     *
+     * The ConfigBoolean enumerator consists in mapping the different
+     * states of the configuration boolean values.
```
theseion:
```suggestion
     * The ConfigBoolean enumerator the states for configuration boolean values.
```

headers/modsecurity/rules_set_properties.h
```diff
@@ -218,14 +225,28 @@ class RulesSetProperties {
...
     enum ConfigBoolean {
         TrueConfigBoolean,
         FalseConfigBoolean,
         PropertyNotSetConfigBoolean
     };
 
+    /**
+     *
+     * The ConfigXMLParseXmlIntoArgs enumerator consists in mapping the
+     * different states of the configuration XMLParseXmlIntoArgs values.
```
theseion:
```suggestion
     * The ConfigXMLParseXmlIntoArgs enumerator defines the states for the configuration
     * XMLParseXmlIntoArgs values.
```

src/actions/ctl/parse_xml_into_args.cc
```diff
@@ -0,0 +1,63 @@
...
+
+    if (what == "on") {
+        m_secXMLParseXmlIntoArgs = RulesSetProperties::TrueConfigXMLParseXmlIntoArgs;
+    } else if (what == "off") {
+        m_secXMLParseXmlIntoArgs = RulesSetProperties::FalseConfigXMLParseXmlIntoArgs;
+    } else if (what == "onlyargs") {
+        m_secXMLParseXmlIntoArgs = RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs;
+    } else {
+        error->assign("Internal error. Expected: On, Off or OnlyArgs; " \
```
theseion, line 40: The config values here use camel case, but in the comparisons above they're all lower case. Ideally, the values should be case-insensitive. (Looking at the parser below, they probably are?)

src/parser/seclang-scanner.ll
```diff
@@ -90,6 +90,7 @@ ACTION_CTL_BDY_JSON                             (?i:ctl:requestBodyProcessor=JSO
 ACTION_CTL_BDY_XML                              (?i:ctl:requestBodyProcessor=XML)
 ACTION_CTL_BDY_URLENCODED                       (?i:ctl:requestBodyProcessor=URLENCODED)
 ACTION_CTL_FORCE_REQ_BODY_VAR                   (?i:ctl:forceRequestBodyVariable)
+ACTION_CTL_PARSE_XML_INTO_ARGS                  (?i:ctl:parseXMLintoArgs)
```
theseion:
```suggestion
ACTION_CTL_PARSE_XML_INTO_ARGS                  (?i:ctl:parseXmlIntoArgs)
```

src/parser/seclang-scanner.ll
```diff
@@ -410,6 +412,7 @@ CONFIG_VALUE_RELEVANT_ONLY              (?i:RelevantOnly)
 CONFIG_VALUE_SERIAL                     (?i:Serial)
 CONFIG_VALUE_WARN                       (?i:Warn)
 CONFIG_XML_EXTERNAL_ENTITY              (?i:SecXmlExternalEntity)
+CONFIG_XML_PARSE_XML_INTO_ARGS          (?i:SecParseXMLIntoArgs)
```
theseion:
```suggestion
CONFIG_XML_PARSE_XML_INTO_ARGS          (?i:SecParseXmlIntoArgs)
```

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,122 @@ namespace RequestBodyProcessor {
...
+            XMLNodes* xml_data = static_cast<XMLNodes*>(ctx);
+            xml_data->nodes.push_back(std::make_shared<NodeData>());
+            xml_data->node_depth++;
+            // FIXME - later if we want to check the depth of XML tree
+            /* if (max_depth > 0 && max_depth > xml_data->node_depth) {
+                std::cout << "Depth of XML tree reached the given maximum value " << xml_data->node_depth << std::endl;
+                exit(1);
+            } */
+            // if it's not the first (root) item, then append a '.'
+            // note, this can't occur because there is always a pseudo root element: 'xml'
```
theseion: By "this", do you mean that the condition should always be true because of the pseudo root element? If so, I suggest this edit:
```suggestion
            // note, the condition should always be true because there is always a pseudo root element: 'xml'
```

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,122 @@ namespace RequestBodyProcessor {
...
+            xml_data->currpath.append(name);
+        }
+
+        void onEndElement(void * ctx, const xmlChar *localname) {
+            std::string name = reinterpret_cast<const char*>(localname);
+            XMLNodes* xml_data = static_cast<XMLNodes*>(ctx);
+            const std::shared_ptr<NodeData>& nd = xml_data->nodes[xml_data->nodes.size()-1];
+            if (nd->has_child == true) {
+                // check the return value
+                // if it false, then stop parsing
```
theseion:
```suggestion
                // if false, then stop parsing
```

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,122 @@ namespace RequestBodyProcessor {
...
+                // this means the number of arguments reached the limit
+                if (xml_data->m_transaction->addArgument("XML", xml_data->currpath, xml_data->currval, 0) == false) {
+                    xmlStopParser(xml_data->parsing_ctx_arg);
+                }
+            }
+            if (xml_data->currpath.length() > 0) {
+                // set an offset to store this is the first item or not -> remove the '.' or not
```
theseion:
```suggestion
                // set an offset to store whether this is the first item, in order to know whether to remove the '.'
```

src/request_body_processor/xml.cc
```diff
@@ -55,6 +165,27 @@ bool XML::init() {
         /*entity = */xmlParserInputBufferCreateFilenameDefault(
             this->unloadExternalEntity);
     }
+    if (m_transaction->m_secXMLParseXmlIntoArgs
+        == RulesSetProperties::TrueConfigXMLParseXmlIntoArgs ||
+        m_transaction->m_secXMLParseXmlIntoArgs
+        == RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs) {
+        ms_dbg_a(m_transaction, 9,
+                "XML: SecParseXMLIntoArgs is set to " \
```
theseion:
```suggestion
                "XML: SecParseXmlIntoArgs is set to " \
```

src/request_body_processor/xml.cc
```diff
@@ -55,6 +165,27 @@ bool XML::init() {
...
+        m_data.sax_handler->initialized = XML_SAX2_MAGIC;
+        m_data.sax_handler->startElementNs = &MSC_startElement;
+        m_data.sax_handler->endElementNs = &MSC_endElement;
+        m_data.sax_handler->characters = &MSC_xmlcharacters;
+
+        // set the parser state struct
+        m_data.xml_parser_state                  = std::make_unique<XMLNodes>(m_transaction);
+        m_data.xml_parser_state->node_depth      = 0;
+        m_data.xml_parser_state->currval         = "";
+        m_data.xml_parser_state->currpath        = "xml.";
```
theseion, line 204:
```suggestion
        // the XML will contain at least one node, which is the pseudo root node 'xml'
        m_data.xml_parser_state->currpath        = "xml.";
```

src/request_body_processor/xml.cc
```diff
@@ -90,27 +221,67 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
-        ms_dbg_a(m_transaction, 4, "XML: Failed parsing document.");
-        return false;
+    if (m_data.parsing_ctx != NULL &&
+        m_transaction->m_secXMLParseXmlIntoArgs
+        != RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs) {
+        xmlSetGenericErrorFunc(m_data.parsing_ctx, null_error);
+        xmlParseChunk(m_data.parsing_ctx, buf, size, 0);
+        m_data.xml_parser_state->parsing_ctx_arg = m_data.parsing_ctx_arg;
+        if (m_data.parsing_ctx->wellFormed != 1) {
+            error->assign("XML: Failed parsing document.");
```
theseion:
```suggestion
            error->assign("XML: Failed to parse document.");
```

src/request_body_processor/xml.cc
```diff
@@ -90,27 +221,67 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+            ms_dbg_a(m_transaction, 4, "XML: Failed parsing document.");
```
theseion:
```suggestion
            ms_dbg_a(m_transaction, 4, "XML: Failed to parse document.");
```

src/request_body_processor/xml.cc
```diff
@@ -90,27 +221,67 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+            m_transaction->m_secXMLParseXmlIntoArgs
+              == RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs
+            ||
+            m_transaction->m_secXMLParseXmlIntoArgs
+              == RulesSetProperties::TrueConfigXMLParseXmlIntoArgs)
+        ) {
+        xmlSetGenericErrorFunc(m_data.parsing_ctx_arg, null_error);
+        xmlParseChunk(m_data.parsing_ctx_arg, buf, size, 0);
+        if (m_data.parsing_ctx_arg->wellFormed != 1) {
+            error->assign("XML: Failed parsing document for ARGS.");
```
theseion:
```suggestion
            error->assign("XML: Failed to parse document for ARGS.");
```

src/request_body_processor/xml.cc
```diff
@@ -90,27 +221,67 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+            ms_dbg_a(m_transaction, 4, "XML: Failed parsing document for ARGS.");
```
theseion:
```suggestion
            ms_dbg_a(m_transaction, 4, "XML: Failed to parse document for ARGS.");
```

src/request_body_processor/xml.cc
```diff
@@ -119,24 +290,51 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
-
-        if (m_data.well_formed != 1) {
-            error->assign("XML: Failed parsing document.");
-            ms_dbg_a(m_transaction, 4, "XML: Failed parsing document.");
-            return false;
+    if (m_data.parsing_ctx != NULL || m_data.parsing_ctx_arg != NULL) {
+        if (m_data.parsing_ctx != NULL &&
+            m_transaction->m_secXMLParseXmlIntoArgs
+            != RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs) {
+            /* This is how we signalise the end of parsing to libxml. */
```
theseion:
```suggestion
            /* This is how we signal the end of parsing to libxml. */
```

src/request_body_processor/xml.cc
```diff
@@ -119,24 +290,51 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+            m_data.doc = m_data.parsing_ctx->myDoc;
+
+            /* Clean up everything else. */
+            xmlFreeParserCtxt(m_data.parsing_ctx);
+            m_data.parsing_ctx = NULL;
+            ms_dbg_a(m_transaction, 4, "XML: Parsing complete (well_formed " \
+                + std::to_string(m_data.well_formed) + ").");
+
+            if (m_data.well_formed != 1) {
+                error->assign("XML: Failed parsing document.");
```
theseion:
```suggestion
                error->assign("XML: Failed to parse document.");
```

src/request_body_processor/xml.cc
```diff
@@ -119,24 +290,51 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+                ms_dbg_a(m_transaction, 4, "XML: Failed parsing document.");
```
theseion:
```suggestion
                ms_dbg_a(m_transaction, 4, "XML: Failed to parse document.");
```

src/request_body_processor/xml.cc
```diff
@@ -119,24 +290,51 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+        }
+        if (m_data.parsing_ctx_arg != NULL &&
+            (
+                m_transaction->m_secXMLParseXmlIntoArgs
+                  == RulesSetProperties::OnlyArgsConfigXMLParseXmlIntoArgs
+                  ||
+                m_transaction->m_secXMLParseXmlIntoArgs
+                  == RulesSetProperties::TrueConfigXMLParseXmlIntoArgs)
+            ) {
+            /* This is how we signalise the end of parsing to libxml. */
```
theseion:
```suggestion
            /* This is how we signale the end of parsing to libxml. */
```

src/request_body_processor/xml.cc
```diff
@@ -119,24 +290,51 @@ bool XML::processChunk(const char *buf, unsigned int size,
...
+            if (xmlParseChunk(m_data.parsing_ctx_arg, NULL, 0, 1) != 0) {
+                if (m_data.xml_error != "") {
+                    error->assign(m_data.xml_error);
+                }
+                else {
+                    error->assign("XML: Failed parsing document for ARGS.");
```
theseion:
```suggestion
                    error->assign("XML: Failed to parse document for ARGS.");
```

src/request_body_processor/xml.h
```diff
@@ -33,12 +34,50 @@ namespace RequestBodyProcessor {
...
+* XMLNodes for parsing XML into args
+*/
+class XMLNodes {
+    public:
+        std::vector<std::shared_ptr<NodeData>> nodes;
+        unsigned long int node_depth;
+        std::string       currpath;
+        std::string       currval;
+        Transaction      *m_transaction;
+        // need to store context - this is the same as at the xml_data
```
theseion:
```suggestion
        // need to store context - this is the same as in xml_data
```

src/request_body_processor/xml.h
```diff
@@ -33,12 +34,50 @@ namespace RequestBodyProcessor {
...
+    std::unique_ptr<xmlSAXHandler> sax_handler;
     xmlParserCtxtPtr parsing_ctx;
     xmlDocPtr doc;
 
     unsigned int well_formed;
+
+    /* error reporting and XML array flag */
+    std::string               xml_error;
+
+    /* another parser context for arguments */
```
theseion:
```suggestion
    /* additional parser context for arguments */
```

theseion (changes requested):

src/request_body_processor/xml.cc
airween: Indeed you're right, thanks.

src/parser/seclang-scanner.ll
airween: Well, the lexical analyzer is case insensitive (see `?i` modifier), so it does not count here.

In [discussion](https://github.com/owasp-modsecurity/ModSecurity/issues/3178) I proposed the syntax in this form as like is above.

Now I realized that I used your suggestion in [v2](https://github.com/owasp-modsecurity/ModSecurity/pull/3358/files#diff-932be5a644786245b37df0cd6e69d7a28eaa5f1e08cec566e0cea0b81e505f07R1389) so I think I have to modify this.

Or modify it in case of v2? What do you think?

src/parser/seclang-scanner.ll
airween: Similar as above  (see `?i` modifier), but here in  [v2 PR](https://github.com/owasp-modsecurity/ModSecurity/pull/3358/files#diff-4d3cca966360f6cd98af24e577945168781977cb27a55e57ee0c2cb5033a8f6dR4502) I used this syntax too, and I explained this in [discussion](https://github.com/owasp-modsecurity/ModSecurity/issues/3178), I would touch this.

So may be I should update the syntax not just here but in #3358 too...

Thoughts?

src/request_body_processor/xml.cc
airween: I'll wait until we decide the final syntax :smile:.

src/parser/seclang-scanner.ll
theseion: My persona preference is to use "strict" camel case, where abbreviations like `XML` follow the same rules. In my experience this leads to more readable variable names. In this case, you may want to follow the naming of existing variables / actions / directives. I'd probably do that, just to be consistent for users.

airween:
A [few](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/14694933615/job/41235279762?pr=3363#step:9:37) [tests](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/14694933615/job/41235279762?pr=3363#step:9:182) were failed after some commit - I have to check the reason.

src/parser/seclang-scanner.ll
theseion: Just read your comment in the other PR. There, `SecXmlExternalEntity`) already follows my preferred format. So from my point of view, I'd use that format for all the new variables / directives / actions (there are a couple more uppercase "XML" that I didn't mention, but there I wasn't sure whether you were following some internal consistency).

airween: Okay, this was fixed in bf707de.

src/parser/seclang-scanner.ll
airween: This was also fixed in bf707de.

src/actions/ctl/parse_xml_into_args.cc
airween: No, the reason here is something else. The syntax parser does this [here](https://github.com/airween/ModSecurity/blob/v3/xmlargsfeat/src/parser/seclang-parser.yy#L2717-L2725).

It's a bit weird solution, but the code follows this way though all other places (where it's necessary) so I used this one here too.

airween:
I added a few new commits here as at [#3358](https://github.com/owasp-modsecurity/ModSecurity/pull/3358#issuecomment-2836266539):

* bf707de - changed the directive SecXMLintoArgs to SecXmlIntoArgs
* e8dc60e - there is a small change in the node value's parsing: if the node value contains a multi-byte character the SAX calls the function multiple times, so we need to concatenate those sub-strings, not creating a new string every time
* 89442ed - changed the directives in tests and add a new test with a multibyte character

fzipi:
Just out of curiosity, have you tried enabling this and sending a big xml file?

airween:
> Just out of curiosity, have you tried enabling this and sending a big xml file?

What do you mean "a big xml file"?

Btw the size of the file is just one thing... There are couple of limits in the engine when you're sending an XML payload and want to process it as ARGS, like:
* `SecRequestBodyNoFilesLimit` - the default value is 128kB
* `SecArgumentsLimit` - default value is 1000

I sent an XML with size > 400kB (I increased the `SecRequestBodyNoFilesLimit`), then I get:
```
ModSecurity: XML parsing error: More than 1000 ARGS (GET + XML) [hostname "localhost"] [uri "/post"] [
ModSecurity: Access denied with code 400 (phase 2). Match of "eq 0" against "REQBODY_ERROR" required. [file "/etc/modsecurity/modsecurity.conf"] [line "78"] [id "200002"] [msg "Failed to parse request body."] [data "XML parsing error: More than 1000 ARGS (GET + XML)"] [severity "CRITICAL"] [hostname "localhost"] [uri "/post"]
```

Actually, the behavior of this function is exactly the same as that used for JSON. There is no difference. You could ask _"have you tried enabling this and sending a big JSON file?"_. But the JSON works **as is**, and this feature is **optional**.

If you tell me what you're curious about, I'll try it - with an XML and a JSON too.

fzipi:
> Just out of curiosity, have you tried enabling this and sending a big xml file?

The sentence starts with `Just out of curiosity`. So yes, _just curious_.

headers/modsecurity/rules_set_properties.h
```diff
@@ -225,8 +225,7 @@ class RulesSetProperties {
 
     /**
      *
-     * The ConfigBoolean enumerator consists in mapping the different
-     * states of the configuration boolean values.
+     * The ConfigBoolean enumerator the states for configuration boolean values.
```
theseion:
```suggestion
     * The ConfigBoolean enumerator defines the states for configuration boolean values.
```

headers/modsecurity/rules_set_properties.h
```diff
@@ -177,6 +182,7 @@ class RulesSetProperties {
         m_secRequestBodyAccess(PropertyNotSetConfigBoolean),
         m_secResponseBodyAccess(PropertyNotSetConfigBoolean),
         m_secXMLExternalEntity(PropertyNotSetConfigBoolean),
+        m_secXMLParseXmlIntoArgs(PropertyNotSetConfigXMLParseXmlIntoArgs),
```
theseion, line 185:
```suggestion
        m_secXmlExternalEntity(PropertyNotSetConfigBoolean),
        m_secXmlParseXmlIntoArgs(PropertyNotSetConfigXMLParseXmlIntoArgs),
```

headers/modsecurity/rules_set_properties.h
```diff
@@ -218,14 +225,27 @@ class RulesSetProperties {
...
+     *
+     * The ConfigXMLParseXmlIntoArgs enumerator defines the states for the configuration
+     * XMLParseXmlIntoArgs values.
+     * The default value is PropertyNotSetConfigXMLParseXmlIntoArgs.
+     */
+    enum ConfigXMLParseXmlIntoArgs {
+        TrueConfigXMLParseXmlIntoArgs,
+        FalseConfigXMLParseXmlIntoArgs,
+        OnlyArgsConfigXMLParseXmlIntoArgs,
+        PropertyNotSetConfigXMLParseXmlIntoArgs
```
theseion, line 247:
```suggestion
     * The ConfigXmlParseXmlIntoArgs enumerator defines the states for the configuration
     * XmlParseXmlIntoArgs values.
     * The default value is PropertyNotSetConfigXmlParseXmlIntoArgs.
     */
    enum ConfigXmlParseXmlIntoArgs {
        TrueConfigXmlParseXmlIntoArgs,
        FalseConfigXmlParseXmlIntoArgs,
        OnlyArgsConfigXmlParseXmlIntoArgs,
        PropertyNotSetConfigXmlParseXmlIntoArgs
```

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,130 @@ namespace RequestBodyProcessor {
...
+NodeData::NodeData() {
+    has_child = false;
+}
+
+NodeData::~NodeData() {};
+
+/*
+* XMLNodes for parsing XML into args
+*/
+XMLNodes::XMLNodes(Transaction *transaction)
```
theseion, line 40:
```suggestion
XmlNodes::XmlNodes(Transaction *transaction) 
```

src/request_body_processor/xml.cc
```diff
@@ -25,11 +25,130 @@ namespace RequestBodyProcessor {
...
+            handler->onEndElement(userData, name);
+    }
+
+    void MSC_xmlcharacters(void *userData, const xmlChar *ch, int len) {
+        MSCSAXHandler* handler = static_cast<MSCSAXHandler*>(userData);
+        handler->onCharacters(userData, ch, len);
+    }
+}
+
 XML::XML(Transaction *transaction)
```
theseion, line 152:
```suggestion
Xml::Xml(Transaction *transaction)
```

theseion (changes requested):

airween:
> The sentence starts with `Just out of curiosity`. So yes, _just curious_.

I saw, but honestly: your question is absolutely legitimate. That's why I'm asking you: what should we expect with what size of XML, and/or how many nodes of XML?

I wrote a small script which helps to generate different payloads in XML and JSON - but with same content. I tried them, and the problem is not in the conversion (JSON to `ARGS` and XML to `ARGS`), but in the rules' processing. With the same structure in XML and JSON, I got the same `ARGS` list in the engine, but above 1000 of built `ARGS` performance deteriorates sharply, and rules which check `ARGS` and `ARGS_NAMES` will be very slow. Now I want to add some other features to that script, eg. generate `x-www-form-urlencoded` payload, and try that too.

Which is clearly visible: the critical value is not the size of the file, but number of nodes (both in XML and in JSON).

And of course: without any rules there is no any performance issue even with a huge number of node JSON/XML file - indeed the aim is not to forget the rules, but that's also a good question with which rules should we test?

headers/modsecurity/rules_set_properties.h
airween: [m_secXMLExternalEntity](https://github.com/owasp-modsecurity/ModSecurity/blob/bbef22b3b5ba2c040b32f93c81338601286abb87/headers/modsecurity/rules_set_properties.h#L179) property is already a member of `RulesSetProperties` class. If we change here the format, then we must change that every place, eg. in [seclang-parser.yy](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/parser/seclang-parser.yy#L1683) (and other line in the same file), and in few other files.

We could do this, but I'm afraid this wouldn't be a fair to those who use the project in some forked form.

(Consider a company uses libmodsecurity3 with their own modifications, and they follow the mainline source tree from our repository. Consider they uses these (modified) class members, functions that we changed their names. I think this is a modification at a level where at least a subversion number should be changed, I mean 3.0.XX to 3.1.XX - but I'm not sure we have any (other) strict reason to do that).

But because the old syntax is `m_SecXML...` therefore I followed that syntax in case of new members and functions, so I would keep this (unless we can change `Xml` to `XML` in case of second occurrence).

headers/modsecurity/rules_set_properties.h
airween: I can change this, because this `enum` type is new, but in this case we will deviate from the above-mentioned syntax (`m_SecXML...`). What do you think?

src/request_body_processor/xml.cc
airween: I can change this syntax too because it's new, but the existing code uses `XML` everywhere. I think my syntax matches with that better.

src/request_body_processor/xml.cc
airween: Same as I wrote [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3363#discussion_r2070781346). This is an existing code, the `XML` class and its constructor is already exists. It wouldn't be fair to change (and if we change we must align all places in the code).

theseion:
> I saw, but honestly: your question is absolutely legitimate. That's why I'm asking you: what should we expect with what size of XML, and/or how many nodes of XML?

Unfortunately, this depends heavily on the installation. With SOAP, for example, you'll probably process 50-100 nodes per request, with some requests sending much larger XMLs. I don't know whether we can do much more than document the performance given different numbers of nodes and let users decide how to configure the engine.

> And of course: without any rules there is no any performance issue even with a huge number of node JSON/XML file - indeed the aim is not to forget the rules, but that's also a good question with which rules should we test?

You could use rules with `@unconditionalMatch` and `multiMatch`, one each for `ARGS`, `ARGS_NAMES`, and `XML`. This should give you worst case performance for each target.

headers/modsecurity/rules_set_properties.h
theseion: That's fine. I just marked all occurrences for consistency.

headers/modsecurity/rules_set_properties.h
theseion: Personally, I'd use the "new" spelling. I don't think it matters here.

airween:
I added a new commit:

* a3876e3 - there is a small change in the node value's parsing: if the node value contains a multi-byte character the SAX calls the function multiple times, so we need to concatenate those sub-strings, not creating a new string every time

theseion (approved):

fzipi (approved):

Issue #3310, opened 2024-12-03
ModSec Audit logs in JSON format - not JSON?


paul-hammant:
**Describe the bug**

I'm running ModSec as a library for testing purposes (with GoTestWAF, but that's not important I don't think). It is printing JSON audit files (one per request) to the file system.  Field time_stamp looks corrupted though as I gaze and the files with JSON that won't parse cos `"time_stamp": "\u0014\u00108....\u0006"` is choking subsequent parsers (like the one in browsers).

**Source code question**

In `src/transaction.cc`, should c_str() be called twice?

```
    std::string ts = utils::string::ascTime(&m_timeStamp).c_str();

    .. snip .. 

    LOGFY_ADD("time_stamp", ts.c_str());
```

airween:
Hi @paul-hammant,

could you show your relevant configuration? I've never faced any issue like this - and I think the regression test framework covers all scenario (haven't checked this specific one), so I would wonder if this is a new issue (if it is). But with a config snippet I try to check that.

paul-hammant:
I'm using ModSec.so from Java in a corporate setting. GoTestWAF is scoring it quite highly, but in analysing the audit logs I see this unparsable JSON. I don't have a need to have machine-processable JSON logs, so it'll never become an issue for me. I would've assumed that it was encoded as UTF-8 as almost everything is these days. It's not though. If I need to in the short term I can swap the file's epoch time for the sequence by bytes in between `"` and `"`.

Another day I'll try to make the local change and see if it works. I'll report back then.

paul-hammant:
After some trial and error, this looks like it:  https://github.com/paul-hammant/ModSecurity/commit/7010ddd8222afa2608ce506697d7c20187364a84

Still to report back from an actual deployment.

airween:
Hi @paul-hammant,

thanks - well, I don't see any relevant difference so if this solves your issue that would be a surprise :smile:

airween:
Hi @paul-hammant,

I've already inspected this issue, and need some clarification.

In your initial comment you wrote:
> Field time_stamp looks corrupted though as I gaze and the files with JSON that won't parse cos `"time_stamp": "\u0014\u00108....\u0006"` is choking subsequent parsers (like the one in browsers).

[...]

PR #3364, opened 2025-04-22
Simplify code for JSON audit log

JakubOnderka:
`yajl_gen_string` doesn't require null terminated string, so we don't need to call `strlen`.

This simplifies code for JSON transaction logging.

### DIFF
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -311,7 +311,7 @@ class RulesSetProperties {
     };
 
 
-    static const char *ruleEngineStateString(RuleEngine i) {
+    static std::string ruleEngineStateString(RuleEngine i) {
         switch (i) {
         case DisabledRuleEngine:
             return "Disabled";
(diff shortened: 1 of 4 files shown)


airween:
Hi @JakubOnderka,

thanks for this PR!

Have you seen the issue #3310? There the OP mentioned a [commit](https://github.com/paul-hammant/ModSecurity/commit/7010ddd8222afa2608ce506697d7c20187364a84) - may be we can consider to add that modification too. What do you think?

JakubOnderka:
Good question. So I added test case that verifies if timestamp is correct in JSON format and it looks good.

JakubOnderka:
@airween I also find out that in some cases, JSON log is invalid because `m_httpVersion` variable is empty string. So I converted `http_version` field in JSON audit log to string, so if  `m_httpVersion` will be empty, JSON will still be valid.

airween (approved):
LGTM

airween:
Hi @JakubOnderka,

a nice catch again - thank you!

PR #3365, opened 2025-04-23
Disable Expect when sending audit logs to remote HTTP server

JakubOnderka:
By default, curl sends `Expect: 100-continue` request HTTP header when sending POST data. This requires another round trip, because then curl waits until receives `HTTP/1.1 100 Continue` from remote server. 

When we disable Expect: 100-continue, sending logs to remote server should be faster.


airween (approved):
LGTM

airween:
Hi @JakubOnderka,

thank you for your PR, it's a nice catch!

PR #3371, opened 2025-05-06
Updated Code from PR #2304

akama-aka:
I won't continue that because I don't have the time and motivation to do so. Even tho it's just vibe coded and thats not a way how to solve issues or missing features.

### DIFF
--- a/src/audit_log/audit_log.cc
+++ b/src/audit_log/audit_log.cc
@@ -369,6 +369,12 @@ bool AuditLog::merge(AuditLog *from, std::string *error) {
     return init(error);
 }
 
+bool AuditLog::reopen(std::string *error) {
+    if (m_writer != NULL) {
+        return m_writer->reopen(error);
+    }
+    return true;
+}
 
 }  // namespace audit_log
 }  // namespace modsecurity
--- a/src/rules_set.cc
+++ b/src/rules_set.cc
@@ -325,6 +325,25 @@ extern "C" void msc_rules_error_cleanup(const char *error) {
     free((void*) error);
 }
 
+extern "C" int msc_rules_reopen_audit_log(RulesSet *rules, const char **error) {
+    bool succeeded = true;
+    std::string errorStr;
+
+    if (rules->m_auditLog != NULL) {
+        succeeded = rules->m_auditLog->reopen(&errorStr);
+    }
+
+    if (!succeeded) {
+        if (!errorStr.empty()) {
+            *error = strdup(errorStr.c_str());
+        } else {
+            // Guarantee an error message is always assigned in the event of a failure
+            *error = strdup("Unknown error reopening audit log");
+        }
+    }
+
+    return succeeded ? 0 : -1;
+}
 
 extern "C" int msc_rules_cleanup(RulesSet *rules) {
     delete rules;
--- a/test/cppcheck_suppressions.txt
+++ b/test/cppcheck_suppressions.txt
@@ -21,7 +21,12 @@ shiftNegative:src/utils/msc_tree.cc
 
 // 
 // ModSecurity v3 code...
-// 
+//
+
+functionConst:src/utils/shared_files.h:60
+useInitializationList:src/utils/shared_files.h:88
+
+
 variableScope:src/operators/rx.cc
 variableScope:src/operators/rx_global.cc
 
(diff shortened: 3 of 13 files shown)


src/audit_log/audit_log.cc
```diff
@@ -369,6 +369,12 @@ bool AuditLog::merge(AuditLog *from, std::string *error) {
     return init(error);
 }
 
+bool AuditLog::reopen(std::string *error) {
+    if (m_writer != NULL) {
```
airween, line 373: As [Sonarcloud](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3371&id=owasp-modsecurity_ModSecurity&open=AZakSaSxaqw6mMcFxJ_a) shows here you should use `nullptr` instead of `NULL`. I see there are many other `NULL` literals in this file - feel free to replace them.

src/rules_set.cc
```diff
@@ -325,6 +325,25 @@ extern "C" void msc_rules_error_cleanup(const char *error) {
...
+        if (!errorStr.empty()) {
+            *error = strdup(errorStr.c_str());
+        } else {
+            // Guarantee an error message is always assigned in the event of a failure
+            *error = strdup("Unknown error reopening audit log");
+        }
+    }
+
+    return succeeded ? 0 : -1;
+}
```
airween, line 347: I know this is a completely valid solution, but as you can see in other "C" API functions, the authors tied to avoid all the logic and use C++ types (like here the `bool` and `std::string`).

My suggestion here: create a thin wrapper, an "extra" function.

You should add a new function to `RulesSet` class with name `reopen()`, move this body there, and in this "C" API you have to call only that function.

test/cppcheck_suppressions.txt
```diff
@@ -21,7 +21,12 @@ shiftNegative:src/utils/msc_tree.cc
 
 // 
 // ModSecurity v3 code...
-// 
+//
+
+functionConst:src/utils/shared_files.h:60
```
airween, line 26: Generally: please don't use this file to collect suppressions. If you need to add some suppression, please use an inline one.

But here: why it this necessary? It seems in that file in line [60](https://github.com/owasp-modsecurity/ModSecurity/blob/b0a56941fab7bc50bd13203dc11810231392e797/src/utils/shared_files.h#L60) there is only a condition...

test/cppcheck_suppressions.txt
```diff
@@ -21,7 +21,12 @@ shiftNegative:src/utils/msc_tree.cc
...
+useInitializationList:src/utils/shared_files.h:88
```
airween, line 27: Similar here, but it seems that file only has [77](https://github.com/owasp-modsecurity/ModSecurity/blob/b0a56941fab7bc50bd13203dc11810231392e797/src/utils/shared_files.h#L77) lines, so adding line 88 makes no sense.

airween (changes requested):

airween:
Hi @akama-aka,

many thanks for this pull request, and special welcome to your first contribution!

I added some questions above, please review them.

And one more question - you wrote:
> Implements the ability to reopen audit log files to ensure compatibility with **Linux** log rotation

Could you try this patch on Windows? I think the result is almost the same, but I'm curios how Windows handles the `reopen()` method. Do you have any idea how can we add a test case to check this method?

Cc: @eduar-hte (regarding to Windows).

And finally: I saw you added a "C" API function. I assume that will be used in Nginx to reopen audit.log files, right? Do you mind to add that feature to Nginx connector if we merge this?

akama-aka:
> Hi @akama-aka,
> 
> many thanks for this pull request, and special welcome to your first contribution!
> 
> I added some questions above, please review them.
> 
> And one more question - you wrote:
> > Implements the ability to reopen audit log files to ensure compatibility with **Linux** log rotation
> 
> Could you try this patch on Windows? I think the result is almost the same, but I'm curios how Windows handles the `reopen()` method. Do you have any idea how can we add a test case to check this method?
> 
> Cc: @eduar-hte (regarding to Windows).
> 
> And finally: I saw you added a "C" API function. I assume that will be used in Nginx to reopen audit.log files, right? Do you mind to add that feature to Nginx connector if we merge this?

Thank you for your comment to this draft pr. I'll try to learn C/C++ more to understand more of it and try to fix those things up ^^

PR #3372, opened 2025-05-09
Fix compiler warnings from GCC

notroj:
## what

* Fix compiler warnings

## why

* Compiler warnings are bad

### DIFF
--- a/apache2/acmp.c
+++ b/apache2/acmp.c
@@ -251,18 +251,6 @@ static void acmp_add_node_to_parent(acmp_node_t *parent, acmp_node_t *child) {
     }
 }
 
-/**
- * Copies values from one node to another, without child/sibling/fail pointers
- * and without state variables.
- */
-static void acmp_clone_node_no_state(acmp_node_t *from, acmp_node_t *to) {
-    memcpy(to, from, sizeof(acmp_node_t));
-    to->child = NULL;
-    to->sibling = NULL;
-    to->fail = NULL;
-    to->hit_count = 0;
-}
-
 static inline acmp_node_t *acmp_btree_find(acmp_node_t *node, acmp_utf8_char_t letter) {
     acmp_btree_node_t *bnode = node->btree;
     for (;;) {
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -797,7 +797,6 @@ static const char *add_rule(cmd_parms *cmd, directory_config *dcfg, int type,
     int type_with_lua = 1;
     int type_rule;
     int rule_actionset;
-    int offset = 0;
 
     #ifdef DEBUG_CONF
     ap_log_perror(APLOG_MARK, APLOG_STARTUP|APLOG_NOERRNO, 0, cmd->pool,
@@ -1935,7 +1934,6 @@ static const char *cmd_conn_read_state_limit(cmd_parms *cmd, void *_dcfg,
         ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_conn_read_state_limit: _dcfg is NULL");
         return NULL;
     }
-    directory_config *dcfg = (directory_config *)_dcfg;
     long int limit;
 
     limit = strtol(p1, NULL, 10);
@@ -1991,7 +1989,6 @@ static const char *cmd_conn_write_state_limit(cmd_parms *cmd, void *_dcfg,
         ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_conn_write_state_limit: _dcfg is NULL");
         return NULL;
     }
-    directory_config *dcfg = (directory_config *)_dcfg;
     long int limit;
 
     limit = strtol(p1, NULL, 10);
@@ -2519,7 +2516,6 @@ static const char *cmd_sever_conn_filters_engine(cmd_parms *cmd, void *_dcfg,
         ap_log_perror(APLOG_MARK, APLOG_EMERG, 0, cmd->pool, "cmd_sever_conn_filters_engine: _dcfg is NULL");
         return NULL;
     }
-    directory_config *dcfg = (directory_config *)_dcfg;
 
     if (strcasecmp(p1, "on") == 0)
     {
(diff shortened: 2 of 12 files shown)


apache2/acmp.c
```diff
@@ -251,6 +251,7 @@ static void acmp_add_node_to_parent(acmp_node_t *parent, acmp_node_t *child) {
     }
 }
 
+#if 0
```
airween: Nice catch. It seems like we don't use this function at all - do we want to keep this unused code?

apache2/msc_xml.c
```diff
@@ -132,29 +132,19 @@ static void msc_xml_on_characters(void *ctx, const xmlChar *ch, int len) {
 }
 
 
-static xmlParserInputBufferPtr
```
airween: Have you removed this function because it always returned a NULL?

apache2/re_operators.c
```diff
@@ -4231,11 +4231,6 @@ static int msre_op_fuzzy_hash_init(msre_rule *rule, char **error_msg)
     return 1;
 #endif
     return 1;
-
-invalid_parameters:
```
airween: It seems like this label is still referenced [here](https://github.com/owasp-modsecurity/ModSecurity/blob/82c5c3a6dec1c68c081dc69eee5faf9a34818852/apache2/re_operators.c#L4171) and [here](https://github.com/owasp-modsecurity/ModSecurity/blob/82c5c3a6dec1c68c081dc69eee5faf9a34818852/apache2/re_operators.c#L4188). Do we really want to remove this?

airween (changes requested):

airween:
Hi @notroj,

first of all, many thanks for this PR.

There are some other warnings in different files (usually with different kind of warnings) - do you plan to fix those too?

apache2/re_operators.c
notroj: Ah, I missed that because I don't have that feature enabled. I'll restore this.

apache2/msc_xml.c
notroj: Ah, this is a mistake too. Since `entity` was set but never used in xml_init I assumed this was safe to remove, but `xmlParserInputBufferCreateFilenameDefault` should still be called here. I'll restore it and fix the warning another way.

notroj:
I've repushed with discussed fixes. I am currently testing with a set of CFLAGS similar to what we build with in Fedora/RHEL, but some low value warnings turned off:

```
CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -mtls-dialect=gnu2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -Wno-maybe-uninitialized -Wno-deprecated-declarations -Wno-unused -Wno-pointer-sign '
```

There is only one outstanding warning:

```
In file included from /usr/include/stdio.h:970,
                 from ../apache2/modsecurity.h:18,
                 from ../apache2/msc_util.c:15:
In function ‘sprintf’,
    inlined from ‘msc_headers_to_buffer’ at ../apache2/msc_util.c:2331:17:
/usr/include/bits/stdio2.h:30:10: warning: ‘__sprintf_chk’ argument 5 overlaps destination object ‘buffer’ [-Wrestrict]
   30 |   return __builtin___sprintf_chk (__s, __USE_FORTIFY_LEVEL - 1,
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   31 |                                   __glibc_objsize (__s), __fmt,
      |                                   ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   32 |                                   __va_arg_pack ());
      |                                   ~~~~~~~~~~~~~~~~~
../apache2/msc_util.c: In function ‘msc_headers_to_buffer’:
../apache2/msc_util.c:2306:64: note: destination object referenced by ‘restrict’-qualified argument 1 was declared here
 2306 | int msc_headers_to_buffer(const apr_array_header_t *arr, char *buffer,
      |                                                          ~~~~~~^~~~~~
  CCLD     msc_test
```

This is similar to the one of the other problems - using the destination buffer in the snprintf format arguments. It would IMO make sense to rewrite this function to use `apr_table_do` which might be simpler anyway, I have a patch in progress but I wasn't able to run the tests.

I see you just fixed the tests :smile: in 888109736bea139a3e0b65df79dd32176181feb3 and now I can this PR broke them again due to using `ap_bin2hex`, I will work on these.

airween:
> I've repushed with discussed fixes. I am currently testing with a set of CFLAGS similar to what we build with in Fedora/RHEL, but some low value warnings turned off:
[...]
> There is only one outstanding warning:
[...]
> This is similar to the one of the other problems - using the destination buffer in the snprintf format arguments. It would IMO make sense to rewrite this function to use `apr_table_do` which might be simpler anyway, I have a patch in progress but I wasn't able to run the tests.

Thank you for taking this.

Let me know if I can help you anything.

> I see you just fixed the tests 😄 in [8881097](https://github.com/owasp-modsecurity/ModSecurity/commit/888109736bea139a3e0b65df79dd32176181feb3) and now I can this PR broke them again due to using `ap_bin2hex`, I will work on these.

Ah yes. This has been on my list for a year now.... I was talking to another developer a few days/week ago and he noticed that there were issues with the tests, so I thought I'd fix the broken parts. Sorry for the inconvenience :smiley:.

airween:
Hi @notroj,

just want to let you know, I plan to release a new version if mod_security2 next week. If you have enough time to make request changes, then we could merge them before the release. Thanks!

notroj:
@airween I went with simpler fixes and repushed. This addresses every warning and `make check` passes.

apache2/acmp.c
notroj: Removed in 9d9a72734988bbd4f101cbea0d5df60eabcc1435

apache2/msc_xml.c
notroj: Removed this change in 9d9a72734988bbd4f101cbea0d5df60eabcc1435

apache2/re_operators.c
notroj: Restored in 9d9a72734988bbd4f101cbea0d5df60eabcc1435

airween (approved):
LGTM

airween:
Thanks, @notroj!

PR #3377, opened 2025-05-10
fix: add PCRE2 capability to standalone module

airween:
## what

This PR adds `PCRE2` related code to `standalone` module.

## why

Previously if the user chose PCRE2 regex engine (with `--with-pcre2` at build time), the standalone module did not handle that, and still wanted to be built with the old regex engine.

Furthermore, the modules `make test` workflow also uses `standalone` module too, and with `PCRE2` the whole test case was broken. This patch fixes that too.

### DIFF
--- a/standalone/regex.c
+++ b/standalone/regex.c
@@ -61,7 +61,11 @@ AP_DECLARE(ap_regex_t *) ap_pregcomp(apr_pool_t *p, const char *pattern,
 
 AP_DECLARE(void) ap_regfree(ap_regex_t *preg)
 {
+#ifdef WITH_PCRE2
+(pcre2_code_free)(preg->re_pcre);
+#else
 (pcre_free)(preg->re_pcre);
+#endif
 }
 
 AP_DECLARE(int) ap_regcomp(ap_regex_t *preg, const char *pattern, int cflags)
@@ -71,6 +75,23 @@ int erroffset;
 int options = 0;
 int nsub = 0;
 
+#ifdef WITH_PCRE2
+if ((cflags & AP_REG_ICASE) != 0) options |= PCRE2_CASELESS;
+if ((cflags & AP_REG_NEWLINE) != 0) options |= PCRE2_MULTILINE;
+int error_number = 0;
+PCRE2_SIZE error_offset = 0;
+PCRE2_SPTR pcre2_pattern = (PCRE2_SPTR)pattern;
+
+preg->re_pcre = pcre2_compile(pcre2_pattern, PCRE2_ZERO_TERMINATED,
+  options, &error_number, &error_offset, NULL);
+preg->re_erroffset = error_offset;
+
+if (preg->re_pcre == NULL) return AP_REG_INVARG;
+
+pcre2_pattern_info((const pcre2_code *)preg->re_pcre, PCRE2_INFO_CAPTURECOUNT, &nsub);
+preg->re_nsub = nsub;
+
+#else
 if ((cflags & AP_REG_ICASE) != 0) options |= PCRE_CASELESS;
 if ((cflags & AP_REG_NEWLINE) != 0) options |= PCRE_MULTILINE;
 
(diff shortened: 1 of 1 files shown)


standalone/regex.c
```diff
@@ -1,162 +1,237 @@
...
+if ((cflags & AP_REG_NEWLINE) != 0) options |= PCRE_MULTILINE;
+
+preg->re_pcre = pcre_compile(pattern, options, &errorptr, &erroffset, NULL);
+preg->re_erroffset = erroffset;
+
+if (preg->re_pcre == NULL) return AP_REG_INVARG;
+
+pcre_fullinfo((const pcre *)preg->re_pcre, NULL, PCRE_INFO_CAPTURECOUNT, &nsub);
+preg->re_nsub = nsub;
+#endif  // end of WITH_PCRE
```
theseion, line 105:
```suggestion
#endif
```

standalone/regex.c
```diff
@@ -1,162 +1,237 @@
...
+  else
+    {
+    ovector = (int *)malloc(sizeof(int) * nmatch * 3);
+    if (ovector == NULL) return AP_REG_ESPACE;
+    allocated_ovector = 1;
+    }
+  }
+
+#ifdef WITH_PCRE2
+{
```
theseion, line 148: Why is this in a brace delimited block?

standalone/regex.c
```diff
@@ -1,162 +1,237 @@
...
+      pcre2_match_data_free(match_data);
+      pcre2_match_context_free(match_context);
+  }
+}
+#else
+rc = pcre_exec((const pcre *)preg->re_pcre, NULL, string, (int)strlen(string),
+  0, options, ovector, nmatch * 3);
+#endif
+
+if (rc == 0) rc = nmatch;    /* All captured slots were filled in */
```
theseion, line 185: I don's see `rc` beeing set in the PCRE2 case. It looks to me like the code in the following couple of lines shouldn't run in case of PCRE2.

standalone/regex.c
```diff
@@ -1,162 +1,237 @@
...
+#ifdef PCRE_ERROR_MATCHLIMIT
+    case PCRE_ERROR_MATCHLIMIT: return AP_REG_ESPACE;
+#endif
+#ifdef PCRE_ERROR_BADUTF8
+    case PCRE_ERROR_BADUTF8: return AP_REG_INVARG;
+#endif
+#ifdef PCRE_ERROR_BADUTF8_OFFSET
+    case PCRE_ERROR_BADUTF8_OFFSET: return AP_REG_INVARG;
+#endif
+#endif // end of WITH_PCRE
```
theseion, line 237:
```suggestion
#endif
```

theseion (changes requested):

standalone/regex.c
airween: Instead of removing this comment, I should add a new comment to line 94:
```
#else // otherwise use PCRE
```

standalone/regex.c
airween: This isn't necessarily, but in C code we use this sometimes. It helps to separate the code. We can remove, but usually it does not disturb here.

theseion: No, that's fine. It just looked like it might have been a mistake.

standalone/regex.c
airween: Same as [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3377#discussion_r2083110071).

standalone/regex.c
airween: You're right. I assumed `rc` is initialized with `0` when it's declared, but that's sure. See c2216b26 where I set it up (with the given comment).

standalone/regex.c
airween: Line 221 already has the comment.

standalone/regex.c
airween: It's fixed in e978a301.

theseion (approved):

PR #3379, opened 2025-05-11
feat: add 'make test' to v2's workflow

airween:
## what

This PR adds mod_security2 module's own test cases to the GH workflow.

The new job uses the same matrix as the existing one, but we can't use `--enable-assertions` configure option (which is important with the valid configuration, where we do not use invalid values), because then the test cases which check invalid values would throw an assert error.

## why

The module's `make tests` command hasn't been working for some time, because the PCRE2 modifications weren't applied that code. With some recent PR's those part of code were aligned, so `make test` works again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -74,3 +74,39 @@ jobs:
       - name: Show mod_security2 audit log
         if: always()
         run: sudo cat /var/log/apache2/modsec_audit.log
+
+  test-linux:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        os: [ubuntu-22.04]
+        platform: [x32, x64]
+        compiler: [gcc, clang]
+        configure:
+          - {label: "with pcre, no study, no jit",      opt: "--enable-pcre-study=no" }
+          - {label: "with pcre, with study, no jit",    opt: "--enable-pcre-study=yes" }
+          - {label: "with pcre, no study, with jit",    opt: "--enable-pcre-study=no --enable-pcre-jit" }
+          - {label: "with pcre, with study, with jit",  opt: "--enable-pcre-study=yes --enable-pcre-jit" }
+          - {label: "with pcre2",                       opt: "--with-pcre2 --enable-pcre-study=no" }
+          - {label: "with pcre2, with study, no jit",   opt: "--with-pcre2 --enable-pcre-study=yes" }
+          - {label: "with pcre2, no study, with jit",   opt: "--with-pcre2 --enable-pcre-study=no --enable-pcre-jit" }
+          - {label: "with pcre2, with study, with jit", opt: "--with-pcre2 --enable-pcre-study=yes --enable-pcre-jit" }
+          - {label: "with lua",                         opt: "--with-lua" }
+          - {label: "wo lua",                           opt: "--without-lua" }
+    steps:
+      - name: Setup Dependencies
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get install -y apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
+      - uses: actions/checkout@v2
+      - name: autogen.sh
+        run: ./autogen.sh
+      - name: configure ${{ matrix.configure.label }}
+        run: ./configure ${{ matrix.configure.opt }} 'CFLAGS=-Werror=format-security'
+      - uses: ammaraskar/gcc-problem-matcher@master
+      - name: make
+        run: make -j `nproc`
+      - name: install module
+        run: sudo make install
+      - name: run tests
+        run: make test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
       - name: Setup Dependencies
         run: |
           sudo apt-get update -y -qq
-          sudo apt-get install -y apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
+          sudo apt-get install -y --no-install-recommends apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
       - uses: actions/checkout@v2
       - name: autogen.sh
         run: ./autogen.sh


.github/workflows/ci.yml
```diff
@@ -74,3 +74,39 @@ jobs:
       - name: Show mod_security2 audit log
         if: always()
         run: sudo cat /var/log/apache2/modsec_audit.log
+
```
theseion, line 77: Run the build job until `make install`, then reuse the artefacts for building and testing. Don't build the artefacts twice.

theseion (changes requested):

.github/workflows/ci.yml
```diff
@@ -74,3 +74,39 @@ jobs:
...
+          - {label: "with pcre2, with study, no jit",   opt: "--with-pcre2 --enable-pcre-study=yes" }
+          - {label: "with pcre2, no study, with jit",   opt: "--with-pcre2 --enable-pcre-study=no --enable-pcre-jit" }
+          - {label: "with pcre2, with study, with jit", opt: "--with-pcre2 --enable-pcre-study=yes --enable-pcre-jit" }
+          - {label: "with lua",                         opt: "--with-lua" }
+          - {label: "wo lua",                           opt: "--without-lua" }
+    steps:
+      - name: Setup Dependencies
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get install -y apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
```
fzipi:
```suggestion
          sudo apt-get install -y --no-install-recommends apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
```

.github/workflows/ci.yml
```diff
@@ -74,3 +74,39 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: autogen.sh
+        run: ./autogen.sh
+      - name: configure ${{ matrix.configure.label }}
+        run: ./configure ${{ matrix.configure.opt }} 'CFLAGS=-Werror=format-security'
+      - uses: ammaraskar/gcc-problem-matcher@master
```
fzipi, line 106:
```suggestion
      - uses: ammaraskar/gcc-problem-matcher@0f9c86f9e693db67dacf53986e1674de5f2e5f28 # ratchet:ammaraskar/gcc-problem-matcher@0.3.0
```

.github/workflows/ci.yml
```diff
@@ -74,3 +74,39 @@ jobs:
...
+          - {label: "with pcre2, no study, with jit",   opt: "--with-pcre2 --enable-pcre-study=no --enable-pcre-jit" }
+          - {label: "with pcre2, with study, with jit", opt: "--with-pcre2 --enable-pcre-study=yes --enable-pcre-jit" }
+          - {label: "with lua",                         opt: "--with-lua" }
+          - {label: "wo lua",                           opt: "--without-lua" }
+    steps:
+      - name: Setup Dependencies
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get install -y apache2-dev libxml2-dev liblua5.1-0-dev libcurl4-gnutls-dev libpcre2-dev pkg-config libyajl-dev apache2 apache2-bin apache2-data
+      - uses: actions/checkout@v2
```
fzipi, line 101:
```suggestion
      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # ratchet:actions/checkout@v4
```

fzipi (commented):
I'll suggest you also to pin versions to well-known hashes and do not use `master` for versions unless necessary.

There is a [nice tool](https://github.com/sethvargo/ratchet) to do this.

.github/workflows/ci.yml
airween: You mean I should insert the rebuild module and `make install` part between `install module` and `prepare config` steps? (After line [40](https://github.com/owasp-modsecurity/ModSecurity/blob/0710feb06c190a6c1121e4ae0ea995ff213e131c/.github/workflows/ci.yml#L40))

If yes, then that has one advantage: we do not run twice almost the whole process, and one disadvantage: in case of error we have to open the GH log. Now we can see the name of the task, eg.
```
test-linux (ubuntu-22.04, x32, gcc, with pcre, no study, no jit, --enable-pcre-study=no)
```
vs.
```
build-linux (ubuntu-22.04, x32, gcc, with pcre, no study, no jit, --enable-pcre-study=no)
```

(the whole  list is [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/14953917314/job/42007008689?pr=3379)).

airween:
> I'll suggest you also to pin versions to well-known hashes and do not use `master` for versions unless necessary.
> 
> There is a [nice tool](https://github.com/sethvargo/ratchet) to do this.

Sorry to ask, what is the advantage of using this form? Why isn't good to use the current `master` state?

fzipi:
> > I'll suggest you also to pin versions to well-known hashes and do not use `master` for versions unless necessary.
> > There is a [nice tool](https://github.com/sethvargo/ratchet) to do this.
> 
> Sorry to ask, what is the advantage of using this form? Why isn't good to use the current `master` state?

Supply chain attacks? See:

- https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions, `Pin actions to a full length commit SHA`
- https://michaelheap.com/pin-your-github-actions/

airween:
> Supply chain attacks? See:

Thanks.

If I'm right, if we use hash's then we should have to care with new releases.

I think this PR is outside of this, so can we skip this change now? Later I try to align these as you suggested with other necessary configuration.

fzipi:
Sounds good. Renovatebot can take care of that, let's do it in a different pr then.

fzipi (approved):

.github/workflows/ci.yml
theseion: > and one disadvantage: in case of error we have to open the GH log
I don't see that as an issue.

I would continue to use two jobs. Upload the build output from the build job and then download it for the test job: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow#passing-data-between-jobs-in-a-workflow.

airween: Sorry, I don't understand this:

> Upload the build output from the build job and then download it for the test job

We **definitely must rebuild** again with different configure arguments (namely without `--enable-assertions`), so if you take a look at line [35](https://github.com/owasp-modsecurity/ModSecurity/blob/0710feb06c190a6c1121e4ae0ea995ff213e131c/.github/workflows/ci.yml#L35) and line [105](https://github.com/owasp-modsecurity/ModSecurity/blob/0710feb06c190a6c1121e4ae0ea995ff213e131c/.github/workflows/ci.yml#L105), you can see that from that point the two jobs are different. We can save (and upload) available data till that point - which (I guess) makes no sense.

theseion: I hadn't realised that the builds differed. How much faster is the test build if you first build for production? Ideally, the compiler only has to recompile a small subset of the code and would be much faster than when running two complete builds. With that in mind, I still think it makes sense to run `make` for the production build, upload everything (including sources), download it in the test job and compile again for test.

I don't know by heart how to express all this in a workflow with a matrix, so maybe it's too complicated. It's just my idea of how I would like to structure the pipeline.

airween: > How much faster is the test build if you first build for production?

It's not significant.

> Ideally, the compiler only has to recompile a small subset of the code and would be much faster than when running two complete builds.

This is not the case. Let me show you an example - here is a small C code:
```
#include <stdio.h>
#include <assert.h>

int main() {

    int * ip = NULL;

    assert(ip != NULL);

    return 0;
}
```
Compile it as a "**non**-production" code:
```
gcc asserttest.c -o asserttest
```
Now run it:
```
./asserttest 
asserttest: asserttest.c:8: main: Assertion `ip != NULL' failed.
Aborted
```
This is what we expected. This is what we need in our CI.

Now compile it as "production" ready - in this case we add a `-DNDEBUG` flag to the compiler - which compiles the C files:
```
gcc -DNDEBUG asserttest.c -o asserttest
```
and run that:
```
./asserttest
```
(Nothing happens).

As you can see we must recompile all the C source files with `-DNDEBUG` flag, therefore there is no any subset of the code that we can re-use.

At this moment I don't see any other solution. We definitely need this `make test` workflow. As I mentioned [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3379#issuecomment-2870028028) we can tune the whole CI later, but there are some other related PR's, so if the job itself is fine, then I wouldn't change it now.

theseion: Ok, thanks for checking.

theseion (approved):

PR #3390, opened 2025-05-22
Buildsystem fixes

arvedarved:
## what

* These changes fix the build for srcdir != builddir.  At several points in the buildsystem relative (to the Makefile) paths were used. this does not work if a clean builddir is used, as the Makefile and generated headers reside in the builddir. auttools provide the variables `$(top_srcdir)` and `$(top_builddir)` to refer to the srcdir and builddir respectively
* If libxml2 is installed in a non-default location linking to libxml2 fails due to a missing LIBXML2_LDFLAGS
* autotools don't support wildcards (see link to documentation below). 

* Note: Because these fixes use a GNU make extension, the build for non-GNU make is broken. If non-GNU make is supported by modsecurity, most likely the files need to be added explicitly. automake warns about this portability, thats why Werror was removed from automake options
* Note2: I only fixed my configuration, there might be other optional parts of modsecurity, that make the same assumptions, which I did not touch. 

## references

* https://www.gnu.org/software/automake/manual/html_node/Wildcards.html  Automake documentation, why wildcards are not supported.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -63,7 +63,7 @@ cppcheck:
 		--enable=warning,style,performance,portability,unusedFunction,missingInclude \
 		--inconclusive \
 		--template="warning: {file},{line},{severity},{id},{message}" \
-		-I headers -I . -I others -I src -I others/mbedtls/include \
+		-I headers -I . -I $(top_srcdir)/others -I $(top_srcdir)/src -I $(top_srcdir)/others/mbedtls/include \
 		--error-exitcode=1 \
 		-i "src/parser/seclang-parser.cc" -i "src/parser/seclang-scanner.cc" \
 		-i others \
--- a/configure.ac
+++ b/configure.ac
@@ -42,7 +42,7 @@ AC_PREFIX_DEFAULT([/usr/local/modsecurity])
 
 
 # General automake options.
-AM_INIT_AUTOMAKE([-Wall -Werror foreign subdir-objects])
+AM_INIT_AUTOMAKE([-Wall foreign subdir-objects])
 
 
 # Check for dependencies (C++, AR, Lex, Yacc and Make)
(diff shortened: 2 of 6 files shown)


airween:
Hi @arvedarved,

first of all, thank you for your first contribution! :tada: 

Unfortunately there are a few issues in build processes in case of Linux and MacOS.

Like [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/15192497010/job/42730061397?pr=3390) one - could you take a look at them?

Thanks!

arvedarved:
Hi Ervin, Thanks for looking at the pull request. Indeed the change in the test directory looks dubious, I tried to fix it. 
I assume the regression tests are run automatically? If not could you trigger them again?

airween:
Hi @arvedarved,

> I assume the regression tests are run automatically? If not could you trigger them again?

yes, the regression tests run automatically - normally. But while this is your first contribution, I have to approve the run. If you're done with it, and the PR will be merged, then your PR's will start the tests automatically.

airween:
Hi @arvedarved,

thanks again for this patch,

>  * Note: Because these fixes use a GNU make extension, the build for non-GNU make is broken. If non-GNU make is supported by modsecurity, most likely the files need to be added explicitly. automake warns about this portability, thats why Werror was removed from automake options
> 
>  * Note2: I only fixed my configuration, there might be other optional parts of modsecurity, that make the same assumptions, which I did not touch.

I approve this PR. but beside of that, it seems like we need to review all of the `automake` related part.

Could you help us with this?

airween (approved):

airween:
For some reason one of the test cases were failed after the merge check, see [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/15505441074/job/43659625776#step:10:98) action log.

I try to reproduce the problem.

PR #3391, opened 2025-05-27
Fix resource leaks in `msc_status_engine_mac_address`

amezin:
## what

Close the socket created and free the memory allocated in `msc_status_engine_mac_address()` in all code paths.

## why

Linux:

The socket is used only temporarily. Current implementation leaks the socket. For example, if `SecRemoteRules` is used anywhere in the config, `systemctl reload httpd`/`systemctl reload apache2` increases the count of open sockets in all Apache processes (well, it happens in the main process, and forked ones just inherit the problem).

Other platforms:

`goto end` just jumped over the calls freeing the memory.

### DIFF
--- a/apache2/msc_status_engine.c
+++ b/apache2/msc_status_engine.c
@@ -182,7 +182,7 @@ int DSOLOCAL msc_status_engine_mac_address (unsigned char *mac)
                 (unsigned char)LLADDR(sdl)[3],
                 (unsigned char)LLADDR(sdl)[4],
                 (unsigned char)LLADDR(sdl)[5]);
-            goto end;
+            break;
         }
     }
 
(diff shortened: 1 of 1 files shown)


amezin:
Hm... After looking a bit more, there seems to be a leak on all platforms. But on Windows and Mac, it's just a memory leak. Maybe `goto end` should be replaced by `break` everywhere?

airween:
Hi @amezin,

> Hm... After looking a bit more, there seems to be a leak on all platforms. But on Windows and Mac, it's just a memory leak. Maybe `goto end` should be replaced by `break` everywhere?

I can check this next weekend. I'll try to add some test to check this PR (or feel free to add some - see `.github` directory).

amezin:
BTW v3 also has the same issue - https://github.com/owasp-modsecurity/ModSecurity/pull/3392 - because the code (for mac address retrieval) is mostly the same

As far as I understand, there are windows/mac tests on v3

airween (approved):

airween:
> BTW v3 also has the same issue - #3392 - because the code (for mac address retrieval) is mostly the same

Thanks again,
 
> As far as I understand, there are windows/mac tests on v3

Yes, v3 has an excellent test framework (but that's a library). And these tests can be run on any platform where the library can be built. In GH workflow, we run the tests on Linux (32 and 64 bit), Mac OSX and Windows.

I'm going to merge this PR - thanks again!

PR #3392, opened 2025-05-27
Fix memory/socket leak in `UniqueId::ethernetMacAddress()`

amezin:
## what

Don't jump over `close()`/`free()` calls. Release the resources correctly in all code paths.

## why

Currently, there are memory, and, in case of Linux, socket leaks. `goto end` jumps over the calls that are supposed to release the memory/close the socket.

## references

Initially discovered this on v2 - https://github.com/owasp-modsecurity/ModSecurity/pull/3391 - but turns out the code is the same on v3 too.


airween:
Hi @amezin,

nice catch, many thanks for this PR!

airween (approved):

PR #3393, opened 2025-05-28
Add hostname to JSON log

JakubOnderka:
## what

Add hostname to JSON log

## why

Hostname that is set by `msc_set_request_hostname` method is not included in JSON log.


airween (approved):
LGTM

airween:
Nice catch, thank you @JakubOnderka!

PR #3398, opened 2025-06-07
fix: change the format of the ascTime() function

airween:
## what

This PR changes the format of [utils::string::ascTime()](https://github.com/owasp-modsecurity/ModSecurity/commit/6248ac1c166b22f4de82680c5ceb26377d1d4e72#diff-e71d5c46f8ce20d4b7e76eb815fc9a3866ab8c854d62e54b9e2fa92bee98be13R77).

The function is used only in one place, in [transaction.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/6248ac1c166b22f4de82680c5ceb26377d1d4e72/src/transaction.cc#L1568), it produces the field `time_stamp` if the audit log format is JSON, for eg.:

```json
{"transaction":{"client_ip":"200.249.12.31","time_stamp":"Sat Jun 07 11:46:23 2025"}}
```

## why

The old (actual) format of that function is `%c`, which is the expected format, but as the documentation writes:

_writes **standard date and time string**, e.g. Sun Oct 17 04:41:13 2010 (locale dependent)_

it's **locale dependent**. This means if the locale uses only 1 digit for the day, then the test is failed, see [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3390#issuecomment-2952210396) comment. The log in that case will look like this:

```json
{"transaction":{"client_ip":"200.249.12.31","time_stamp":"Sat Jun  7 11:46:23 2025"}}
```
Please note that there are two spaces and 1 digit.

This modification changes the format and if the day is below 10, then adds a leading zero, without extra space.

## references

C++ reference of [strftime](https://en.cppreference.com/w/c/chrono/strftime).

## other notes

**Please review this PR carefully as it may break audit log file JSON parsers if the server localization uses a single-digit date representation.**

### DIFF
--- a/src/utils/string.h
+++ b/src/utils/string.h
@@ -74,7 +74,7 @@ inline std::string ascTime(const time_t *t) {
     struct tm timeinfo;
     localtime_r(t, &timeinfo);
     char tstr[std::size("Www Mmm dd hh:mm:ss yyyy")];
-    strftime(tstr, std::size(tstr), "%c", &timeinfo);
+    strftime(tstr, std::size(tstr), "%a %b %d %H:%M:%S %Y", &timeinfo);
     return tstr;
 }
 
--- a/src/utils/string.h
+++ b/src/utils/string.h
@@ -74,7 +74,11 @@ inline std::string ascTime(const time_t *t) {
     struct tm timeinfo;
     localtime_r(t, &timeinfo);
     char tstr[std::size("Www Mmm dd hh:mm:ss yyyy")];
-    strftime(tstr, std::size(tstr), "%a %b %d %H:%M:%S %Y", &timeinfo);
+    // `%c` is equivalent to `%a %b %e %T %Y` with a fixed length, even though zero-padded
+    // month of day is optional, depending on the locale. This would lead to an empty space, e.g.,
+    // `Sat Jun  7 11:46:23 2025` (notice the two spaces before the day of month).
+    // Use `%d` instead and enforce padding.
+    strftime(tstr, std::size(tstr), "%a %b %d %T %Y", &timeinfo);
     return tstr;
 }
 


airween:
@JakubOnderka could you take a review too?

theseion:
I was confused as to why this fix is necessary, so I looked into the LLVM implementation. It turns out that the use `%a %b %e %T %Y` (`%e` instead of `%c`) but just use a fixed size for the output string:
https://github.com/llvm/llvm-project/blob/995d74f8663edb2e20f2d2672556582a6f4cc3f1/libc/src/time/strftime_core/composite_converter.h#L180. That explains the additional space.

theseion (approved):

src/utils/string.h
```diff
@@ -74,7 +74,7 @@ inline std::string ascTime(const time_t *t) {
     struct tm timeinfo;
     localtime_r(t, &timeinfo);
     char tstr[std::size("Www Mmm dd hh:mm:ss yyyy")];
-    strftime(tstr, std::size(tstr), "%c", &timeinfo);
+    strftime(tstr, std::size(tstr), "%a %b %d %H:%M:%S %Y", &timeinfo);
```
theseion:
```suggestion
    // `%c` is equivalent to `%a %b %e %T %Y` with a fixed length, even though zero-padded
    // month of day is optional, depending on the locale. This would lead to an empty space, e.g.,
    // `Sat Jun  7 11:46:23 2025` (notice the two spaces before the day of month).
    // Use `%d` instead and enforce padding.
    strftime(tstr, std::size(tstr), "%a %b %d %T %Y", &timeinfo);
```

theseion (changes requested):

airween:
> so I looked into the LLVM implementation.

Thanks for reviewing that.

Moreover, your suggestion to `%T` makes sense, this escaped my attention, thanks too.

Unfortunately, after my initial commit the error [still persists](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/15507349086/job/43663833303#step:10:98), even on Linux (with `gcc`).

airween:
Hmm... all test have passed. It's very weird.

Note, that the original problem was (in my opinion) that previously the test [expected](https://github.com/owasp-modsecurity/ModSecurity/blob/169e719e7a639482ef96c21622debda84f81bc9e/test/test-cases/regression/auditlog.json#L215) this format: `\S{3} \S{3} \d{2}`, but yesterday the output was `Sat Jun  7`, which didn't match.

JakubOnderka:
@airween I am not sure if changing that is the best idea as even the new format will still be problematic. There is no good solution for that, but maybe the best will be include new field (maybe something like like `timestamp_iso`) that will include timestamp in ISO 8601 format (`2025-06-09T12:30:00Z`), that is much easier to parse and also contains timezone (that is crucial when comparing logs from different systems).

airween:
@JakubOnderka,

thanks for reviewed this.

> @airween I am not sure if changing that is the best idea as even the new format will still be problematic. There is no good solution for that, but maybe the best will be include new field (maybe something like like `timestamp_iso`) that will include timestamp in ISO 8601 format (`2025-06-09T12:30:00Z`), that is much easier to parse and also contains timezone (that is crucial when comparing logs from different systems).

ah, my bad... I thought this issue is a serious one, and I didn't understand why was it [failed](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/15505441074/job/43659625776#step:10:98). I assumed this tests is an old one, but meanwhile I realized that this is a new test - you added that in one of your [recent commit](https://github.com/owasp-modsecurity/ModSecurity/commit/6408bf9237f3e7fd6c8db795792756724cb99d42).

So we can keep this format as is, but please change that [test pattern](https://github.com/owasp-modsecurity/ModSecurity/commit/6408bf9237f3e7fd6c8db795792756724cb99d42#diff-2f716529bbbde1819ae8d43de16f74764ea92ee37cdeda2f7036badd8afa41b7R215):
```
      \"time_stamp\":\"\\S{3} \\S{3} \\d{2} \\d{2}:\\d{2}:\\d{2} \\d{4}\"",
```

The first `\\d{2}` is not good there.

airween:
Closing this PR, we should align the test, not the function.

The affected test is [this](https://github.com/owasp-modsecurity/ModSecurity/commit/6408bf9237f3e7fd6c8db795792756724cb99d42#diff-2f716529bbbde1819ae8d43de16f74764ea92ee37cdeda2f7036badd8afa41b7R215).

Issue #3399, opened 2025-06-10
Outdated product version in the MSI installer


skvoboo-gh:
**Describe the bug**

The **Product Version** property in the MSI Installer is 2.9.7.

**To Reproduce**

[installer.wxs](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/iis/installer.wxs#L10)

**Expected behavior**

The **Product Version** property in the MSI Installer must match the ModSecurity version (the latest released v2 version is 2.9.10).
The MSI installer must be published in the ModSecurity release assets.

**Server (please complete the following information):**
 - ModSecurity version (and connector): 2.9.10 MSI installer
 - WebServer: IIS
 - OS (and distro): Windows

airween:
Hi @skvoboo-gh,

thanks for reporting this. I sent a PR to fix this issue, and added the task into our release procedure.

Unfortunately we can update the version string only in the next release. Sorry for the inconvenience.

skvoboo-gh:
Please note that you will need to bump the **Product Version** in MSI Installer every time you upgrade ModSecurity version. Therefore, in the next release, you should bump the **Product Version** in MSI Installer again.

Don't forget to publish the MSI installer (ModSecurityIIS_2.x.y-64b-64.msi) in the ModSecurity next release assets please.

airween:
> Please note that you will need to bump the **Product Version** in MSI Installer every time you upgrade ModSecurity version. Therefore, in the next release, you should bump the **Product Version** in MSI Installer again.

yes, we do. As I wrote above, I added this step into our release procedure documentation.

> Don't forget to publish the MSI installer (ModSecurityIIS_2.x.y-64b-64.msi) in the ModSecurity next release assets please.

Well, I can't promise this. I don't have Windows or IIS. Furthermore, I'm not familiar in IIS, so I don't know what are the expectations. I'm not a fan of pre-packaged assets, because there are many options which use must decide in compilation time. And this is what I can't decide...

airween:
@skvoboo-gh do you have any other question? If not, could we close this issue?

skvoboo-gh:
I have no other questions regarding this issue.
But I'll create another issue, because no one but you will be able to build ModSecurity 2.x for Windows. The task is impossible (requiring precompiled Apache in the same **outdated** Visual Studio is overkill).

Issue #3397, opened 2025-06-06
2,9,7 Appl Error 1000 in w3wp.exe Module ucrtbase.dll


FrankWarius:
I uploade an image on WordPress 6.8.1 / PHP 8.4. 7 / Win Srv 2025 and IIS crashes. I tried it several times, always the same result

  AppName w3wp.exe 
  AppVersion 10.0.26100.1882 
  AppTimeStamp c6bd45bd 
  ModuleName ucrtbase.dll 
  ModuleVersion 10.0.26100.4061 
  ModuleTimeStamp 8ff2ad62 
  ExceptionCode c0000409 
  FaultingOffset 000000000001cba8 
  ProcessId 0xa4d0 
  ProcessCreationTime 0x1dbd34709f41b85 
  AppPath c:\windows\system32\inetsrv\w3wp.exe 
  ModulePath C:\WINDOWS\System32\ucrtbase.dll 
  IntegratorReportId 789362e0-4e34-48e8-9d01-7c419221a70b 
  PackageFullName  
  PackageRelativeAppId  

**Logs and dumps**

**To Reproduce**

Try to upload this special image (4.5 MB, 4032 x 3024, 72dpi)

**Expected behavior**

No crash

**Server (please complete the following information):**
 - ModSecurity version v2.9.7-64b-64.msi
 - WebServer: IIS on  Windows Server 2025 Datacenter, Version 24H2, Build 26100.4061

**Rule Set (please complete the following information):**
 out of the box, without any changes

**Additional context**

I found https://answers.microsoft.com/en-us/windows/forum/all/windows-11-ucrtbasedll-module-error-solution/811a79cc-da02-4d65-95fc-fdec9bbc02df with hint:

1.- Open Windows Settings
2.- Navigate to Apps > Installed Apps
3.- Locate the following:
• Microsoft Visual C++ 20XX Redistributable - x64
• Microsoft Visual C++ 20XX Redistributable - x86
4.- Uninstall

and uninstalled ModSecurity. 

The Upload of  the image worked.

airween:
Hi @FrankWarius,

I think if the cause is the module, then we can reproduce it on other platform too. Could you sent the file which triggers this error to modsecurity@owasp.org?

FrankWarius:
why do ship and install such an old MS Visual c++ version. 

i primary use 
- MS Visual c++ 2015-2002 x64 /x32 14.44.35208
also installed
- 2010 x64/x32 10.0.40219
[...]

PR #3400, opened 2025-06-10
chore: bump version in MSI installer.wxs

airween:
## what

This PR sets the correct version string in MSI installer file.

## why

There is a user report about the wrong version string.

## references

See #3399 for the report.

### DIFF
--- a/iis/installer.wxs
+++ b/iis/installer.wxs
@@ -7,7 +7,7 @@
     lightArgs: 
 -->
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.7" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
+    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.10" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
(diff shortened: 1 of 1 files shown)


iis/installer.wxs
```diff
@@ -7,7 +7,7 @@
     lightArgs: 
 -->
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.7" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
+    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.10" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
         <Package Description="ModSecurityISS" Comments="none" InstallerVersion="405" Compressed="yes" InstallPrivileges="elevated" InstallScope="perMachine" />
         <?define ProductName = "ModSecuirty IIS" ?>
```
fzipi:
```suggestion
        <?define ProductName = "ModSecurity IIS" ?>
```
typo

fzipi (approved):
This file should be cleaned up. There is a lot of commented stuff (e.g. old crs versions). Also, do you know which pcre version is used by default in the windows build? It should be using the newer version also, right?

iis/installer.wxs
airween: Thanks, committed.

airween:
> This file should be cleaned up. 

Yes, definitely, it should.

> There is a lot of commented stuff (e.g. old crs versions). Also, do you know which pcre version is used by default in the windows build? It should be using the newer version also, right?

I have no idea. Unfortunately, I don't have any Windows and can't check the build flow. I know this is a big problem, because there is another Windows related issue (#3397). But for that (and for this too) we need not just a Windows, but an IIS too.

fzipi (approved):

theseion (approved):

PR #3401, opened 2025-06-11
Plug memory leak when msre_op_validateSchema_execute() exits normally (validateSchema)

nic-prgs:
validateSchema rule leaks memory under normal exit

Add explicit xmlSchemaFreeParserCtxt call

Because sometimes memory is not an infinite resource


airween (approved):
Nice catch, thanks!

airween:
Thanks @nic-prgs!

Fortunately, v3 [does not affected](https://github.com/owasp-modsecurity/ModSecurity/blob/31507404e6a6c0da46a85b478301fc73b8c202d4/src/operators/validate_schema.cc#L115) by this issue.

PR #3411, opened 2025-07-02
fix: remove unwanted '\0' string terminator from argument's value

airween:
## what

This PR removes the unwanted `\0` character from the end of parsed argument value, if the `SecParseXmlIntoArgs` feature is enabled and if the node is empty.

## why

Now if the node is empty the engine adds an extra byte to the argument's value, which is the string terminator. That character will be part of the value, so if the node is `<foo></foo>`, the value will be `\0` with length 1. Then the parsed argument will be `xml.foo=\x00` - which can cause false positives.


airween:
Cc: @RedXanadu

theseion (approved):

RedXanadu (approved):
I've tested this over the past few days and can confirm this PR resolves the issue where stray null characters were appearing in ARGS and causing lots of false positives.

No false positives and no unexpected null chars after applying this change.

Issue #3410, opened 2025-07-02
Error in msc_status_engine.c


Greentears:
**Describe the bug**

In the apache2/msc_status_engine.c:176

The variable `i` is compared to 6 but `i` is not defined.
This `i` was removed in this commit
https://github.com/owasp-modsecurity/ModSecurity/commit/9d9a72734988bbd4f101cbea0d5df60eabcc1435#diff-1c8f9a366617177bdf8e978ba4be1230c2086c1f4d2338cb5e14e21b8162504e

I'm not sure if it is still compiling with the DARWIN flag

airween:
Hi @Greentears,

thanks, you're right, that's a bug.

And it seems that we should extend our CI with an OSX build to avoid this kind of issues. I'm going to fix this soon.

airween:
@Greentears,

unfortunately I don't have any Darwin environment so I can't check other errors - is there any other issue which blocks the build flow?

Greentears:
Hello @airween, 
Sorry I actually don't know, I was just reviewing changes done those last few months

airween:
I made some research and this is what I found:
* [here](https://github.com/owasp-modsecurity/ModSecurity/blob/0c6a661c696e0cb248cf79e32a20aaccb95c3926/apache2/msc_status_engine.c#L128-L144) is the initial version of this file/feature
* and [here](https://github.com/owasp-modsecurity/ModSecurity/blob/410aca9d78db14e5c18667377c70e03e37deef1a/apache2/msc_status_engine.c#L157-L178) is where the author removed the loop which used variable `i`

Therefore the `i` is not necessary at all, because it's a constant, its value never changes. I'm going to prepare the fix.

airween:
See #3412.

PR #3412, opened 2025-07-02
fix: remove unused condition from msc_status_engine.c

airween:
## what

This PR removes an unused condition from `msc_status_engin.c` if the target is `DARWIN`.

## why

In one of a previous [commit](https://github.com/owasp-modsecurity/ModSecurity/commit/9d9a72734988bbd4f101cbea0d5df60eabcc1435#diff-1c8f9a366617177bdf8e978ba4be1230c2086c1f4d2338cb5e14e21b8162504eL167) we removed some unused variables, but it seems like it's still in use.

Unfortunately this was not revealed because that part is only used if the target is `DARWIN`.

## references

* issue: #3410 
* initial [commit](https://github.com/owasp-modsecurity/ModSecurity/blob/0c6a661c696e0cb248cf79e32a20aaccb95c3926/apache2/msc_status_engine.c#L128-L144)
* remove loop which used that variable [commit](https://github.com/owasp-modsecurity/ModSecurity/blob/410aca9d78db14e5c18667377c70e03e37deef1a/apache2/msc_status_engine.c#L157-L178)

Thanks @Greentears.

### DIFF
--- a/apache2/msc_status_engine.c
+++ b/apache2/msc_status_engine.c
@@ -173,7 +173,7 @@ int DSOLOCAL msc_status_engine_mac_address (unsigned char *mac)
     for ( ifap = ifaphead; ifap; ifap = ifap->ifa_next ) {
         struct sockaddr_dl* sdl = (struct sockaddr_dl*)ifap->ifa_addr;
         if ( sdl && ( sdl->sdl_family == AF_LINK ) && ( sdl->sdl_type == IFT_ETHER )
-                && mac[0] && mac[1] && mac[2] && i < 6) {
+                && mac[0] && mac[1] && mac[2]) {
 
             apr_snprintf(mac, MAC_ADDRESS_SIZE, "%02x:%02x:%02x:%02x:%02x:%02x",
                 (unsigned char)LLADDR(sdl)[0],


theseion:
[This](https://github.com/owasp-modsecurity/ModSecurity/commit/9d9a72734988bbd4f101cbea0d5df60eabcc1435#diff-1c8f9a366617177bdf8e978ba4be1230c2086c1f4d2338cb5e14e21b8162504e) is the commit where variable was removed.

theseion (approved):

airween:
> [This](https://github.com/owasp-modsecurity/ModSecurity/commit/9d9a72734988bbd4f101cbea0d5df60eabcc1435#diff-1c8f9a366617177bdf8e978ba4be1230c2086c1f4d2338cb5e14e21b8162504e) is the commit where variable was removed.

I think I wrote the same in my initial commit:
> In one of a previous [commit](https://github.com/owasp-modsecurity/ModSecurity/commit/9d9a72734988bbd4f101cbea0d5df60eabcc1435#diff-1c8f9a366617177bdf8e978ba4be1230c2086c1f4d2338cb5e14e21b8162504eL167) we removed some unused variables, but it seems like it's still in use.

:smiley:

Btw thanks for the reviewing!

PR #3418, opened 2025-07-14
fix: clean matched vars after chained and non-chained rule

### LINKED ISSUE #3382: Semantic of MATCHED_VARS / MATCHED_VARS_NAMES
Hi, 

I have a question about the exact intended semantic of MATCHED_VARS / MATCHED_VARS_NAMES

The [documentation](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#matched_vars_names) is a little bit vague with

> Similar to MATCHED_VAR_NAME except that it is a collection of all matches for the current operator check.
```
SecRule ARGS pattern "chain,deny,id:28"
  SecRule MATCHED_VARS_NAMES "@eq ARGS:param"
```

And all the examples and also the CoreRuleset always have this form: The second rule in the chain is accessing the matches from the first rule. So far, so good.

But what should happen if we have multiple rules (without chain)

From the documentation and my intuition, I would assume that the request `GET /?foo=1&bar=2&baz=2` against these rules
```
SecRule ARGS "@rx 1" "id:1, phase:1, pass"
SecRule ARGS "@rx 2" "id:2, phase:1, pass"
SecRule MATCHED_VARS_NAMES "@contains ARGS:foo" "id:3, phase:1, deny, status:403"
```
would **not block in rule 3**, because `MATCHED_VARS_NAMES` should only contain  the matches from rule 2: `ARGS:bar` and `ARGS:baz` at this point in time.

My assumption is:

After execution of rule 1:
 - MATCHED_VAR_NAME = "ARGS:foo", 
 - MATCHED_VARS_NAMES = { "ARGS:foo" }

After execution of rule 2:
 - MATCHED_VAR_NAME = "ARGS:baz" (or "ARGS:bar" depending on the order)
 - MATCHED_VARS_NAMES = { "ARGS:bar", "ARGS:baz" }
 
So the 3rd rule should not match.

But when I add this as a regression test
```
[
  {
    "enabled":1,
    "version_min":300000,
    "title":"Testing Variables :: MATCHED_VARS_NAMES (3/3)",
    "request":{
      "uri":"/?foo=1&bar=2&baz=2",
      "method":"GET"
    },
    "expected":{
     "http_code": 200
    },
    "rules":[
      "SecRuleEngine On",
      "SecRule ARGS \"@rx 1\" \"id:1,phase:1,pass\"",
      "SecRule ARGS \"@rx 2\" \"id:2,phase:1,pass\"",
      "SecRule MATCHED_VARS_NAMES \"@contains ARGS:foo\" \"id:3,phase:1,deny,status:403\""
    ]
  }
]
```

and run it with the latest build
```
commit aab47091b1b95add953a54bbd8769540d2dfc0fe (HEAD -> v3/master, origin/v3/master, origin/HEAD)
Merge: 990d99b1 797f7dc4
Date:   Sun May 11 10:54:47 2025 +0200
```

the test is failing with

```
Test failed. From: test-cases/regression/avi.json.
Test name: Testing Variables :: MATCHED_VARS_NAMES (3/3).
Reason: 
HTTP code mismatch. expecting: 200 got: 403

Debug log:
[174732390351.534433] [] [4] Initializing transaction
[174732390351.534433] [] [4] Transaction context created.
[174732390351.534433] [] [4] Starting phase CONNECTION. (SecRules 0)
[174732390351.534433] [] [9] This phase consists of 0 rule(s).
[174732390351.534433] [] [4] Starting phase URI. (SecRules 0 + 1/2)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Adding request argument (GET): name "foo", value "1"
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Adding request argument (GET): name "bar", value "2"
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Adding request argument (GET): name "baz", value "2"
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Starting phase REQUEST_HEADERS.  (SecRules 1)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] This phase consists of 3 rule(s).
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] (Rule: 1) Executing operator "Rx" with param "1" against ARGS.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "1" (Variable: ARGS:foo)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Matched vars updated.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "2" (Variable: ARGS:bar)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "2" (Variable: ARGS:baz)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Rule returned 1.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Running (disruptive)     action: pass.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Running action pass
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] (Rule: 2) Executing operator "Rx" with param "2" against ARGS.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "1" (Variable: ARGS:foo)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "2" (Variable: ARGS:bar)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Matched vars updated.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "2" (Variable: ARGS:baz)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Matched vars updated.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Rule returned 1.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Running (disruptive)     action: pass.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Running action pass
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] (Rule: 3) Executing operator "Contains" with param "ARGS:foo" against MATCHED_VARS_NAMES.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "ARGS:foo" (Variable: MATCHED_VARS_NAMES:ARGS:foo)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Matched vars updated.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "ARGS:bar" (Variable: MATCHED_VARS_NAMES:ARGS:bar)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Target value: "ARGS:baz" (Variable: MATCHED_VARS_NAMES:ARGS:baz)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Rule returned 1.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Running action: status
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Running (disruptive)     action: deny.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Running action deny
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Skipping this phase as this request was already intercepted.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Appending request body: 0 bytes. Limit set to: 0.000000
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Starting phase REQUEST_BODY. (SecRules 2)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] This phase consists of 0 rule(s).
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Starting phase RESPONSE_HEADERS. (SecRules 3)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] This phase consists of 0 rule(s).
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] Appending response body: 0 bytes. Limit set to: 0.000000
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Starting phase RESPONSE_BODY. (SecRules 4)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Response body is disabled, returning... 2
[174732390351.534433] [/?foo=1&bar=2&baz=2] [4] Starting phase LOGGING. (SecRules 5)
[174732390351.534433] [/?foo=1&bar=2&baz=2] [9] This phase consists of 0 rule(s).
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Checking if this request is suitable to be saved as an audit log.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Checking if this request is relevant to be part of the audit logs.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [5] Audit log engine was not set.
[174732390351.534433] [/?foo=1&bar=2&baz=2] [8] Request was relevant to be saved. Parts: 4430

Error log:
ModSecurity: Warning. Matched "Operator `Rx' with parameter `1' against variable `ARGS:foo' (Value: `1' ) [file "avi.json"] [line "2"] [id "1"] [rev ""] [msg ""] [data ""] [severity "0"] [ver ""] [maturity "0"] [accuracy "0"] [hostname ""] [uri "/"] [unique_id "174732390351.534433"] [ref "o0,1v10,1"]
ModSecurity: Warning. Matched "Operator `Rx' with parameter `2' against variable `ARGS:baz' (Value: `2' ) [file "avi.json"] [line "3"] [id "2"] [rev ""] [msg ""] [data ""] [severity "0"] [ver ""] [maturity "0"] [accuracy "0"] [hostname ""] [uri "/"] [unique_id "174732390351.534433"] [ref "o0,1v16,1o0,1v22,1"]
[client ] ModSecurity: Access denied with code 403 (phase 1). Matched "Operator `Contains' with parameter `ARGS:foo' against variable `MATCHED_VARS_NAMES:ARGS:foo' (Value: `ARGS:foo' ) [file "avi.json"] [line "4"] [id "3"] [rev ""] [msg ""] [data ""] [severity "0"] [ver ""] [maturity "0"] [accuracy "0"] [hostname ""] [uri "/"] [unique_id "174732390351.534433"] [ref "o0,8v30,8"]

```

I would consider this as a bug because apparently the collection MATCH_VARS_NAMES is never cleaned and collect all the data from all the previous rules.

If this is intended, I think at least a documentation update is needed and maybe the CoreRuleset rules should be adapted (other project, I know)

airween:
## what

This PR changes the code behavior: now the engine cleans the `MATCHED_VAR*` variables after chained and non-chained rules too.

## why

Until now if there was a single (non-chained) rule, and if any of the `MATCHED_VAR*` variable were filled, then the next rule which used them accessed the filled value, even the rule does not use `chain` action.

## references

See issue #3382.

This PR fixes #3382.

## other notes

please see commit [5572ac0](https://github.com/airween/ModSecurity/commit/5572ac0798c2afb728c5bc665900636edb8db91c); I added this change because the first test on Windows was [failed](https://github.com/airween/ModSecurity/actions/runs/16276603396/job/45956909406#step:8:309). It seems like the argument processing order is non-deterministic, at least it's different on Windows (see the log: the first argument is the last from the `QUERY_STRING`, and the tests were success on all other platforms).

### DIFF
--- a/src/rule_with_operator.cc
+++ b/src/rule_with_operator.cc
@@ -355,6 +355,9 @@ end_exec:
 
     /* last rule in the chain. */
     performLogging(trans, ruleMessage, true, true);
+    if (m_ruleId > 0) {
+        cleanMatchedVars(trans);
+    }
     return true;
 }
 
--- a/test/test-cases/regression/variable-MATCHED_VAR.json
+++ b/test/test-cases/regression/variable-MATCHED_VAR.json
@@ -2,7 +2,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing Variables :: MATCHED_VAR (1/2)",
+    "title":"Testing Variables :: MATCHED_VAR (1/5)",
     "client":{
       "ip":"200.249.12.31",
       "port":123
@@ -42,7 +42,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing Variables :: MATCHED_VAR (2/2)",
+    "title":"Testing Variables :: MATCHED_VAR (2/5)",
     "client":{
       "ip":"200.249.12.31",
       "port":123
@@ -80,6 +80,129 @@
 
       "SecRule MATCHED_VAR \"@contains other_value\" \"id:29,pass\"",
       "SecRule MATCHED_VAR \"@contains other_value\" \"id:30,pass\""
+    ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing Variables :: MATCHED_VAR (3/5)",
+    "client":{
+      "ip":"200.249.12.31",
+      "port":123
+    },
+    "server":{
+      "ip":"200.249.12.31",
+      "port":80
+    },
+    "request":{
+      "headers":{
+        "Host":"localhost",
+        "User-Agent":"curl/7.38.0",
+        "Accept":"*/*"
+      },
+      "uri":"/?foo=1&bar=2&baz=2",
+      "method":"GET"
+    },
+    "response":{
+      "headers":{
+        "Date":"Mon, 13 Jul 2015 20:02:41 GMT",
+        "Last-Modified":"Sun, 26 Oct 2014 22:33:37 GMT",
+        "Content-Type":"text/html"
+      },
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "http_code": 200
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule ARGS \"@rx 1\" \"id:1,phase:1,pass\"",
+      "SecRule ARGS \"@rx 2\" \"id:2,phase:1,pass\"",
+      "SecRule MATCHED_VAR \"@eq 1\" \"id:3,phase:1,deny,status:403\""
+    ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing Variables :: MATCHED_VAR (4/5)",
+    "client":{
+      "ip":"200.249.12.31",
+      "port":123
+    },
+    "server":{
+      "ip":"200.249.12.31",
+      "port":80
+    },
+    "request":{
+      "headers":{
+        "Host":"localhost",
+        "User-Agent":"curl/7.38.0",
+        "Accept":"*/*"
+      },
+      "uri":"/?foo=1&bar=2&baz=2",
+      "method":"GET"
+    },
+    "response":{
+      "headers":{
+        "Date":"Mon, 13 Jul 2015 20:02:41 GMT",
+        "Last-Modified":"Sun, 26 Oct 2014 22:33:37 GMT",
+        "Content-Type":"text/html"
+      },
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "http_code": 200
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule ARGS \"@rx 1\" \"id:1,phase:1,pass\"",
+      "SecRule ARGS \"@rx 2\" \"id:2,phase:1,pass\"",
+      "SecRule MATCHED_VAR \"@eq 2\" \"id:3,phase:1,deny,status:403\""
+    ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing Variables :: MATCHED_VAR (5/5)",
+    "client":{
+      "ip":"200.249.12.31",
+      "port":123
+    },
+    "server":{
+      "ip":"200.249.12.31",
+      "port":80
+    },
+    "request":{
+      "headers":{
+        "Host":"localhost",
+        "User-Agent":"curl/7.38.0",
+        "Accept":"*/*"
+      },
+      "uri":"/?foo=1&bar=2&baz=2",
+      "method":"GET"
+    },
+    "response":{
+      "headers":{
+        "Date":"Mon, 13 Jul 2015 20:02:41 GMT",
+        "Last-Modified":"Sun, 26 Oct 2014 22:33:37 GMT",
+        "Content-Type":"text/html"
+      },
+      "body":[
+        "no need."
+      ]
+    },
+    "expected":{
+      "http_code": 403
+    },
+    "rules":[
+      "SecRuleEngine On",
+      "SecRule ARGS \"@rx 1\" \"id:1,phase:1,pass\"",
+      "SecRule ARGS \"@rx 2\" \"id:2,phase:1,deny,status:403,chain\"",
+      "SecRule MATCHED_VAR \"@eq 2\""
     ]
   }
 ]
--- a/test/test-cases/regression/variable-MATCHED_VARS.json
+++ b/test/test-cases/regression/variable-MATCHED_VARS.json
@@ -2,7 +2,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing Variables :: MATCHED_VARS (1/2)",
+    "title":"Testing Variables :: MATCHED_VARS (1/6)",
     "client":{
       "ip":"200.249.12.31",
       "port":123
@@ -43,7 +43,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing Variables :: MATCHED_VARS (2/2)",
+    "title":"Testing Variables :: MATCHED_VARS (2/6)",
     "client":{
       "ip":"200.249.12.31",
       "port":123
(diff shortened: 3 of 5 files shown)


airween:
@mirkodziadzka-avi could you take a review on this?

MirkoDziadzka (approved):
Looks good to me, thanks.

Can we also update the documentation? I think this is at least as important than this change.

airween:
> Looks good to me, thanks.

Great, thanks!
 
> Can we also update the documentation? I think this is at least as important than this change.

Indeed, after the merge I'm going to update it. Thank you!

fzipi (approved):
Thanks for adding so many tests.

src/rule_with_operator.cc
```diff
@@ -355,6 +355,9 @@ bool RuleWithOperator::evaluate(Transaction *trans,
 
     /* last rule in the chain. */
     performLogging(trans, ruleMessage, true, true);
+    if (m_ruleId > 0) {
```
theseion, line 358: Why this check? If I understand correctly, `m_ruleId == 0` would be an invalid rule (caught exception). Would we even get here then? As far as I can tell, nothing bad would happen if `m_ruleId == 0`, since `cleanMatchedVars()` operates on the transaction only.

airween: > If I understand correctly, `m_ruleId == 0` would be an invalid rule

Yes, except if it's a chained rule. Generally, every chained rule (rules?) has (have?) only one unique `id`. In libmodsecurity3, despite you set up the `id` action at the first rule (in a chained rule), the last rule will own that `id`. Therefore this condition (`m_ruleId > 0`) tells us this is the end of a rule, no matter that's chained or not, we should clean the `MATCHED_*` variables.

theseion: I see. From the code (and the comment) it seems like the condition is redundant though, since `end_exec:` will always be evaluated for the last rule. So I still don't see why the extra condition is necessary.

airween: Because we need to clear `MATCHED_*` variables **only** if the rule is a standalone rule (not chained), or it's the last part of a chained rule.

This means if the rule is part if a chained rule, but not the last, then we don't clean these group of variables.

If you see the diff, we can say **this solves** the issue.

theseion: I understand, and that's exactly what I'm saying... [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3418/files/5572ac0798c2afb728c5bc665900636edb8db91c#diff-85d45620fd71a5a26b40225ff74c6b98da5537337e8f60b8f1f86c6de52e8757R346) will only be true when all chained rules have been processed. [This](https://github.com/owasp-modsecurity/ModSecurity/pull/3418/files/5572ac0798c2afb728c5bc665900636edb8db91c#diff-85d45620fd71a5a26b40225ff74c6b98da5537337e8f60b8f1f86c6de52e8757R332) will only be true when the rule is not chained. In all cases,`end_exec` will only be evaluated when the last rule has been reached, whether it's the last in a chain or a single rule. Hence: `end_exec` is equivalent to `m_ruleId > 0`. Unless I still don't get it...

airween: Let me approach it from a different direction: if that condition were not there, the `MATCHED_*` variables would be deleted after each rule was evaulated. Not just after standalone rules, but in case of every part in chained rules. For eg. take a look at the initial rule in original [issue](https://github.com/owasp-modsecurity/ModSecurity/issues/3382#issue-3066705350):
```
SecRule ARGS pattern "chain,deny,id:28"
  SecRule MATCHED_VARS_NAMES "@eq ARGS:param"
```
and the request:
```
GET /?foo=1&bar=2&baz=pattern
```
If the condition were not there, then the first part of chained rule matches, `MATCHED_*` variables are set, but the engine clears them immediately after the end of processing, therefore the second part of the chained rule wont match, because `MATCHED_VARS_NAMES` is empty.

MirkoDziadzka: 2 cents: 

- I think when there are different opinions what the code is doing, having a test to confirm which interpretation  is correct would be always nice ;-)
- How about clearing matched vars not at the end, but as the first action when starting to evaluate a rule (or chain of rules)? Should be semantically the same and we need only a single place to clean matched vars.

airween: When I worked on this fix I made several tests, this is why I could make this patch :smiley:.

Btw I try to figure out how can I solve the issue with your idea (clear vars **before** the evaluation). (As I remember I tried to go on this way, but I rejected for some reason... Let me check again.)

airween: Okay, I think I know why did I reject that solution.

Let's approach the issue from the transaction's direction. Transaction object calls the `evaluate()` function [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/transaction.cc#L249), for eg. (there are several other call, in each phases).

The `evaluate()` is part of transaction's `m_rules` member, which is a [RuleSet](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set.h#L49) type, see [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/transaction.cc#L119). `m_rules->evaluate()` has two arguments: phase and transaction.

It grabs the rules in phase that called, see [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rules_set.cc#L114), and [iterates through](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rules_set.cc#L135) them. There is an `auto` cast [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rules_set.cc#L138), but the rule which created here can be a [RuleWithActions](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rule_with_actions.cc#L59) or [RuleWithOperator](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rule_with_operator.cc#L54).

But only the `RuleWithOperator` has [cleanMatchedVars](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/src/rule_with_operator.cc#L93).

I didn't want to touch this, therefore I tried to find a solution at the end of the rule processing. But we can find a different solution, eg. we can move the `cleanMatchedVars()` to another class. Then I think we can make clean method easier.

theseion: Thanks for the explanation @airween. I still don't understand why the condition is necessary, but your tests prove that the code works, so that's good enough for me. I suggest you add a comment before the condition that explains what the condition is for. Then I'll be happy.

airween:
Fixed in #3422, closing this one.

PR #3419, opened 2025-07-14
Fix SecRequestBody(NoFiles)Limit overflow

airween:
## what

This PR fixes the possible overflow of `SecRequestBodyLimit` and `SecRequestBodyNoFilesLimit` directives.

## why

There are two reported issues:
* #3356
* #3352 

There are a few problems in handling these config variables:
* the type of variables were `double` before this patch - this makes no sense, there is no 1234.6 bytes...
* the type of variables were signed - also makes sense, there is no negative value
* the parser converted the read value with `atoi()`, which gives an integer with different bitsize

Therefore if a user gave an extra high value, then the `atoi()` converted it into a negative value.

## references

Fixes #3356 and #3352

## side note

Now there is no hardcoded limit for these values. Beside of that if the admin gives `0` (zero) value then the engine allows unlimited bytes.

I want to stop this, which means if the value of these variables aren't set or set to 0, that means the value is `0` and engine will allow only `0` byte. Hopefully this will force the admin to set an explicit value.

Issue #3356, opened 2025-03-31
libModSecurity3: requests are blocked when `SecRequestBodyNoFilesLimit` is set to a very high value


EsadCetiner:
**Describe the bug**

When `SecRequestBodyNoFilesLimit` is set to a very high value, all requests get incorrectly blocked as being too large. This does not happen on Apache.

**Logs and dumps**

```
---zcw5rmqN---A--
[31/Mar/2025:16:45:39 +1100] 174339993925.897781 ::1 52958 ::1 8080
---zcw5rmqN---B--
POST / HTTP/1.1
Host: localhost:8080
User-Agent: curl/8.5.0
Accept: */*
Content-Length: 10
Content-Type: application/x-www-form-urlencoded

---zcw5rmqN---C--
test=hello

---zcw5rmqN---D--

---zcw5rmqN---E--
<html>\x0d\x0a<head><title>400 Bad Request</title></head>\x0d\x0a<body>\x0d\x0a<center><h1>400 Bad Request</h1></center>\x0d\x0a<hr><center>nginx/1.24.0 (Ubuntu)</center>\x0d\x0a</body>\x0d\x0a</html>\x0d\x0a

---zcw5rmqN---F--
HTTP/1.1 400
Server: nginx/1.24.0 (Ubuntu)
Date: Mon, 31 Mar 2025 05:45:39 GMT
Content-Length: 166
Content-Type: text/html
Connection: close

---zcw5rmqN---H--
ModSecurity: Access denied with code 400 (phase 2). Matched "Operator `Eq' with parameter `0' against variable `REQBODY_ERROR' (Value: `1' ) [file "/etc/modsecurity/modsecurity.conf"] [line "69"] [id "200002"] [rev ""] [msg "Failed to parse request body."] [data "Request body excluding files is bigger than the maximum expected."] [severity "2"] [ver ""] [maturity "0"] [accuracy "0"] [hostname "::1"] [uri "/"] [unique_id "174339993925.897781"] [ref "v0,1"]
```

According to the debug log, it looks like an integer overflow is happening? I've attached the full debug log just in case.
```
[174339993925.897781] [/] [9] Appending request body: 10 bytes. Limit set to: 13107200.000000
[174339993925.897781] [/] [4] Starting phase REQUEST_BODY. (SecRules 2)
[174339993925.897781] [/] [5] Request body excluding files is bigger than the maximum expected. Limit: -2147483648.000000
[...]

PR #3420, opened 2025-07-23
fix: integer type limits

### LINKED ISSUE #3352: curl with PUT method can upload a huge file (>2GB) on remote server but hangs and timeout
**Describe the bug**
When we want to upload a huge file (>2GB) to a remote web site (Artifactory in our case) , our http client hangs indefinitely.
However, the file is correctly uploaded to Artifactory.

**Logs and dumps**
Output of:
 1. DebugLogs (level 9), cf attachments
 2. AuditLogs : this file is empty
 3. Error logs : 

**To Reproduce**
On a GNU/Linux machine do : 
```bash
dd if=/dev/urandom of=/tmp/2G.txt bs=1M count=2048
curl -u <myuser>:<my_api_key> -T /tmp/2G.txt 'https://artifactory.local/artifactory/generic/perf_debug/2G.txt
```
Then, the `curl`hangs.

**Expected behavior**
`curl` ends whithout error.

**Server :**
 - ModSecurity version (and connector): `ModSecurity` `v3.0.12` and `3.0.14` with `nginx-connector` `v1.0.3`
 - WebServer: `nginx-1.18.0`
 - OS: `GNU/Linux`, `Ubuntu 22.04`

 - Running any public or commercial rule set? [Core Security Rules](https://github.com/coreruleset/coreruleset), version `4.5.0`

**Additional context**

* The infrastructure is made like : `client -> haproxy -> nginx(+ modesec) -> Artifactory`
* More than the file size , the time to upload is more than 250 seconds when the upload fails
* The problem persists when we disable `Core Security Rules` usage
* The only way to avoid the problem is to set `modsecurity off;` in my vhost configuration
* Artifactory return a `HTTP 201`
* HTTP method used is `PUT`
* strace -> see attachments
* TCP segment missing when modsec is enabled (on nginx machine, interface connected to haproxy machine) : 
```
69719  272.819868  172.16.3.13  172.16.3.12  TCP  66  443 → 36912 [FIN, ACK] Seq=6734 Ack=2104577102 Win=2021120 Len=0 TSval=55543961 TSecr=2937796858
```
**Attachments:**
- [2000m_debug.log](https://github.com/user-attachments/files/19294975/2000m_debug.log)
- [final_strace.log](https://github.com/user-attachments/files/19295097/final_strace.log)

airween:
## what

This PR fixes the handling of integer type limits, eg. `SecRequestBodyNoFilesLimit`.

**Changes**:
* in `headers/modsecurity/rules_set_properties.h` I implemented a new template with name `ConfigValue`
* this implements the `parse()` method which reads safely the integer value from a string, which parsed by engine's SecLang parser (Bison)
* the template also sets the default `min` and `max` value to `0`
* this method checks the type limits, eg. if the type (for eg. `ConfigInt`) implements signed integer, then the min and max limits will be -32767 and +32767 (`INT_MIN` and `INT_MAX`)
* out-of-bounds values will trigger an error at startup with a message in error.log
* with help of this template I implemented three types: `ConfigInt`, `ConfigUnsignedInt` and `ConfigUnsignedLong`
* it's enough to set the min and max values in all new types (eg. if we want to add new integer types) - see existing new types; I used `std::numeric_limits<T>::min()` and `std::numeric_limits<T>::max()`
* I changed the type of variables: `m_argumentsLimit`, `m_requestBodyJsonDepthLimit`, `m_requestBodyLimit`, `m_requestBodyNoFilesLimit`, `m_responseBodyLimit`, `m_pcreMatchLimit`, `m_uploadFileLimit` and `m_uploadFileMode`;  the current types are [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L507-L514); now there is no more `ConfigDouble` type variable
* in `src/parser/seclang-parser.yy` I changed the type in grammar parser; there you can see the error handling (detailed error messages are in template's `parse()` method)
* I also regenerated `src/parser/seclang-parser.cc` (it's a generated file by Bison, don't need to check it)

**Other changes**:
* I fixed some SonarCloud issues in case of `ConfigDouble`, `ConfigString` and `ConfigSet` implementations (also in `headers/modsecurity/rules_set_properties.h`)
* in top of `headers/modsecurity/rules_set_properties.h` I had to add undefining macro directives if the target is `WIN32`, because the CI workflow was unsuccessful on Windows, because - probably - `min` and `max` are Windows C++ macros

## why

The problem with current solution is that the type of these variables are `ConfigDouble`, see the [variable declarations](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L507-L511) and their [type implementation](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L93).

A quick review of current implementation:
* the type of variables were double before this patch - this makes no sense, there is no 1234.6 bytes...
* the type of variables were signed - also makes sense, there is no negative value
* the parser converted the read value with `atoi()`, which gives an integer with different bitsize

Therefore if a user gave an extra high value, then the `atoi()` converted it into a negative value.

Previously I sent another PR (#3419), but I realized that solution wasn't complete. In this PR I reworked the whole storage implementation.

## references

Fixes #3352 and #3356.

### DIFF
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -614,8 +614,8 @@ class RulesSetProperties {
     ConfigXMLParseXmlIntoArgs m_secXMLParseXmlIntoArgs;
     ConfigBoolean m_tmpSaveUploadedFiles;
     ConfigBoolean m_uploadKeepFiles;
-    ConfigDouble m_argumentsLimit;
-    ConfigDouble m_requestBodyJsonDepthLimit;
+    ConfigUnsignedInt m_argumentsLimit;
+    ConfigUnsignedInt m_requestBodyJsonDepthLimit;
     ConfigUnsignedLong m_requestBodyLimit;
     ConfigUnsignedLong m_requestBodyNoFilesLimit;
     ConfigUnsignedLong m_responseBodyLimit;
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -13,6 +13,14 @@
  *
  */
 
+#ifdef WIN32
+#ifdef max
+#undef max
+#endif
+#ifdef min
+#undef min
+#endif
+#endif
 
 #ifdef __cplusplus
 #include <ctime>
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -111,17 +111,18 @@ public:
         }
         catch (const std::invalid_argument&) {
             // probably can't occur, but we handle it anyway
-            if (errmsg) *errmsg = "Invalid number format (not numeric)";
+            set_error(errmsg, "Invalid number format (not numeric)");
             return false;
         }
         catch (const std::out_of_range&) {
-            if (errmsg) *errmsg = "Number out of range";
+            // the value is out of range, we can not handle it
+            set_error(errmsg, "Number out of range");
             return false;
         }
         catch (...) {
             // we don't need to handle all exceptions, the engine's BISON parser
             // does not allow other symbols than numbers
-            if (errmsg) *errmsg = "An unknown error occurred while parsed the value.";
+            set_error(errmsg, "An unknown error occurred while parsed the value.");
             return false;
         }
 
@@ -135,7 +136,7 @@ public:
             (val > static_cast<unsigned long long>(maxValue()))
 
         ) {
-            if (errmsg) *errmsg = "Value is too big.";
+            set_error(errmsg, "Value is too big.");
             return false;
         }
 
@@ -149,7 +150,7 @@ public:
                 ||
                 (val < static_cast<unsigned long long>(minValue()))
             ) {
-                if (errmsg) *errmsg = "Value is too small.";
+                set_error(errmsg, "Value is too small.");
                 return false;
             }
         }
@@ -161,7 +162,7 @@ public:
                 ||
                 (val < static_cast<long long>(minValue()))
             ) {
-                if (errmsg) *errmsg = "Value is too small.";
+                set_error(errmsg, "Value is too small.");
                 return false;
             }
         }
@@ -178,6 +179,11 @@ protected:
     virtual T maxValue() const = 0;
     // minValue is optional
     virtual T minValue() const { return 0; }
+
+private:
+    static inline void set_error(std::string* err, const char* msg) {
+        if (err) *err = msg;
+    }
 };
 
 /** @ingroup ModSecurity_CPP_API */
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -119,7 +119,7 @@ public:
             set_error(errmsg, "Number out of range");
             return false;
         }
-        catch (...) {
+        catch (...) { // NOSONAR
             // we don't need to handle all exceptions, the engine's BISON parser
             // does not allow other symbols than numbers
             set_error(errmsg, "An unknown error occurred while parsed the value.");
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -22,6 +22,7 @@
 #include <list>
 #include <set>
 #include <cstring>
+#include <limits>
 #endif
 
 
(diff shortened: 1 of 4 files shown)


airween:
Okay, I think I'm done with this.

First: @jonathansmith9485, @SonNgo2211, @EsadCetiner - please take a review this.

Please read the PR's description carefully, eg. if you set an out-of-bound value, then the engine will drop an error and won't start.

If you could make a test, please try to decrease the limit and check the limit works as you expected too.

Please note, that this PR fixes other similar issues too, eg. `SecArgumentsLimit` - but those values aren't too big typically, like `SecRequestBodyNoFilesLimit` and `SecRequestBodyLimit`. So, any help are welcome with wide tests.

SonNgo2211:
Hi @airween,
Thank you so much for your great work on this fix! I really appreciate your effort. I plan to try this fix in our production environment soon and will provide feedback if needed. 
Also, I wanted to ask if it’s possible to increase the hard limits for **SecRequestBodyNoFilesLimit** and **SecRequestBodyLimit**?
According to the ModSecurity 3.x documentation I’ve seen, the current hard limits are set at 1GB, and I’m wondering if those can be increased safely or configured for larger values.
Looking forward to your advice. Thanks again!

airween:
Hi @SonNgo2211,

> Also, I wanted to ask if it’s possible to increase the hard limits for **SecRequestBodyNoFilesLimit** and **SecRequestBodyLimit**?
> According to the ModSecurity 3.x documentation I’ve seen, the current hard limits are set at 1GB, and I’m wondering if those can be increased safely or configured for larger values.

Thanks for bringing this up. Unfortunately that's a false information. I assume the author just copied the [v2](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v2.x)#secrequestbodylimit) documentation and left as is in [v3](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-(v3.x)#secrequestbodylimit).

In v3 (libmodsecurity3) there isn't any hardcoded limit at all. I have to fix this in the mentioned documentation.

EsadCetiner:
@airween I just tested the PR and I'm still observing the same behavior, however unlike your previous PR, the directive is now overflowing instead of being set to 0

``
[175335928862.517382] [/remote.php/dav/files/esadc/] [5] Request body excluding files is bigger than the maximum expected. Limit: -2147483648.000000
``

> Please read the PR's description carefully, eg. if you set an out-of-bound value, then the engine will drop an error and won't start.

What values are considered out of bounds? I have `SecRequestBodyNoFilesLimit` set to `10737418240` and I didn't get any errors when restarting NGINX

airween:
> I just tested the PR and I'm still observing the same behavior, however unlike your previous PR, the directive is now overflowing instead of being set to 0
> 
> `[175335928862.517382] [/remote.php/dav/files/esadc/] [5] Request body excluding files is bigger than the maximum expected. Limit: -2147483648.000000`

Hmmm... sorry to ask, but are you sure you pulled down the correct branch and tested **that** code?

I tried with your settings:
```
$ grep ^SecRequestBodyNoFilesLimit /etc/nginx/modsecurity.conf
SecRequestBodyNoFilesLimit 10737418240
```
and send your original request:
```
curl -v -d "test=hello" localhost
```
Then I see only these lines in debug.log:
```
[175336092253.135496] [/] [4] Starting phase REQUEST_BODY. (SecRules 2)
[175336092253.135496] [/] [4] Adding request argument (POST): name "test", value "hello"
```
With the original code I got the same result as you.

> What values are considered out of bounds? I have `SecRequestBodyNoFilesLimit` set to `10737418240` and I didn't get any errors when restarting NGINX

The type of this variable is `unsigned long`, which means the maximum value on 64 bit systems is 18446744073709551615. The usable interval is between 0 and this value - all other values are out of bounds.

Could you build and run this C++ code on your machine? If yes, what do you get?
```C++
// build: g++ ulonglimit.cpp -o ulonglimit
#include <limits>
#include <iostream>

int main() {
    std::cout << std::numeric_limits<unsigned long>::max() << std::endl;
    return 0;
}
```
My result:
```
$ ./ulonglimit 
18446744073709551615
```

It's **very-very weird** that you get this message: `Limit: -2147483648.000000`. This means your code still uses the old convert method, and the type of this variable is **still a signed type** - otherwise it couldn't be a negative value. I thought you use a 32 bit system, and unsigned long is a less bit-size type, but in that case there wouldn't be a negative value with this patch.

EsadCetiner:
@airween 

> Hmmm... sorry to ask, but are you sure you pulled down the correct branch and tested that code?

I pulled this branch down `git clone airween/ModSecurity --branch v3/secreqbodylimit` 

I'm on a 64 bit system, so I'm not hitting the limit
```
./ulonglimit
18446744073709551615
```

> This means your code still uses the old convert method, and the type of this variable is still a signed type

That shouldn't be the case since I removed the old binaries installed via apt, I'll try compiling it in a completely fresh environment, play with it a bit there, then report back.

airween:
> I pulled this branch down `git clone airween/ModSecurity --branch v3/secreqbodylimit`

cool,
 
> I'm on a 64 bit system, so I'm not hitting the limit
> 
> ```
> ./ulonglimit
> 18446744073709551615
> ```

okay, thanks,

> I removed the old binaries installed via apt

Please note that if you use libmodsecurity3 through APT then the installed Nginx module looks the library under `/lib/x86_64-linux-gnu/`
```
$ ldd /usr/lib/nginx/modules/ngx_http_modsecurity_module.so | grep modsecurity
	libmodsecurity.so.3 => /lib/x86_64-linux-gnu/libmodsecurity.so.3 (0x00007fe5a473f000)
```
If you don't set the path before you run `./configure`, then the new library will be installed under `/usr/local/`, and the Nginx module will use the old one. Or remove the current symlink (`/lib/x86_64-linux-gnu/libmodsecurity.so.3` is a symbolic link) and create a new one to your new `.so` file.

Anyway, now I see I have to change the types, because on 32 bit systems the `unsigned long` is just a 32 bit wide.

EsadCetiner:
@airween 
> Please note that if you use libmodsecurity3 through APT then the installed Nginx module looks the library under /lib/x86_64-linux-gnu/

Thanks for that hint, I ran `apt remove` instead of `apt autopurge` which removes all traces of the packages. I've been testing it for a few minutes and everything is working as it should, the problem was me😅.

I set a stupidly large limit for `SecRequestsBodyNoFilesLimit` and I'm getting an error when running `nginx -t`, so I know for sure I'm using the right binaries.

airween:
> Thanks for that hint, I ran `apt remove` instead of `apt autopurge` which removes all traces of the packages. I've been testing it for a few minutes and everything is working as it should, the problem was me😅.

no worries, it's a very good news that it works as well.
 
> I set a stupidly large limit for `SecRequestsBodyNoFilesLimit` and I'm getting an error when running `nginx -t`, so I know for sure I'm using the right binaries.

Cool.

Meanwhile I realized that I (and the previous authors) used `long` and `int` types, but the problem is that these types width are different on different architectures. Now I'm closing this PR and opened a new one where I use `int64_t` and `uint64_t` - which are the same on different architectures.

I'll ask a help again - thanks again.

PR #3421, opened 2025-07-24
fix: integer type limits, 2nd try

### LINKED ISSUE #3352: curl with PUT method can upload a huge file (>2GB) on remote server but hangs and timeout

airween:
## what

This PR fixes the handling of integer type limits, eg. `SecRequestBodyNoFilesLimit`.

**Changes**:
* in `headers/modsecurity/rules_set_properties.h` I implemented a new template with name `ConfigValue`
* this implements the `parse()` method which reads safely the integer value from a string, which parsed by engine's SecLang parser (Bison)
* the template also sets the default `min` and `max` value to `0`
* this method checks the type limits, eg. if the type (for eg. `ConfigInt`) implements `int32_t` (`signed int` on AMD64), then the min and max limits will be -2147483648 and +2147483647
  * if you take a look at this minimum value, it can be familiar with the mentioned log message in [#3356](https://github.com/owasp-modsecurity/ModSecurity/issues/3356#issue-2959545090): `Request body excluding files is bigger than the maximum expected. Limit: -2147483648.000000`; the reason is (I write that below too) that the current implementation uses `atoi()` function, which converts a `string` to `signed int`, but if the value is bigger than the max value above, then it overflows and casts to `int`'s minimum
* out-of-bounds values will trigger an error at startup with a message in error.log
* with help of this template I implemented three types: `ConfigInt` (`int32_t`), `ConfigUnsignedInt` (`uint32_t`) and `ConfigUnsignedLong` (`uint64_t`)
* it's enough to set the min and max values in all new types (eg. if we want to add new integer types) - see existing new types; I used `std::numeric_limits<T>::min()` and `std::numeric_limits<T>::max()`
* I changed the type of variables: `m_argumentsLimit`, `m_requestBodyJsonDepthLimit`, `m_requestBodyLimit`, `m_requestBodyNoFilesLimit`, `m_responseBodyLimit`, `m_pcreMatchLimit`, `m_uploadFileLimit` and `m_uploadFileMode`;  the current types are [here](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L507-L514); now there is no more `ConfigDouble` type variable
* in `src/parser/seclang-parser.yy` I changed the type in grammar parser; there you can see the error handling (detailed error messages are in template's `parse()` method)
* I also regenerated `src/parser/seclang-parser.cc` (it's a generated file by Bison, don't need to check it)

**Other changes**:
* I fixed some SonarCloud issues in case of `ConfigDouble`, `ConfigString` and `ConfigSet` implementations (also in `headers/modsecurity/rules_set_properties.h`)
* in top of `headers/modsecurity/rules_set_properties.h` I had to add undefining macro directives if the target is `WIN32`, because the CI workflow was unsuccessful on Windows, because - probably - `min` and `max` are Windows C++ macros

**How does it work**
The most important part is the new `parse()` method. This decides what integer type is used in the derived class (signed/unsigned), and uses the necessary function (`stoll()` / `stoull()`). Then it checks that the conversion was success with the biggest sized integer. If not, then the `errno` will be `1` and the value will be the max/min value of used type (`int64_t` / `uin64_t`). There is another check: does the parsed value fit into the chosen type (eg. `int32_t`). If any condition matches, then the method returns with false at startup time.

## why

The problem with current solution is that the type of these variables are `ConfigDouble`, see the [variable declarations](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L507-L511) and their [type implementation](https://github.com/owasp-modsecurity/ModSecurity/blob/cf24aeaeadefd758d0220e485bd8e312945f06df/headers/modsecurity/rules_set_properties.h#L93).

A quick review of current implementation:
* the type of variables were double before this patch - this makes no sense, there is no 1234.6 bytes...
* the type of variables were signed - also makes sense, there is no negative value
* the parser converted the read value with `atoi()`, which gives an integer with different bitsize

Therefore if a user gave an extra high value, then the `atoi()` converted it into a negative value.

Previously I sent another PR (#3419), but I realized that solution wasn't complete. In this PR I reworked the whole storage implementation. Then I sent a new one (#3420), but there I continued to use `int` and `unsigned long` types, but I realized those are not the same on different platforms (amd64 vs i386). In this PR I changed the types with fixed width types.

## references

Fixes #3352 and #3356.

### DIFF
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -196,22 +196,6 @@ protected:
     }
 };
 
-class ConfigDouble {
- public:
-    bool m_set = false;
-    double m_value = 0.0;
-    ConfigDouble() = default;
-
-    void merge(const ConfigDouble *from) {
-        if (m_set == true || from->m_set == false) {
-            return;
-        }
-        m_set = true;
-        m_value = from->m_value;
-        return;
-    }
-};
-
 
 class ConfigString {
  public:
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -88,7 +88,9 @@ public:
     ConfigValue() = default;
 
     void merge(const ConfigValue<T>* from) {
-        if (m_set || !from->m_set) return;
+        if (m_set || !from->m_set) {
+            return;
+        }
         m_set = true;
         m_value = from->m_value;
     }
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -166,7 +166,9 @@ protected:
 
 private:
     static inline void set_error(std::string* err, const char* msg) {
-        if (err) *err = msg;
+        if (err) {
+            *err = msg;
+        }
     }
 };
 
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -130,10 +130,10 @@ public:
         }
 
         if (
-            // first condition will be true if the value is bigger than int64/uint64 max value
-            // the second condition checks if the value is fit as int64/uint64, but not fit for
-            // designed type, eg. uint32; in that case the errno will be 0, but
-            // we must check the value is not bigger than the given max() at the type class
+            // The first condition will be true when the value is bigger than int64/uint64 maximum value.
+            // The second condition checks whether the value fits into int64/uint64, but not
+            // into the designed type, e.g., uint32; in that case the errno will be 0, but
+            // we must check the value is not bigger than the defined maximum of the class.
             (errno == ERANGE && val == std::numeric_limits<LimitSigned>::max())
             ||
             (val > static_cast<LimitSigned>(maxValue()))
--- a/headers/modsecurity/rules_set_properties.h
+++ b/headers/modsecurity/rules_set_properties.h
@@ -13,6 +13,14 @@
  *
  */
 
+#ifdef WIN32
+#ifdef max
+#undef max
+#endif
+#ifdef min
+#undef min
+#endif
+#endif
 
 #ifdef __cplusplus
 #include <ctime>
@@ -22,6 +30,8 @@
 #include <list>
 #include <set>
 #include <cstring>
+#include <limits>
+#include <cstdint>
 #endif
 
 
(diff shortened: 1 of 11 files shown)


airween:
@jonathansmith9485, @SonNgo2211, @EsadCetiner - here is the new one, please take a review this.

EsadCetiner:
@airween I've been testing for a day or so and so far I haven't had any issues

airween:
> @airween I've been testing for a day or so and so far I haven't had any issues

Great news, thank you!

theseion:
Can you elaborate on why this caused #3352 and #3356?

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,29 +78,127 @@ class Driver;
...
+    T m_value = 0;
 
-    void merge(const ConfigInt *from) {
-        if (m_set == true || from->m_set == false) {
-            return;
-        }
+    ConfigValue() = default;
+
+    void merge(const ConfigValue<T>* from) {
+        if (m_set || !from->m_set) return;
```
theseion: Personally, I very much hate unbraced if bodies. It's just too easy to make mistakes.

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,29 +78,127 @@ class Driver;
...
+        }
+        catch (const std::out_of_range&) {
+            // the value is out of range, we can not handle it
+            set_error(errmsg, "Number out of range");
+            return false;
+        }
+        catch (...) { // NOSONAR
+            // we don't need to handle all exceptions, the engine's BISON parser
+            // does not allow other symbols than numbers
+            set_error(errmsg, "An unknown error occurred while parsed the value.");
```
theseion:
```suggestion
            set_error(errmsg, "An unknown error occurred while parsing number.");
```

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,29 +78,127 @@ class Driver;
...
+
+    // default parser
+    bool parse(const std::string& a, std::string* errmsg = nullptr) {
+
+        // use an alias type because the template can convert both signed and unsigned int
+        using LimitSigned = std::conditional_t<std::is_signed_v<T>, std::int64_t, std::uint64_t>;
+        LimitSigned val;
+
+        // clear errno variable, wee need that later
+        errno = 0;
```
theseion, line 106: There's only this single assignment to `errno`, which means the conditions further below will never by true when checking `errno`.

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,29 +78,127 @@ class Driver;
...
+            // does not allow other symbols than numbers
+            set_error(errmsg, "An unknown error occurred while parsed the value.");
+            return false;
+        }
+
+        if (
+            // first condition will be true if the value is bigger than int64/uint64 max value
+            // the second condition checks if the value is fit as int64/uint64, but not fit for
+            // designed type, eg. uint32; in that case the errno will be 0, but
+            // we must check the value is not bigger than the given max() at the type class
```
theseion:
```suggestion
            // The first condition will be true when the value is bigger than int64/uint64 maximum value.
            // The second condition checks whether the value fits into int64/uint64, but not
            // into the designed type, e.g., uint32; in that case the errno will be 0, but
            // we must check the value is not bigger than the defined maximum of the class.
```

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,29 +78,127 @@ class Driver;
...
+};
+
+class ConfigUnsignedLong : public ConfigValue<uint64_t> {
+protected:
+    uint64_t maxValue() const override {
+        return std::numeric_limits<uint64_t>::max();
+    }
+};
 
 class ConfigDouble {
```
theseion: It doesn't look like this class is used anymore. Can't w just delete it?

theseion (changes requested):
Much better!

airween:
> Can you elaborate on why this caused #3352 and #3356?

I explained in my initial comment:

>   * if you take a look at this minimum value, it can be familiar with the mentioned log message in [#3356](https://github.com/owasp-modsecurity/ModSecurity/issues/3356#issue-2959545090): `Request body excluding files is bigger than the maximum expected. Limit: -2147483648.000000`; the reason is (I write that below too) that the current implementation uses `atoi()` function, which converts a `string` to `signed int`, but if the value is bigger than the max value above, then it overflows and casts to `int`'s minimum

headers/modsecurity/rules_set_properties.h
airween: Fixed in 7e3c807.

headers/modsecurity/rules_set_properties.h
airween: No, `error` is not a local variable, it's used by globally. Here [stoll()](https://github.com/owasp-modsecurity/ModSecurity/blob/9a8ce8b5f5753144b2bce003d54c6a5c0d8e6f22/headers/modsecurity/rules_set_properties.h#L108) and/or [stoull](https://github.com/owasp-modsecurity/ModSecurity/blob/9a8ce8b5f5753144b2bce003d54c6a5c0d8e6f22/headers/modsecurity/rules_set_properties.h#L110) set this variable, if any error occurred.

Please take a look at their [reference](https://en.cppreference.com/w/cpp/string/basic_string/stol#Exceptions):

> * [std::out_of_range](https://en.cppreference.com/w/cpp/error/out_of_range.html) if the converted value would fall out of the range of the result type or if the underlying function ([std::strtol](https://en.cppreference.com/w/cpp/string/byte/strtol.html) or [std::strtoll](https://en.cppreference.com/w/cpp/string/byte/strtol.html)) sets [errno](https://en.cppreference.com/w/cpp/error/errno.html) to [ERANGE](https://en.cppreference.com/w/cpp/error/errno_macros.html).

headers/modsecurity/rules_set_properties.h
airween: Btw: you're right, I usually try to avoid these syntax in code.

I think we should create a syntax convention and a new job in GH workflow to avoid these syntax.

headers/modsecurity/rules_set_properties.h
airween: I was thinking that I remove that, but I kept that because if we needed that by any reason, we can use it again.

Well, Git will remember this class - only we need to remember :smiley:.

airween: Removed in abed778.

theseion (approved):

headers/modsecurity/rules_set_properties.h
```diff
@@ -68,46 +78,130 @@ class Driver;
...
+
+protected:
+    // derived classes must implement the maxValue
+    virtual T maxValue() const = 0;
+    // minValue is optional
+    virtual T minValue() const { return 0; }
+
+private:
+    static inline void set_error(std::string* err, const char* msg) {
+        if (err) *err = msg;
```
theseion:
```suggestion
        if (err) {
            *err = msg;
        }
```

theseion (changes requested):

theseion (approved):

PR #3424, opened 2025-07-28
fix: confusing indentation at acmp.cc

wooffie:
## what

Change indentation, which can confuse users

## why

On first sight on code it can be seems that this block in for cycle, but it's not true

### DIFF
--- a/src/utils/acmp.cc
+++ b/src/utils/acmp.cc
@@ -226,7 +226,7 @@ static void acmp_build_binary_tree(ACMP *parser, acmp_node_t *node) {
 
     /* We have array with all children of the node and number of those children
      */
-    for (i = 0; i < count - 1; i++)
+    for (i = 0; i < count - 1; i++) {
         for (j = i + 1; j < count; j++) {
             acmp_node_t *tmp;
 
@@ -236,6 +236,7 @@ static void acmp_build_binary_tree(ACMP *parser, acmp_node_t *node) {
             nodes[i] = nodes[j];
             nodes[j] = tmp;
         }
+    }       
     if (node->btree != NULL) {
         free(node->btree);
         node->btree = NULL;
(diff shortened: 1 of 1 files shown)


src/utils/acmp.cc
```diff
@@ -236,10 +236,10 @@ static void acmp_build_binary_tree(ACMP *parser, acmp_node_t *node) {
             nodes[i] = nodes[j];
             nodes[j] = tmp;
         }
-        if (node->btree != NULL) {
-            free(node->btree);
-            node->btree = NULL;
-        }
+    if (node->btree != NULL) {
```
airween, line 240: I see your idea and you're right, this can be confusing.

But could you add a bit more clarification?

I suggest you to add an open `{` at the and of line [229](https://github.com/owasp-modsecurity/ModSecurity/pull/3424/files#diff-09c8944650ce24e352a8a965ff221eeeb5f788735a89614db64dd6a943fd3d48R229), and a close one before this `if (...) {` block? I think that would make the whole part more clear.

Note, that we can check the behavior through regression test against [@pm](https://github.com/owasp-modsecurity/ModSecurity/blob/5d05ba359cff56adfd0bbefed5f6cab3254f1b4d/test/test-cases/regression/operator-pm.json) and [secrule-language-test/@pm](https://github.com/owasp-modsecurity/secrules-language-tests/blob/a3d4405e5a2c90488c387e589c5534974575e35b/operators/pm.json)

airween (changes requested):

src/utils/acmp.cc
wooffie: Yes, I just check codebase of project, and there are a lot places with alone operator-body in if/for. Changed, now I run tests locally to check behavior =)

wooffie:
Tests look fine

src/utils/acmp.cc
airween: Thanks.

Just fyi: [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3421#discussion_r2233977662) and [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3421#discussion_r2234102075) we discussed similar issues.

Cc @theseion.

airween (approved):

airween:
Thanks @wooffie!

PR #3426, opened 2025-07-29
fix: rx operators better null checks

wooffie:
## what

Add nullptr check in one more place

## why

I think it can cause null derefence, because before we check for null:
https://github.com/owasp-modsecurity/ModSecurity/blob/1ff9f2a943529301a35e188ced84c3f977f3cd96/src/operators/rx.cc#L62

But after forget about this check and do dereference:
https://github.com/owasp-modsecurity/ModSecurity/blob/1ff9f2a943529301a35e188ced84c3f977f3cd96/src/operators/rx.cc#L71

I cannot create PoC, maybe I'm wrong and using  regex without limit will never cause `!= Utils::RegexResult::Ok`, but looks legit

### DIFF
--- a/src/operators/rx.cc
+++ b/src/operators/rx.cc
@@ -68,18 +68,19 @@ bool Rx::evaluate(Transaction *transaction, RuleWithActions *rule,
 
     // FIXME: DRY regex error reporting. This logic is currently duplicated in other operators.
     if (regex_result != Utils::RegexResult::Ok) {
-        transaction->m_variableMscPcreError.set("1", transaction->m_variableOffset);
+        if (transaction) {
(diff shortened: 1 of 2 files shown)


airween:
Hi @wooffie,

thanks for this PR!

> because before we check for null:
> ...
> But after forget about this check and do dereference:

I think the first one is not necessary, because `evaluate()` is called from a `Transaction` object originally. But indeed, checking it won't hurt anyone.

> maybe I'm wrong and using regex without limit will never cause

yes, you're right - checking nullptr value is always a good step.

Thank you!

airween (approved):

PR #3427, opened 2025-07-29
fix: malloc checks at msc_tree

wooffie:
## what

1. Check for malloc success
2. Avoid check for NULL after dereference (at memset)

## why

Avoid NPD

## references

Some more info: #2890

P.S. Maybe at 419 we should return something another?

### DIFF
--- a/src/utils/msc_tree.cc
+++ b/src/utils/msc_tree.cc
@@ -419,6 +419,9 @@ TreeNode *CPTAddElement(unsigned char *ipdata, unsigned int ip_bitmask, CPTTree 
                 node->count++;
                 new_node = node;
                 node->netmasks = reinterpret_cast<unsigned char *>(malloc(node->count *  sizeof(unsigned char)));
+                if (node->netmasks == NULL) {
+                    return node;
+                }
                 memset(node->netmasks, 0, (node->count *  sizeof(unsigned char)));
 
                 if ((node->count -1) == 0) {
@@ -491,14 +494,14 @@ TreeNode *CPTAddElement(unsigned char *ipdata, unsigned int ip_bitmask, CPTTree 
             }
 
             i_node->netmasks = reinterpret_cast<unsigned char *>(malloc((node->count - i) * sizeof(unsigned char)));
-            memset(i_node->netmasks, 0, ((node->count - i) * sizeof(unsigned char)));
-
             if(i_node->netmasks == NULL) {
                 free(new_node->prefix);
                 free(new_node);
                 free(i_node);
                 return NULL;
             }
+
+            memset(i_node->netmasks, 0, ((node->count - i) * sizeof(unsigned char)));
 
             j = 0;
             while (j < (node->count - i))   {


wooffie:
Also, maybe we need NULL check for `node` here?
https://github.com/owasp-modsecurity/ModSecurity/blob/1ff9f2a943529301a35e188ced84c3f977f3cd96/src/utils/msc_tree.cc#L685

airween:
Hi @wooffie,

thanks again for this PR.

Honestly, I would avoid `malloc()` calls in a C++ code. May be it's a bit bigger work, but if we want to modernize the code, we should modify the `TreeNode` structure: replace `unsigned char *` by `std::vector<unsigned char>`, and replace `malloc()` **AND** `memset()` by `node->netmasks.assign(node->count, 0);`.

But of course, to apply these modifications, we have to review all occurrences and places where this structure is used, and align the code.

I assume this change will disappear the SonarCloud quality issue (now you can see in check list that it's failed, because of the code complexity (with the new `if()` it increased that level which Sonar reports).

What do you think?

wooffie:
Yes, no problem, I check, this is one of three files with using malloc in `src`. I can test changes by test, or not?

airween:
> Yes, no problem, I check, this is one of three files with using malloc in `src`.

Thank you!

> I can test changes by test, or not?

I think you can, but may be we have to review existing tests too.

It seems like this classes and structures are used in `@ipMatch` and similar operators, and there is a [test](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/test/test-cases/regression/operator-ipMatchFromFile.json) for this (not exactly, but I think for test is enough), and there are some other tests where the case uses `@ipMatch` operator.

There is a [seclang-test](https://github.com/owasp-modsecurity/secrules-language-tests/blob/a3d4405e5a2c90488c387e589c5534974575e35b/operators/ipMatch.json) file too, you can run that too.

Let me know if you need any help!

wooffie:
@airween we really need to save C-API for msc_tree.h ? 
Like for all structs and so on?

airween:
> @airween we really need to save C-API for msc_tree.h ? Like for all structs and so on?

Sorry, you mean you would rewrite that part and modernize the code? If yes, I think that would be awesome!

Btw I think these two files are legacy from mod_security2 code base (which is pure C source), see [msc_tree.h](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/msc_tree.h) and [msc_tree.c](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/msc_tree.c), this is why this part is in C (too - like many others).

wooffie:
I'm about C-compat with extern "C" linking for this header file. I think we should rewrite in without C. I will start to work around, but can't say when back with some enhancements

airween:
> I'm about C-compat with extern "C" linking for this header file.

But this header is not accessible through (library's) C-API, only library's components access them.

> I think we should rewrite in without C.

Agree.

> I will start to work around, but can't say when back with some enhancements

No problem, as I wrote, it's a legacy and we've been using it for about 8 years or so :)

wooffie:
I close this one, 
If I find time to rewrite it, I'll come back.

PR #3428, opened 2025-07-29
fix: memory leak at msc_tree

wooffie:
## what

Prefix memory can be leaked at place there I suggest patch.
After prefix can be freed by node->prefix, it's okay


airween:
HI @wooffie,

As in case of #3427 here I would like to suggest also a complete review instead of adding "patches" :smiley:.

What do you think?

Issue #3429, opened 2025-07-30
Unable to parse absolute Windows path as value of modsecurity.conf directive


skvoboo-gh:
**Describe the bug**

Unable to parse absolute Windows path as value of modsecurity.conf directive.
`RulesSet::loadFromUri()` failed with error message:

```

Rules error. File: C:\Program Files (x86)\ModSecurity\modsecurity.conf. Line: 3. Column: 59. Invalid input:  SecAuditLog "D:\Data Directory\ModSecurity\logs\audit.log"

```

**To Reproduce**

Load `modsecurity.conf` with content:

```
SecAuditEngine RelevantOnly
SecAuditLogType Serial
SecAuditLog "D:\Data Directory\ModSecurity\logs\audit.log"

```
**Expected behavior**

There are no errors.

**Server (please complete the following information):**
 - ModSecurity version: ModSecurity v3.0.14
 - OS: Windows

airween:
@skvoboo-gh,

thanks for the report. It seems like the parser does not have any rule to parse this directive with `"`, see the source:
#https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/parser/seclang-scanner.ll#L760

Only one syntax is there without quote marks.

Could you try without `"` to confirm this?

skvoboo-gh:
Relative Linux path with `"` can be parsed successfully:

```
SecAuditLog "logs/audit.log"

```

airween:
> Relative Linux path with `"` can be parsed successfully:
> 
> ```
> SecAuditLog "logs/audit.log"
> ```

Sorry, there **is** a rule with `"`:
https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/parser/seclang-scanner.ll#L764

I'll check this out what could be the problem.

skvoboo-gh:
> Could you try without " to confirm this?

RulesSet::loadFromUri() failed with error message:

```

Rules error. File: C:\Program Files (x86)\ModSecurity\modsecurity.conf. Line: 3. Column: 57. Invalid input:  \Data Directory\ModSecurity\logs\audit.log

```

airween:
Thanks, I think I found the problem.

Now the `CONFIG_VALUE_PATH` syntax is [this](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/parser/seclang-scanner.ll#L408):
```
[...]

PR #3430, opened 2025-07-30
fix: add Windows specific characters to config value syntax

airween:
## what

This PR fixes a bug: if the host is a Windows based server, and admin wants to set any path where the directory name contains space (and the separator is the blackslash), then parser throws an error.

These characters have been added:
* space
* `\` (backslash) (it's escaped, so this is it: `\\`
* `(` and `)` (parenthesis) (these are also escaped)

## why

The problem is described in #3429.

## references

#3429


airween:
See #3429 - @skvoboo-gh tested this on Windows.

src/parser/seclang-scanner.ll
```diff
@@ -405,7 +405,7 @@ CONFIG_VALUE_ONLYARGS                   (?i:OnlyArgs)
 CONFIG_VALUE_OFF                        (?i:Off)
 CONFIG_VALUE_ON                         (?i:On)
 CONFIG_VALUE_PARALLEL                   (?i:Parallel|Concurrent)
-CONFIG_VALUE_PATH                       [0-9A-Za-z_\/\.\-\*\:]+
+CONFIG_VALUE_PATH                       [0-9A-Za-z_\/\.\-\*\: \\\(\)]+
```
theseion: You're in a character class here, no need to escape meta characters.
```suggestion
CONFIG_VALUE_PATH                       (?i:[0-9a-z_/.*: \\()-]+)
```

theseion (changes requested):

src/parser/seclang-scanner.ll
airween: That was the original form, I just added the new characters. Let's see how does it work.

Also good to know that Bison has its own regex engine, not PCRE or other 3rd-party. Theoretically it follows POSIX standard, so I hope this changes won't change the behavior.

theseion (dismissed):

src/parser/seclang-scanner.ll
airween: Unfortunately this does not work, I'm afraid we need to escape the sequences too - see failed tests, like [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/16668799425/job/47180283694?pr=3430#step:9:60).

I'm going to revert that commit.

airween:
Okay, I found the problem and fixed it. Recent changes:
* d8574c8: I fixed the regex in the failed regression test; see previous [issue](https://github.com/owasp-modsecurity/ModSecurity/pull/3398#issuecomment-3036899211)
* 013c8c2: I put back @theseion's suggestion
  * it seems like the syntax produced the same tokenizer, because I re-generated the file [seclang-scanner.cc](https://github.com/owasp-modsecurity/ModSecurity/blob/f2170d8576bca3d49e272aed947093201150c321/src/parser/seclang-scanner.cc), and it's the same, no diff

test/test-cases/regression/auditlog.json
```diff
@@ -212,7 +212,7 @@
       ]
     },
     "expected": {
-      "audit_log": "{\"transaction\":{\"client_ip\":\"200.249.12.31\",\"time_stamp\":\"\\S{3} \\S{3} \\d{2} \\d{2}:\\d{2}:\\d{2} \\d{4}\"",
+      "audit_log": "{\"transaction\":{\"client_ip\":\"200.249.12.31\",\"time_stamp\":\"\\S{3} \\S{3} [ \\d]\\d \\d{2}:\\d{2}:\\d{2} \\d{4}\"",
```
theseion, line 215: I don't understand this change. `\d{2}` is identical to `[\d]\d`.

theseion (changes requested):

test/test-cases/regression/auditlog.json
airween: In first 9 days in every month the current pattern does not match with the used syntax, because the day format is a single number with a leading space, like this:

```"Sun Aug  3 09:33:26 2025"```

(There is an extra space before the day `3`)

But later, the format is like this:
```"Sun Aug 10 09:33:26 2025"```

This is why we need to change that syntax and add an optional space or a number.

theseion: Thanks. I didn't see the space in the character class.

theseion (approved):

Issue #3435, opened 2025-08-17
running `./build.sh` on Ubuntu 24.04 gives warnings: "warning: wildcard utils/*.h: non-POSIX variable name"


Danrancan:
**Describe the bug**

When running ./build.sh I get the following warnings:

```
./build.sh
libtoolize: putting auxiliary files in '.'.
libtoolize: copying file './ltmain.sh'
libtoolize: putting macros in AC_CONFIG_MACRO_DIRS, 'build'.
libtoolize: copying file 'build/libtool.m4'
libtoolize: copying file 'build/ltoptions.m4'
libtoolize: copying file 'build/ltsugar.m4'
libtoolize: copying file 'build/ltversion.m4'
libtoolize: copying file 'build/lt~obsolete.m4'
configure.ac:50: installing './compile'
configure.ac:45: installing './missing'
examples/multiprocess_c/Makefile.am: installing './depcomp'
src/Makefile.am:70: warning: wildcard actions/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard actions/ctl/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard actions/data/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard actions/disruptive/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard actions/transformations/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard debug_log/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard audit_log/writer/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard collection/backend/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard operators/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
src/Makefile.am:70: warning: wildcard parser/*.h: non-POSIX variable name
src/Makefile.am:70: (probably a GNU make extension)
[...]

PR #3437, opened 2025-08-25
fix: eliminate non-posix standard variables

airween:
## what

This PR fixes the error that `./build.sh` gives namely the `src/Makefile.am` contains non-POSIX variables.

## why

GNU `automake` does not support `wildcard`, but the project uses that in [src/Makefile.am](https://github.com/owasp-modsecurity/ModSecurity/blob/a07ed61e74b0ba61a35471c4689f082815204887/src/Makefile.am#L70-L85). There is an open issue about this so we should fix that.

## references

#3435 
https://www.gnu.org/software/automake/manual/html_node/Wildcards.html

## fix

As the documentation says above GNU automake expects a programmer should list the files instead of using `wildcard`, I used a hack: in `build.sh` I collect all header files in previously mentioned directories, and generate a file: `src/headers.mk`. This file is included in `src/Makefile.am`. The file is generated all cases when user runs `./build.sh`. The generated file is not part of the source tree - see `.gitignore` changes in this PR.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -4,7 +4,7 @@ rm -rf autom4te.cache
 rm -f aclocal.m4
 
 cd src
-rm -f haders.mk
+rm -f headers.mk
 echo "noinst_HEADERS = \\" > headers.mk
 ls -1 \
     actions/*.h \
--- a/build.sh
+++ b/build.sh
@@ -2,6 +2,28 @@
 
 rm -rf autom4te.cache
 rm -f aclocal.m4
+
+cd src
+rm -f haders.mk
+echo "noinst_HEADERS = \\" > headers.mk
+ls -1 \
+    actions/*.h \
+    actions/ctl/*.h \
+    actions/data/*.h \
+    actions/disruptive/*.h \
+    actions/transformations/*.h \
+    debug_log/*.h \
+    audit_log/writer/*.h \
+    collection/backend/*.h \
+    operators/*.h \
+    parser/*.h \
+    request_body_processor/*.h \
+    utils/*.h \
+    variables/*.h \
+    engine/*.h \
+    *.h | tr "\012" " " >> headers.mk
+cd ../
+
 case `uname` in Darwin*) glibtoolize --force --copy ;;
   *) libtoolize --force --copy ;; esac
 autoreconf --install
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ src/location.hh
 src/position.hh
 src/stack.hh
 src/stamp-h1
+src/headers.mk
 /test/rules_optimization
 /test/regression_tests
 /test/unit_tests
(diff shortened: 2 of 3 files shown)


build.sh
```diff
@@ -2,6 +2,28 @@
 
 rm -rf autom4te.cache
 rm -f aclocal.m4
+
+cd src
+rm -f haders.mk
```
fzipi:
```suggestion
rm -f headers.mk
```

fzipi (commented):
Typo

fzipi:
Two questions:
- why do you need to distribute these? https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/Makefile.am#L37-L58
- isn't your change being captured by https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/Makefile.am#L70-L85 ?

airween:
> * why do you need to distribute these? https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/Makefile.am#L37-L58

These are part of ModSecurity's API (the header files which will be installed under `/usr/include/modsecurity`) and I think `make` knows from these lines what headers need to install.
 
> * isn't your change being captured by https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/Makefile.am#L70-L85 ?

I'm not sure I understand your question here, but yes, these lines are problematic. See the [diff](https://github.com/owasp-modsecurity/ModSecurity/pull/3437/files#diff-4cb884d03ebb901069e4ee5de5d02538c40dd9b39919c615d8eaa9d364bbbd77L70-L85) of the mentioned file. That's the point: eliminate the `wildcard` macro.

fzipi (approved):

build.sh
```diff
@@ -2,6 +2,28 @@
 
 rm -rf autom4te.cache
 rm -f aclocal.m4
+
+cd src
+rm -f headers.mk
```
theseion, line 7: Redundant, since you truncate the file on the next line.

build.sh
```diff
@@ -2,6 +2,28 @@
...
+echo "noinst_HEADERS = \\" > headers.mk
+ls -1 \
```
theseion, line 9: I suppose this is ok, would have preferred `find` maybe.

theseion (changes requested):

build.sh
airween: Yes, but I like this way much better, because if there is any problem occurs during the file creation, then it's possible that the new file won't be created (and the old one will be used). If there is no file in this case, then it's a sign.

build.sh
airween: I was thinking what's the better, but my aspects:
* I wanted to keep the method, I mean what was the previous intent
* `find` finds all `.h` files; if someone adds a new directory, `find` will put those headers too without any notification. I know the chance is very low for this, but just for sure I would like to avoid that.

Honestly, I'm not sure this `make` variable is used during the build/install flow, but I want to maintain until we can be sure that it can be removed.

theseion (approved):

Issue #3441, opened 2025-09-13
libModSecurity3: `REQUEST_HEADERS` names are treated as case sensitive


EsadCetiner:
**Describe the bug**

libModSecurity3 treats `REQUEST_HEADERS` names as case sensitive when they are [supposed to be treaten as case insensitive](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers). This is not the case for ModSecurity2.

**Logs and dumps**

N/A

**To Reproduce**

Create the following rules, the first is meant to simulate a rule exclusion and the second a blocking rule:

```
SecRule REQUEST_FILENAME "@unconditionalMatch" \
    "id:1,\
    phase:1,\
    pass,\
    t:none,\
    ctl:ruleRemoveTargetById=2;REQUEST_HEADERS:referer"

SecRule REQUEST_HEADERS:Referer "@contains <evil-string>" \
    "id:2,\
    phase:1,\
    deny,\
    t:none,\
    log"
```

Then run this curl command, you'll be blocked by the WAF when you should've be allowed:
`curl -H "Referer: <evil-string>" localhost`

If I match the exact case in my rule exclusion then the request is allowed as expected:

```
SecRule REQUEST_FILENAME "@unconditionalMatch" \
    "id:1,\
    phase:1,\
    pass,\
    t:none,\
    ctl:ruleRemoveTargetById=2;REQUEST_HEADERS:Referer"

SecRule REQUEST_HEADERS:Referer "@contains <evil-string>" \
    "id:2,\
    phase:1,\
    deny,\
    t:none,\
    log"
```

**Expected behavior**

Request header names should be treated as case insensitive, just like ModSecurity2.

**Server (please complete the following information):**
- OS: Ubuntu 24.04
- ModSecurity Version: 3.0.14
- NGINX Connector Version: v1.0.4
- NGINX Version: 1.24.0

**Rule Set (please complete the following information):**
 - N/A

**Additional context**

N/A

airween:
Hi @EsadCetiner,

thanks for reporting this.

I think this happens in case of exclusion handling, I mean if the admin adds an exclusion with `ctl:ruleRemoveTargetById` action, then the target will be case sensitive.

I think the behavior is correct if we add a rule like this:
```
SecRule REQUEST_HEADERS:FakeHeader "@rx bash" \
    "id:999112,\
    phase:1,\
    deny,\
    t:none,\
[...]

PR #3442, opened 2025-09-16
fix: ignore case when evaluating exceptions

airween:
## what

This PR fixes the wrong behavior: variable names are treated as case sensitive. Also it contains two new test cases for regression tests.

## why

An issue opened (see below) where the reporter explained the problem: if an exclusion adds a collection member with lowercase, it won't have any effect during the rule evaluation.

The issue mentions only the `REQUEST_HEADERS` collection, but that's true for all targets.

mod_security2 Apache module [uses](https://github.com/owasp-modsecurity/ModSecurity/blob/9d9a72734988bbd4f101cbea0d5df60eabcc1435/apache2/re.c#L118) `strncasecmp()` in all cases, not just `REQUEST_HEADERS`.

## additional notes

I didn't want to mix C and C++ functions, this is why I ignored `strncasecmp()`. I also wanted to avoid copying strings and create new functions to comparing strings, so the only solution seemed to be to use `std::equal()` with a lambda function.

## references

#3441

### DIFF
--- a/test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
+++ b/test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
@@ -133,7 +133,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing CtlRuleRemoveTargetById (5)",
+    "title":"Testing CtlRuleRemoveTargetById (5): lowercase `Referer` header",
     "expected":{
       "http_code": 200
     },
--- a/test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
+++ b/test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
@@ -99,7 +99,7 @@
   {
     "enabled":1,
     "version_min":300000,
-    "title":"Testing CtlRuleRemoveTargetById (4)",
+    "title":"Testing CtlRuleRemoveTargetById (4): uppercase `Referer` header",
     "expected":{
       "http_code": 200
     },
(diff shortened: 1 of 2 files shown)


test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
```diff
@@ -95,5 +95,73 @@
         "SecRule REQUEST_FILENAME \"@endsWith /wp-login.php\" \"id:9002100,phase:2,t:none,nolog,pass,ctl:ruleRemoveTargetById=1;ARGS\"",
         "SecRule ARGS \"@contains lhebs\" \"id:1,phase:3,t:none,status:202,block,deny,tag:'CRS'\""
     ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing CtlRuleRemoveTargetById (4)",
```
theseion:
```suggestion
    "title":"Testing CtlRuleRemoveTargetById (4): uppercase `Referer` header",
```

test/test-cases/regression/action-ctl_rule_remove_target_by_id.json
```diff
@@ -95,5 +95,73 @@
...
+    "rules":[
+        "SecRuleEngine On",
+        "SecRule REQUEST_FILENAME \"@unconditionalMatch\" \"id:1,phase:1,pass,t:none,ctl:ruleRemoveTargetById=2;REQUEST_HEADERS:referer\"",
+        "SecRule REQUEST_HEADERS:Referer \"@contains attack\" \"id:2,phase:1,deny,t:none,log\""
+    ]
+  },
+  {
+    "enabled":1,
+    "version_min":300000,
+    "title":"Testing CtlRuleRemoveTargetById (5)",
```
theseion:
```suggestion
    "title":"Testing CtlRuleRemoveTargetById (5): lowercase `Referer` header",
```

theseion (changes requested):

theseion (approved):

Issue #1988, opened 2018-12-24
Modsecurity IIS remote_address empty in case of IPv6 address - IPmatch: bad IPv4 specification


erikvdijk:
Using Modsecurity 2.9.3 with IIS and OWASP-CRS in combination with IPv6, the IPv6 address does not show up in remote_address and thereby always false-positive.

Auditlog:
`{"transaction":{"time":"24/Dec/2018:10:29:11 +0100","transaction_id":"17221764977212260572","remote_address":"","remote_port":80,"local_address":"127.0.0.1","local_port":80},"request":{"request_line":"GET / HTTP/1.1","headers":{"Connection":"keep-alive","Content-Length":"0","Accept":"text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8","Accept-Encoding":"gzip, deflate","Accept-Language":"nl,en-GB;q=0.7,en;q=0.3","Host":"inter.testipv6.waf","User-Agent":"Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:64.0) Gecko/20100101 Firefox/64.0","DNT":"1","Upgrade-Insecure-Requests":"1"}},"response":{"protocol":"HTTP/1.1","status":0,"headers":{}},"audit_data":{"messages":["IPmatch: bad IPv4 specification \"\".","Rule processing failed (id=, msg=)."],"handler":"IIS","stopwatch":{"p1":15626,"p2":31254,"p3":0,"p4":0,"p5":0,"sr":0,"sw":0,"l":0,"gc":0},"producer":["ModSecurity for IIS (STABLE)/2.9.3 (http://www.modsecurity.org/)","OWASP_CRS/3.1.0"],"server":"ModSecurity Standalone","engine_mode":"ENABLED"}}`

It looks like it is something here:
https://github.com/SpiderLabs/ModSecurity/blob/v2/master/iis/mymodule.cpp#L91
https://github.com/SpiderLabs/ModSecurity/blob/v2/master/iis/mymodule.cpp#L123

Or can this be config?

Reproducable on different machines.

mujiansu:
the function GetIpAddr should provide the ability to handle IPV6, below is my code, it's  not fully tested, but worked
```c
char *GetIpAddr(apr_pool_t *pool, PSOCKADDR pAddr)
{
    DWORD addrSize = pAddr->sa_family == AF_INET ? sizeof SOCKADDR_IN : sizeof SOCKADDR_IN6;
[...]

PR #3443, opened 2025-09-24
fix(iis):  IPv6 Handling in ModSecurity IIS Module

A13501350:
## what

- Replacing the `WSAAddressToString` and manual string parsing approach with the `GetNameInfo` function for IP address retrieval.
- Adding proper support for both IPv4 and IPv6 address families in the `CopySockAddr` function.

## why

The original implementation incorrectly assumed all addresses were IPv4, causing failures when processing IPv6 connections.

## references

- #1988
- Implementation by @mujiansu

### DIFF
--- a/iis/mymodule.cpp
+++ b/iis/mymodule.cpp
@@ -90,63 +90,66 @@ class REQUEST_STORED_CONTEXT : public IHttpStoredContext
 
 char *GetIpAddr(apr_pool_t *pool, PSOCKADDR pAddr)
 {
-	const char *format = "%15[0-9.]:%5[0-9]";
-	char ip[16] = { 0 };  // ip4 addresses have max len 15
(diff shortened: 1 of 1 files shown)


airween:
Hi @A13501350,

thanks for this PR.

Unfortunately SonarCloud found some errors. I reviewed them and seems like those are not critical and easy to fix them. Could you take a look at them? See the SonarCloud check line or [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3443/checks?check_run_id=51126542620) link.

A13501350:
Yeah. The fixes themselves are straightforward, but implementing them would introduce changes that differ from the existing code in terms of features and style.

airween (approved):

airween:
Thanks @A13501350, I merged this PR.

Do you think this fixes #1988?

PR #3447, opened 2025-10-15
correct handling result of CPTRetriveNode

chenuduss:
## what

correct handling result of CPTRetriveNode(). 

## why

Function CPTRetriveNode may return NULL, and null check required

### DIFF
--- a/src/utils/msc_tree.cc
+++ b/src/utils/msc_tree.cc
@@ -668,14 +668,21 @@ TreeNode *CPTFindElementIPNetblock(unsigned char *ipdata, unsigned char ip_bitma
 
         node = CPTRetriveNode(ipdata, ip_bitmask, node);
 
-        if (node && node->bit != ip_bitmask)    {
+        if(node == NULL)    {
+            //if (msr && msr->txcfg->debuglog_level >= 9) {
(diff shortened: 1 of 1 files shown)


airween:
Hi @chenuduss,

thanks for this PR, nice catch.

Just out of curiosity: have you faced any problem in practice with this error?

Or how did you catch it?

Btw: the PR looks good to me so after the discussion I'm going to merge this. And thanks again!

Note: in v2 this was fixed last year, see [this](https://github.com/owasp-modsecurity/ModSecurity/commit/91da5872c1173f6cf42d888be6819f3b180305d2#diff-90ec48a7f1bb224be4bd3c69afbdfb602a61a6bf38dacdf5e7e6e9ff961691daL659) PR. But as I see v2 contained already this explicit `node` check. This is interesting, because many functions were just copied from v2 into v3, and it seems like the author left the `node` check part.

chenuduss:
> Just out of curiosity: have you faced any problem in practice with this error?
> Or how did you catch it?

I found this using a static analyzer.

airween (approved):

airween:
Thanks again @chenuduss!

PR #3452, opened 2025-10-22
Add CMake and CI Pipeline for ModSecurityIIS in ModSecurity V2

A13501350:
## What
- Added CMake build scripts for IIS components
- Implemented Windows CI pipeline (GitHub Actions) using CMake builds
- Used vcpkg for dependencies
- Modified WiX Toolset scripts for MSI package installer

## Why
- Enables building newer versions of ModSecurity IIS via CI

**Note:** 
- These changes still require additional testing verification. 
- ssdeep support exists in the original Makefile but is not enabled by default, as it requires msys2 environment and runtime dependencies


airween:
Hi @A13501350,

first of all, many thanks for this PR.

> **Note:**
> * These changes still require additional testing verification.

What kind of testing you think about? Do you want to add them before the merge (I mean are you still working on it?)?

> * ssdeep support exists in the original Makefile but is not enabled by default, as it requires msys2 environment and runtime dependencies

A readme would be grateful, I would add that into our Wiki that describe how to make module for IIS on Windows. That could contain these dependencies too.

What do you think?

.github/workflows/test-ci-windows.yml
```diff
@@ -0,0 +1,333 @@
+name: CI/CD for IIS Module
+
+on:
+  push:
+    branches:
+      - v2/test-ci-windows
```
airween: I'm not sure this branch is a good choice. You should use `v2/master` or keep this empty as we use this on Linux [CI](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/.github/workflows/ci.yml#L3-L5).

.github/workflows/test-ci-windows.yml
```diff
@@ -0,0 +1,333 @@
...
+  pull_request:
+    branches:
+      - v2/test-ci-windows
```
airween: Same as above.

.github/workflows/test-ci-windows.yml
```diff
@@ -0,0 +1,333 @@
...
+        $goBinPath = ""
+        if ($env:GOBIN) {
+            $goBinPath = $env:GOBIN
+        } elseif ($env:GOPATH) {
+            $goBinPath = Join-Path $env:GOPATH "bin"
+        } else {
+            $goBinPath = Join-Path $env:USERPROFILE "go\bin"
+        }
+
+        & "$goBinPath\go-ftw.exe" run -d $testRuleDir --cloud -e "920100-2$|920100-4$|920100-8$|920100-12$|920272-5$|920290-1$|920620-1$|920380-1$" --show-failures-only
```
airween, line 324: This is awesome (I mean you added CRS as a test), but why does it need to exclude these rules?

iis/installer.wxs
```diff
@@ -7,7 +7,7 @@
     lightArgs: 
 -->
 <Wix xmlns="http://schemas.microsoft.com/wix/2006/wi" xmlns:util="http://schemas.microsoft.com/wix/UtilExtension">
-    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.12" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
+    <Product Id="22B04FDB-9BAB-46B0-87B8-A39544ECECD3" Name="ModSecurity IIS" Language="1033" Version="2.9.13" Manufacturer="ModSecurity" UpgradeCode="82F09489-1678-4C38-ADCB-08C3757653DB">
```
airween: Why have you changed the version number to `2.9.13`? The current version is `2.9.12`.

(We have a release procedure documentation, there is this important step (increase the version here too)).

airween (changes requested):

airween:
And one more thing: could you pick up your last PR (#3443) into this PR? I think a rebase from v2/master would be enough. Just see it works as we expect. Thanks!

.github/workflows/test-ci-windows.yml
A13501350: I'm not deeply familiar with IIS. My understanding is that IIS preprocesses and corrects certain malformed requests earlier in its pipeline, likely before the `OnBeginRequest` stage. This normalization causes the request headers to appear valid by the time the rules are evaluated, so the rules won't fire.

airween: So you've faced false positive hits with these rules?

iis/installer.wxs
A13501350: Sorry. I was testing the update feature of the MSI package (which did not seem to be enabled). I will revert it.

airween:
@A13501350, please take a look at the SonarCloud issues:
https://sonarcloud.io/project/security_hotspots?id=owasp-modsecurity_ModSecurity&pullRequest=3452&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true

A13501350:
```log
Run msys2/setup-msys2@847bbebdb774c27cd90b723718456dd14637b641
  with:
    msystem: UCRT64
    update: true
    install: git make autoconf automake libtool mingw-w64-ucrt-x86_64-gcc mingw-w64-ucrt-x86_64-pkg-config
  
    path-type: minimal
    pacboy: false
    release: true
    location: RUNNER_TEMP
    platform-check-severity: fatal
    cache: true
  env:
    APACHE_ROOT: C:\tools\Apache24
Error: File not found: 'D:\a\_actions\msys2\setup-msys2\847bbebdb774c27cd90b723718456dd14637b641\index.js'
```

I tried using the full commit hash msys2/setup-msys2@847bbebdb774c27cd90b723718456dd14637b641, but it didn't work. Also, since this dependency updates frequently, I'm not sure if pinning to a specific commit is the best approach.

.github/workflows/test-ci-windows.yml
A13501350: yeah i think so

airween:
> I tried using the full commit hash [msys2/setup-msys2@847bbeb](https://github.com/msys2/setup-msys2/commit/847bbebdb774c27cd90b723718456dd14637b641), but it didn't work. Also, since this dependency updates frequently, I'm not sure if pinning to a specific commit is the best approach.

No worries, at this time it would be enough to add a stable IIS build on Windows. Just make those lines as comment. Probably the Sonar issues will disappear too.

A13501350:
I noticed ssdeep was disabled in commit b02256cf1e1858321531a63081ae3343f1e99ec8, but after testing, ssdeep appears to compile and link normally with no apparent functionality impact, so I've re-enabled it.

A13501350:
Though it does require msys2 for compilation

airween:
> Though it does require msys2 for compilation

I see. Then let's see how can you solve that (it would be nice to disappear the Sonar issue).

A13501350:
Done. However, there are other actions using tags instead of SHA hashes, and I believe the best practice is to use a bot like [Dependabot](https://docs.github.com/code-security/getting-started/dependabot-quickstart-guide) to check for updates.

airween:
Thanks! I think that's excellent now!

> I believe the best practice is to use a bot like [Dependabot](https://docs.github.com/code-security/getting-started/dependabot-quickstart-guide) to check for updates.

yes, you're right, we should involve that.

airween (approved):

Issue #3457, opened 2025-11-01
ModSecurity v2.9.12: find_yajl.m4 fails to detect YAJL installed from source


weida:
**Describe the bug**

When building branch v2/master with the latest yajl library 2.1.0 compiled from source, the build fails to detect yajl when provided with the correct path.

**Logs and dumps**

yajl was compiled from source under /usr/local/yajl-2.1.0, Compiling Modsecurity with `--with-yajl=/usr/local/yajl-2.1.0`s, reporting that the YAJL library is not found.

```bash
checking for libyajl config script... no
checking for yajl install... no
configure: optional yajl library not found
````

**To Reproduce**

1. Install yajl 2.1.0 from source 
```bash
wget https://github.com/lloyd/yajl/archive/refs/tags/2.1.0.tar.gz -O yajl-2.1.0.tar.gz
tar zxvf yajl-2.1.0.tar.gz
cd yajl-2.1.0
./configure -p /usr/local/yajl-2.1.0
make 
sudo make install 
sudo chmod -R 755 /usr/local/yajl-2.1.0
```
2. Run configure with  modsecurity-v2.9.12
```bash
cd modsecurity-v2.9.12
./autogen.sh
./configure --with-yajl=/usr/local/yajl-2.1.0 --prefix=/usr/local/modsecurity-v2.9.12
```
3. Check the outcome of YAJL detection::
```bash
checking for libyajl config script... no
checking for yajl install... no
configure: optional yajl library not found
```
4.The yajl headers and libraries exist：
```bash
/usr/local/yajl-2.1.0/include/yajl:
total 36
-rwxr-xr-x 1 root root  363 Nov  1 16:02 yajl_version.h
-rwxr-xr-x 1 root root 7171 Mar 19  2014 yajl_tree.h
-rwxr-xr-x 1 root root 9821 Mar 19  2014 yajl_parse.h
-rwxr-xr-x 1 root root 7060 Mar 19  2014 yajl_gen.h
-rwxr-xr-x 1 root root 2572 Mar 19  2014 yajl_common.h
...
/usr/local/yajl-2.1.0/lib:
total 104
-rwxr-xr-x 1 root root 40688 Nov  1 16:03 libyajl.so.2.1.0
lrwxrwxrwx 1 root root    16 Nov  1 15:13 libyajl.so.2 -> libyajl.so.2.1.0
[...]

PR #3458, opened 2025-11-01
fix: yajl detection for source installations (#3457)

weida:
## what

- Modified `find_yajl.m4` to correctly detect yajl when installed from source.  
- Added a check for the `lib` subdirectory in addition to the top-level path.  
- Ensures that ModSecurity compilation succeeds when `--with-yajl` is specified. 

## why

- The build failed to detect yajl 2.1.0 compiled from source, even when the correct path was provided.  

## references

- Issue: #3457

### DIFF
--- a/build/find_yajl.m4
+++ b/build/find_yajl.m4
@@ -89,10 +89,14 @@ else
                 yajl_lib_path="${x}/"
                 yajl_lib_name="yajl"
                 break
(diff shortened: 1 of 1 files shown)


airween (approved):

airween:
Thanks for your patch, @weida!

PR #3464, opened 2025-11-13
fix: encode possible binary characters from user input in audit logs

### LINKED ISSUE #3463: ModSecurity v3.0.12 audit logs contain unescaped binary data making logs invisible to kubectl and unparseable by log processors
**Describe the bug**
ModSecurity v3.0.12 audit logs contain raw binary/hex-encoded data that completely breaks JSON validation. This causes:
1. **kubectl logs**: Problematic logs become INVISIBLE - they don't appear in output at all
2. **Log processors**: CrowdSec, Vector, Fluent-bit fail to parse with JSON errors
3. **Security monitoring**: Critical security events are completely lost from visibility

**Logs and dumps**

Example of corrupted log that disappears from kubectl output:
```json
{
  "transaction": {
    "messages": [{
      "details": {
        "match": "Matched against variable `ARGS_NAMES:�d allow_url_include=1 �d auto_prepend_file=php://input'",
        "data": "Matched Data: found within ARGS_NAMES:�d allow_url_include=1 �d auto_prepend_file=php://input",
        "ruleId": "933100",
        "severity": "2"
      }
    }]
  }
}
```

**To Reproduce**

1. **Setup ModSecurity with OWASP CRS 4.4.0:**
```apache
SecAuditLogParts ABHJZ
SecAuditLogFormat JSON  
SecAuditLog /dev/stdout

SecRule REQUEST_URI ".*" \
    "phase:1,pass,nolog,ctl:ruleRemoveTargetById=1-9999999;ARGS,ctl:ruleRemoveTargetById=1-9999999;ARGS_NAMES"
```

2. **Send attack with binary data:**
```bash
curl "http://target/?%ADd+allow_url_include%3d1+%ADd+auto_prepend_file%3dphp://input"
```

3. **Check kubectl logs:**
```bash
kubectl logs <modsecurity-pod> | grep "933100"
```
**Result:** No output - the log entry is completely missing

4. **Check raw logs directly:**
```bash
kubectl logs <modsecurity-pod> --previous > raw.log
```
**Result:** File contains corrupted JSON with binary `�` characters

**Expected behavior**
- All audit logs should be visible via `kubectl logs`
- JSON should be valid and parseable by all standard tools
- Security events should not be lost

**Actual behavior**
- **kubectl logs**: Problematic logs DISAPPEAR from output entirely
- **CrowdSec**: Crashes parsing and stops processing all logs
- **Vector**: Drops entries with JSON parse errors  
- **Fluent-bit**: Fails to process and forward
- **Security events**: Completely lost from monitoring

**Server**
- ModSecurity: v3.0.12 + nginx-connector v1.0.3
- WebServer: nginx/1.25.5  
- OS: Kubernetes (ingress-nginx)

**Rule Set**
- OWASP CRS 4.4.0

**Additional context**

**CRITICAL DATA LOSS:**
This is worse than just corrupted logs - we're losing security events completely:

1. **kubectl logs**: Shows only "good" logs, hides all ModSecurity alerts with binary data
2. **No debugging**: Cannot see what attacks are being blocked
3. **Security blindspot**: Attack detection events vanish from all monitoring
4. **Silent failure**: No indication that logs are being lost

**Evidence of data loss:**
- Normal application logs appear in `kubectl logs`
- ModSecurity logs with binary data are COMPLETELY MISSING
- Only by examining raw container logs do we see the corrupted entries
- Security teams have no visibility into attacks being blocked

**Impact:**
- ❌ **Zero visibility** into ModSecurity blocks via standard kubectl
- ❌ **Security monitoring** completely broken - attacks are invisible
- ❌ **Compliance failure** - cannot audit security events
- ❌ **Production debugging** impossible for WAF issues

This makes ModSecurity unusable for production security - we're blocking attacks but can't see them happening.

airween:
## what

This PR encodes the possible binary characters from user input in audit.log.

## why

Actually if the user sends a request with invalid encoding then it appears in audit.log in binary format. The problem is that not all parser can handle those binary characters.

For more details see #3463.

## references

A similar issue was fixed in #2854.

Closes: #3463.

### DIFF
--- a/test/test-cases/regression/auditlog.json
+++ b/test/test-cases/regression/auditlog.json
@@ -541,7 +541,7 @@
     "enabled": 1,
     "version_min": 300000,
     "version_max": 0,
-    "title": "auditlog : Binary char from input",
+    "title": "auditlog : Binary char from input, check message",
     "client": {
       "ip": "200.249.12.31",
       "port": 2313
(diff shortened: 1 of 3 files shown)


test/test-cases/regression/auditlog.json
```diff
@@ -536,5 +536,64 @@
       "SecAuditLogType Serial",
       "SecAuditLogRelevantStatus \"^(?:5|4(?!04))\""
     ]
+  },
```
theseion, line 539: Add tests for body and headers as well.

theseion (changes requested):

test/test-cases/regression/auditlog.json
airween: Body content is JSON log is the same, but sure, I'll add that.

Headers are not encoded, actually Nginx (and probably most other servers) doesn't allow binary header. Eg. in Nginx log there is an entry:
```
...client sent invalid header line: "X-%ADd+custom: Some %AD value" while reading client request headers...
```

But probably we should add that too for other servers, which allows those headers.

airween: Done, see [102275f](https://github.com/owasp-modsecurity/ModSecurity/pull/3464/commits/102275fd01007ff18ca570a0f4de80d82e0a6435).

theseion: I asked for a test for the headers because you've added encoding there too: https://github.com/owasp-modsecurity/ModSecurity/pull/3464/files#diff-974db259ad96aebcb69f17e473d6dcde871731ca153572295e15b932073b3af4R1625

theseion (approved):

test/test-cases/regression/auditlog.json
airween: Yes, but I added it in my last [commit](https://github.com/owasp-modsecurity/ModSecurity/pull/3464/commits/102275fd01007ff18ca570a0f4de80d82e0a6435#diff-974db259ad96aebcb69f17e473d6dcde871731ca153572295e15b932073b3af4R1625) :). It was not added originally.

Never mind, I think now it's more clear - thanks.

PR #3467, opened 2025-11-20
Minor edit with remove useless checks

chenuduss:
Small PR with remove useless checks

### DIFF
--- a/src/request_body_processor/multipart.cc
+++ b/src/request_body_processor/multipart.cc
@@ -580,7 +580,7 @@ int Multipart::process_part_data(std::string *error, size_t offset) {
                 m_mpp->m_tmp_file->Open();
 
                 /* do we have an opened file? */
-                if (!m_mpp->m_tmp_file || m_mpp->m_tmp_file->getFd() < 0) {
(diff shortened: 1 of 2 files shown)


airween:
Thanks for this patch.

It seems good to me, I just realized that in the `Transaction` object there is no check for the `nullptr` value of the `Rule` argument, nothing at all. Probably we should add that into the [constructor](https://github.com/owasp-modsecurity/ModSecurity/blob/09fec93cb7621175a1631b630f482a87dca1e0b8/src/transaction.cc#L154-L157)...

airween (approved):

chenuduss:
Any attempts to create Transaction with null pointer of rules (or invalid pointer) now impossible because segfault will be caused immidiately on line:

`m_secXMLParseXmlIntoArgs(rules->m_secXMLParseXmlIntoArgs),`

Therefore, in the current implementation, if the object is created and we are not in the constructor, then we can be sure that m_rules is valid. :-) This means that all existing library clients are guaranteed to ensure that the rules pointer is valid. :-)

PR #3468, opened 2025-11-20
Performance tweaks

edding3000:
## what

Range based for loops with references are already used in this project, but in a few places not.
I am not sure why. I changed this code locations and executed unit tests.
Also use '*' for pointers when using auto keyword, makes it more readable.

## why

Improve performance a little bit.

## questions

"const auto&" could also be used in many places. Is there a reason why it is not used?


src/engine/lua.cc
```diff
@@ -297,7 +297,7 @@ int Lua::getvars(lua_State *L) {
     variables::Variable::stringMatchResolveMulti(t, varname, &l);
 
     lua_newtable(L);
-    for (auto i : l) {
+    for (auto* i : l) {
```
airween:
```suggestion
    for (const auto* i : l) {
```
The [original](https://github.com/owasp-modsecurity/ModSecurity/commit/a676f313c3f9277ae007b09d79fa698f61947ce6) code was written about 8 years ago, the author is no longer involved in development.

If I had to guess why they used value copy mechanism, I would say they wanted to avoid any chance to modify the value on Lua stack (except for intentional change with `setvar()`).

If we want to keep this restriction, please consider to use `const` modifier here.

src/rules_exceptions.cc
```diff
@@ -195,7 +195,7 @@ bool RulesExceptions::contains(int a) {
         }
     }
 
-    for (auto z : m_ranges) {
+    for (auto& z : m_ranges) {
```
airween:
```suggestion
    for (const auto& z : m_ranges) {
```
Please see `cppcheck`'s [warning](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/19540276744/job/55952868889?pr=3468#step:5:3289):
```
warning: src/rules_exceptions.cc,198,style,constVariableReference,Variable 'z' can be declared as reference to const
```

src/rules_exceptions.cc
```diff
@@ -212,7 +212,7 @@ bool RulesExceptions::merge(RulesExceptions *from) {
             return ret;
         }
     }
-    for (auto b : from->m_ranges) {
+    for (auto& b : from->m_ranges) {
```
airween:
```suggestion
    for (const auto& b : from->m_ranges) {
```
Please see cppcheck's [warning](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/19540276744/job/55952868889?pr=3468#step:5:3290):
```
warning: src/rules_exceptions.cc,215,style,constVariableReference,Variable 'b' can be declared as reference to const
```

src/transaction.cc
```diff
@@ -1700,7 +1700,7 @@ std::string Transaction::toJSON(int parts) {
             reinterpret_cast<const unsigned char*>("messages"),
             strlen("messages"));
         yajl_gen_array_open(g);
-        for (auto a : m_rulesMessages) {
+        for (auto& a : m_rulesMessages) {
```
airween:
```suggestion
        for (const auto& a : m_rulesMessages) {
```
Probably here you could use `const` too.

src/transaction.cc
```diff
@@ -1721,7 +1721,7 @@ std::string Transaction::toJSON(int parts) {
                 reinterpret_cast<const unsigned char*>("tags"),
                 strlen("tags"));
             yajl_gen_array_open(g);
-            for (auto b : a.m_tags) {
+            for (auto& b : a.m_tags) {
```
airween:
```suggestion
            for (const auto& b : a.m_tags) {
```
Probably here you could use `const` too.

airween (changes requested):

src/rules_exceptions.cc
airween: Sonarcloud also reports [this](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&pullRequest=3468&issueStatuses=OPEN,CONFIRMED&sinceLeakPeriod=true).

edding3000:
I was not sure if there was any restriction not to use const, thats why i asked. I could extend the PR with more positions for "const auto&" if you like.

airween:
> I was not sure if there was any restriction not to use const, thats why i asked. I could extend the PR with more positions for "const auto&" if you like.

Sure, go! If the tests will be passed, I'll be happy :).

Btw you could review the tests too, may be you will find some lack there too (for eg. you're checking the code...)

airween:
Hi @edding3000,

sorry for the long delay, it seems like this PR is ready to merge. Just one question: could you merge the current v3/master into this one?

airween:
@edding3000,

also please take a look at the few request changes.

src/rules_exceptions.cc
edding3000: done

src/rules_exceptions.cc
edding3000: done

src/transaction.cc
edding3000: done

src/transaction.cc
edding3000: done

edding3000:
@airween I merged v3/master. cppcheck fails now. Can you see the reason?

airween:
> @airween I merged v3/master. cppcheck fails now. Can you see the reason?

Yes:
```
2026-02-17T18:48:51.1404560Z warning: src/run_time_string.cc,54,style,functionConst,The member function 'modsecurity::RunTimeString::evaluate' can be made a const function. Making this function 'const' should not cause compiler errors. Even though the function can be made const function technically it may not make sense conceptually. Think about your design and the task of the function first - is it a function that must not change object internal state?
```

I think `cppcheck` realized that you [changed](https://github.com/owasp-modsecurity/ModSecurity/pull/3468/changes#diff-2bd76cacd9e1f795d3b62a65bd3d129d6c3672c38ec902b732fea39f8ad4b551R56) the loop's variable to `const`, and figured out the function can't change the object (I don't see the relationship what this `const` triggered this error), so it expects to make the fn to `const` - but warns us that it might be wrong.

So what you can try now:
* change the function signature to `const`: `std::string RunTimeString::evaluate(Transaction *t, Rule *r) const {`
* if the build will be failed, rollback that change and add an inline suppression to that line: `std::string RunTimeString::evaluate(Transaction *t, Rule *r) { // cppcheck-suppress functionConst`
* in this case, please add this note to the PR's description

Do you have `cppcheck` on your local dev environment? It's important to use the latest release, and on Linux only Debian SID contains that version. [Here](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/.github/workflows/ci.yml#L141-L176) you can see how do we use it in Docker.

edding3000:
Thx for your hint. I build cppcheck from source, but it did not find this - perhaps I did not use correct settings. Anyway, warning is fixed now.

headers/modsecurity/rules_exceptions.h
```diff
@@ -51,7 +51,7 @@ class RulesExceptions {
     bool addRange(int a, int b);
     bool addNumber(int a);
     bool contains(int a);
```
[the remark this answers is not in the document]
airween: @edding3000 I think this suggestion makes sense.

src/run_time_string.h
```diff
@@ -48,9 +48,9 @@ class RunTimeString {
...
     void appendText(const std::string &text);
     void appendVar(std::unique_ptr<modsecurity::variables::Variable> var);
-    std::string evaluate(Transaction *t);
-    std::string evaluate(Transaction *t, Rule *r);
-    std::string evaluate() {
+    std::string evaluate(Transaction *t) const;
+    std::string evaluate(Transaction *t, Rule *r) const;
+    std::string evaluate() const{
         return evaluate(NULL);
     }
```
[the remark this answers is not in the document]
airween: This one also makes sense.

headers/modsecurity/rule_with_actions.h
```diff
@@ -80,7 +80,7 @@ class RuleWithActions : public Rule {
 
     std::vector<actions::Action *> getActionsByName(const std::string& name,
         const Transaction *t);
```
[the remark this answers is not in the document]
airween: This suggestion makes sense.

airween (changes requested):
@edding3000 thanks, please take a look at Copilot reviews and my suggestions below.

headers/modsecurity/rule_with_actions.h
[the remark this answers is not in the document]
airween: Just for sure I took a second review.

Because of this header file is part of the public API, and `containsTag` appears between the symbols, this can cause to an unresolved symbol error.
```bash
$ nm -D /path/to/libmodsecurity.so.3 | grep containsTag
0000000000276cc6 T _ZN11modsecurity15RuleWithActions11containsTagERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEEPNS_11TransactionE
```

headers/modsecurity/rules_exceptions.h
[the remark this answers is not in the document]
airween: Same as above, this is public also.

airween:
@edding3000,

one more thing: if you will align those suggested modifications (make wrappers to functions which changed their signatures), could you take a look at [this](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/variables/variable.h#L157) too?
```C++

    bool toOmit(std::string a) {
        for (auto &z : *this) {
            if (z->match(a)) {
                return true;
            }
        }
        return false;
    }
```
This can be a `const` argument too, it seems like the newest `cppcheck` [warns](https://github.com/owasp-modsecurity/ModSecurity/pull/3504#issuecomment-4019424800) about that.

airween:
@edding3000 could you take a look at this PR and align the necessary changes? (Eg. update the branch to current v3/master, check comments, ...)

Issue #2848, opened 2022-12-21
Memory leaks with nginx reload (without restart)


martinhsv:
When, instead of restarting nginx, one performs a ```reload``` of the configuration, memory may leak.

The memory leaks are not large per reload, but if doing so frequently, the free memory reduction will become noticeable. A near-term mitigation is to at least periodically do a true restart.

This issue is being created in lieu of #2502 and others.

Volatus:
@martinhsv Would it be possible to do a small patch release soon that would include the commit e9a7ba4a60be48f761e0328c6dfcc668d70e35a0 ? According to https://github.com/kubernetes/ingress-nginx/issues/8166#issuecomment-1311495548 it did alleviate the leaks that were significant in size and would be very helpful for the users for the time being as other leaks related to reloading get ironed out. Thanks.

IanRobertson-wpe:
Hi @martinhsv, I believe my organization is encountering this issue as well. What is needed in order to address this? Is the root cause already identified?

martinhsv:
Hello @Volatus ,

The ModSecurity project has not had a practice of doing patch releases for single issues that are separate from accumulated changes in v3/master. When the next 3.0.x release appears, it will almost certainly be from v3/master as that branch appears at that time.

The next release of 3.0.x is tentatively planned for the moderately soon period (by which I mean within 8-10 weeks) -- if that is within your definition of 'soon'. If that is problematically long for you, you could always build ModSecurity yourself from a point in time that includes the commit that you have referenced.

I do, understand, of course, that some installations may have a policy of only allowing use of official releases. If that is the case for you, and the aforementioned timeframe is problematic, I could take that into consideration. But that would have to be weighed against some other factors.

Hi @IanRobertson-wpe ,

Perhaps my response to Volatus answers most of what you are asking about as well.

One other thing I'll mention is that there is one (smaller) unresolved leak present with common ModSecurity usage. I have a fix for this for which I'll likely create the PR within a few days.

Volatus:
> Hello @Volatus ,
> 
> The ModSecurity project has not had a practice of doing patch releases for single issues that are separate from accumulated changes in v3/master. When the next 3.0.x release appears, it will almost certainly be from v3/master as that branch appears at that time.
> 
[...]

PR #3470, opened 2025-11-21
Fix memory leak

FireBurn:
This is a potential fix to https://github.com/owasp-modsecurity/ModSecurity/issues/2848

It was AI generated by Google AI Studio

Here are it's workings:

Based on the Valgrind output and the provided code, the memory leak is located in the **`TransactionAnchoredVariables`** class (which is part of the `Transaction` object).

The stack trace indicates that memory allocated during the construction of `AnchoredSetVariableTranslationProxy` (specifically inside a `std::function` initialization) is never freed.

```text
==27426== 1 bytes in 1 blocks are possibly lost ...
==27426== at 0x482BBB5: operator new(unsigned int) ...
==27426== by 0x94EB43A: ... function<...>::operator= ...
==27426== by 0x94EB43A: AnchoredSetVariableTranslationProxy ...
==27426== by 0x94EB43A: modsecurity::TransactionAnchoredVariables::TransactionAnchoredVariables(modsecurity::Transaction*) ...
```

In `src/transaction.cc`, the `Transaction` constructor initializes `TransactionAnchoredVariables`:

```cpp
Transaction::Transaction(ModSecurity *ms, RulesSet *rules, const char *id,
    void *logCbData, const time_t timestamp)
    : ...
    TransactionAnchoredVariables(this) { // <--- Initialization here
    ...
```

And in `src/variables/variable.h`, the code accesses `m_variableArgsNames` as if it were an object (using `&`), not a pointer:

```cpp
anchoredSetVariableTranslationProxy = &t->m_variableArgsNames;
```

This usage pattern combined with the leak indicates that **`TransactionAnchoredVariables` declares its members (like `m_variableArgsNames`) as references (`&`) but initializes them using `new` in its constructor.**

Because they are references, the `TransactionAnchoredVariables` destructor does not (and cannot) delete the objects they refer to. Consequently, the `AnchoredSetVariableTranslationProxy` objects (and their internal `std::function` members) are leaked when the `Transaction` is destroyed.

**To fix this:**
The members in `TransactionAnchoredVariables` should be changed from references (e.g., `AnchoredSetVariableTranslationProxy &`) to pointers (e.g., `std::unique_ptr<AnchoredSetVariableTranslationProxy>`), or they should be deleted manually in the `TransactionAnchoredVariables` destructor if raw pointers are used. Given the `&t->member` usage, they were likely intended to be value members or references binding to dynamic memory (which is the bug).

### DIFF
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -205,9 +205,12 @@ class TransactionAnchoredVariables {
         m_variableFilesTmpNames(t, "FILES_TMPNAMES"),
         m_variableMultipartPartHeaders(t, "MULTIPART_PART_HEADERS"),
         m_variableOffset(0),
-        m_variableArgsNames("ARGS_NAMES", &m_variableArgs),
-        m_variableArgsGetNames("ARGS_GET_NAMES", &m_variableArgsGet),
-        m_variableArgsPostNames("ARGS_POST_NAMES", &m_variableArgsPost)
+        m_pVariableArgsNames(std::make_unique<AnchoredSetVariableTranslationProxy>("ARGS_NAMES", &m_variableArgs)),
+        m_variableArgsNames(*m_pVariableArgsNames),
+        m_pVariableArgsGetNames(std::make_unique<AnchoredSetVariableTranslationProxy>("ARGS_GET_NAMES", &m_variableArgsGet)),
+        m_variableArgsGetNames(*m_pVariableArgsGetNames),
+        m_pVariableArgsPostNames(std::make_unique<AnchoredSetVariableTranslationProxy>("ARGS_POST_NAMES", &m_variableArgsPost)),
+        m_variableArgsPostNames(*m_pVariableArgsPostNames)
         { }
 
     AnchoredSetVariable m_variableRequestHeadersNames;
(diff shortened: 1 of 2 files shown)


airween:
@FireBurn,

first of all, thank you for this PR.

It was a bit suspicious, for eg. I don't understand the idea behind the new constructor's variable creating (you pass a non-existent reference to the new pointer type variable), but before I continue the understanding of the code, please could you show me a rule (a SecRule) or a configuration snippet and a request where I can check this memleak (with Valgrind).

Thank you!

airween:
@FireBurn,

> before I continue the understanding of the code, please could you show me a rule (a SecRule) or a configuration snippet and a request where I can check this memleak (with Valgrind).

It's still not clear how can we reproduce the issue, but could you take a look at #3500? Probably that really fixes the issue.

Thank you!

FireBurn:
Hey, it wasn't me that reported the memory leak, it was reported in https://github.com/owasp-modsecurity/ModSecurity/issues/2848 I use modsecurity with haproxy not nginx, this was just a driveby commit

I've asked the original reporter to confirm if it fixed things for them

airween:
Any news?

PR #3473, opened 2025-12-09
Collection: Include `<cstdint>`.

Gacko:
This fixes builds on Alpine v3.23.0 (GCC v15.2.0, musl v1.2.5).

```
headers/modsecurity/collection/collection.h:x:x: error: 'int32_t' has not been declared
headers/modsecurity/collection/collection.h:x:x: note: 'int32_t' is defined in header '<cstdint>'; this is probably fixable by adding '#include <cstdint>'
```

### DIFF
--- a/headers/modsecurity/collection/collection.h
+++ b/headers/modsecurity/collection/collection.h
@@ -21,6 +21,7 @@
 #include <vector>
 #include <algorithm>
 #include <memory>
+#include <cstdint>
 #endif
 
 


airween:
Just a side note: the referenced type (`int32_t`) is used [here](https://github.com/owasp-modsecurity/ModSecurity/blob/dc6cce5f0cc99d0922ad740971e6ef453e115f5a/headers/modsecurity/collection/collection.h#L57).

And one question: is there no other file where it is missing?

Gacko:
> And one question: is there no other file where it is missing?

I haven't checked, but we're relying on it in Ingress NGINX and this change made the build of our NGINX base image work: https://github.com/kubernetes/ingress-nginx/blob/main/images/nginx/rootfs/build.sh#L337-L367.

airween (approved):

airween:
@Gacko: thanks for this patch!

PR #3476, opened 2025-12-16
Allow partial processing of XML, JSON, and multipart request body

hnakamur:
## what

- This pull request prevents a parse error that occurs when `SecRequestBodyLimitAction` is set to `ProcessPartial` and the length of an XML or JSON request body exceeds the configured limit.

## why

- When `SecRequestBodyLimitAction` is `ProcessPartial`, the expected behavior is to process the partial XML or JSON body up to the defined limit.
- The request should then be rejected if any invalid values are found within that processed part, or passed through if the processed part is valid.
- The previous implementation failed to handle the partial body correctly, leading to a parse error when the limit was exceeded.

## references

- the same modification for v2 is at: https://github.com/owasp-modsecurity/ModSecurity/pull/3483


airween:
Hi @hnakamur!

Thanks for this excellent patch!

Would you also consider adding more tests to the [test/test-cases/regression/config-body_limits.json](https://github.com/owasp-modsecurity/ModSecurity/blob/7267c1dc21883e8572f164c9564cebbf6d593dc6/test/test-cases/regression/config-body_limits.json)?

And I think this test file is not correct, for eg. [this](https://github.com/owasp-modsecurity/ModSecurity/blob/7267c1dc21883e8572f164c9564cebbf6d593dc6/test/test-cases/regression/config-body_limits.json#L253-L309) test uses a `POST` request with a multipart body, but there is no `Content-Type` header, therefore I'm not sure it really does what the author expected. So, taking a big review also would be a huge help here.

Regarding to the Sonarcloud [issue](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3476&id=owasp-modsecurity_ModSecurity&open=AZsk26kWrBm5rHtOlNeO): I see that you only added a new argument to that function, and used the existing `new` operator. I never expect devs to fix another developer's Sonar problems, so you can leave it as is, but if you think about it, you can change the solution  [proposed](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3476&id=owasp-modsecurity_ModSecurity&open=AZsk26kWrBm5rHtOlNeO&tab=why)  there.

dune73:
This is interesting @hnakamur. Could you explain a bit more how you do this?

I mean we are technically facing a body that has been cut short, are not we? Thus a broken partial body. How can you handle this or am I getting this wrong?

hnakamur:
Hello @airween, thanks for the review!

I've added additional tests as requested.

To keep this PR from becoming too large, I'll handle the existing multipart test cases in a separate PR.

Also, I've updated the code to use `std::unique_ptr` instead of raw `new` calls.

hnakamur:
@dune73 You're correct that the body is technically incomplete. However, when `SecRequestBodyLimitAction` is set to `ProcessPartial`, the intended behavior is for ModSecurity to parse and inspect the data it has received up to the limit, while simply ignoring the remainder.

I have added some tests to demonstrate exactly how this partial body is handled. You can see the examples here: https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/6ba8eac908a5b3b8d39296808555cf1f39d2ff4b

dune73:
Thank you very much. This is really interesting. Did not think this could be implemented.

airween:
> I've added additional tests as requested.

thanks - have you seen the results? Almost all of tests are failed now.
 
> To keep this PR from becoming too large, I'll handle the existing multipart test cases in a separate PR.

I'll check that later, when all tests will be passed. Could you check them please? [Here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/20301063138/job/58306519016?pr=3476#step:9:66) is one,
 
> Also, I've updated the code to use `std::unique_ptr` instead of raw `new` calls.

because there is no `new()` operator, you can't use `delete()` either.

airween:
> @dune73 You're correct that the body is technically incomplete. However, when `SecRequestBodyLimitAction` is set to `ProcessPartial`, the intended behavior is for ModSecurity to parse and inspect the data it has received up to the limit, while simply ignoring the remainder.
> 
> I have added some tests to demonstrate exactly how this partial body is handled. You can see the examples here: [6ba8eac](https://github.com/owasp-modsecurity/ModSecurity/commit/6ba8eac908a5b3b8d39296808555cf1f39d2ff4b)

Could you add some more tests please?

I'm afraid this PR modifies other thing too than the mentioned one.

Consider the request:
```json
{"foo":1, "bar":[0,1,2,3]]}
```
which is not well-formed (because of the double close square bracket).
With a config like this:
```
SecRuleEngine On
SecRequestBodyAccess On
SecRequestBodyLimitAction ProcessPartial
SecRequestBodyLimit 1024
```
where the limit is bigger than the payload size, but the `LimitAction` is `ProcessPartial`. I'm afraid with this modification we will allow the non well-formed payloads which fit the `BodyLimit`.

I really don't know that, this is why would be good to add more tests (including this scenario too).

And a good advice: if you have a chance, could you test the payloads and configs with Apache too? (There is no regression test framework, that is, there is, but it is difficult to use, so if you're curious, you need to check that by hand.)

hnakamur:
@airween Thanks for your comments.

I have added more tests and updated the implementation to ensure we require well-formed payloads that fit the `BodyLimit`, even when `LimitAction` is set to `ProcessPartial`.

> thanks - have you seen the results? Almost all of tests are failed now.

My apologies. In my local environment, the code compiled successfully and all regression tests passed for my commit.

> because there is no new() operator, you can't use delete() either.

I previously removed the delete() call in [this commit](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/b77a1a77adbaf1db6cfbb1c2ac383a34a97592fe). I reviewed the error you mentioned, but I'm not sure why it's occurring or how to resolve it. I have reverted the code to use new and delete for the time being.

> And a good advice: if you have a chance, could you test the payloads and configs with Apache too? (There is no regression test framework, that is, there is, but it is difficult to use, so if you're curious, you need to check that by hand.)

I will try to run tests with ModSecurity 2.x on Apache as soon as I have the time.

airween:
Hi @hnakamur,

thanks for additional commits.

Two important things:
* `cppcheck` gave some warnings [1], please take a look at them
* SonarCloud also gave some fixable things; As I previously wrote I don't expect you to fix all of them, but it would be not to increase the number of Sonar issues :), so please take a look at the list below [2]

The answer for [my question](https://github.com/owasp-modsecurity/ModSecurity/pull/3476#issuecomment-3665697111) is this, right?

> I have added more tests and updated the implementation to ensure we require well-formed payloads that fit the `BodyLimit`, even when `LimitAction` is set to `ProcessPartial`.


* 1:
```
2025-12-31T14:08:00.8726840Z warning: src/request_body_processor/multipart.cc,107,warning,uninitMemberVar,Member variable 'Multipart::m_allow_partial' is not initialized in the constructor. Member variables of native types, pointers, or references are left uninitialized when the class is instantiated. That may cause bugs or undefined behavior.

2025-12-31T14:13:01.4190980Z warning: test/common/modsecurity_test.h,34,warning,uninitMemberVar,Member variable 'ModSecurityTest::m_show_request_body_size' is not initialized in the constructor. Member variables of native types, pointers, or references are left uninitialized when the class is instantiated. That may cause bugs or undefined behavior.

2025-12-31T14:13:44.5416030Z warning: test/common/modsecurity_test.h,34,warning,uninitMemberVar,Member variable 'ModSecurityTest < modsecurity_test :: RegressionTest >::m_show_request_body_size' is not initialized in the constructor. Member variables of native types, pointers, or references are left uninitialized when the class is instantiated. That may cause bugs or undefined behavior.

2025-12-31T14:13:50.8854720Z warning: test/common/modsecurity_test.h,34,warning,uninitMemberVar,Member variable 'ModSecurityTest < modsecurity_test :: UnitTest >::m_show_request_body_size' is not initialized in the constructor. Member variables of native types, pointers, or references are left uninitialized when the class is instantiated. That may cause bugs or undefined behavior.
```

* 2
  * [Replace redundant type with "auto"](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3476&id=owasp-modsecurity_ModSecurity&open=AZt0s6BXI-fJ87E45-Mu)
  * [Reduce the code nesting level](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3476&id=owasp-modsecurity_ModSecurity&open=AZt0s6BXI-fJ87E45-Mt) - only in line [1176](https://sonarcloud.io/code?id=owasp-modsecurity_ModSecurity&pullRequest=3476&selected=owasp-modsecurity_ModSecurity%3Asrc%2Frequest_body_processor%2Fmultipart.cc&line=1176)
  * [Use the in-class initializer](https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=3476&id=owasp-modsecurity_ModSecurity&open=AZsxbZSdzghPq7Y2qBUD)
  * [Reduce the code nesting level](https://sonarcloud.io/code?id=owasp-modsecurity_ModSecurity&pullRequest=3476&selected=owasp-modsecurity_ModSecurity%3Asrc%2Ftransaction.cc&line=792)
  * [Use the in-class initializer](https://sonarcloud.io/code?id=owasp-modsecurity_ModSecurity&pullRequest=3476&selected=owasp-modsecurity_ModSecurity%3Atest%2Fcommon%2Fmodsecurity_test.h&line=39)
  * [Change thy syntax](https://sonarcloud.io/code?id=owasp-modsecurity_ModSecurity&pullRequest=3476&selected=owasp-modsecurity_ModSecurity%3Atest%2Fregression%2Fregression.cc&line=516)

Hint: if you fix `cppcheck` warnings, probably that will solve some Sonar issues too. But you can run `cppcheck` locally too (`make cppcheck`).

hnakamur:
Hello @airween. Thanks for your comment.

Fixed warning reported by cppcheck.
I also aligned the tests for ProcessPartial with those in v2 so that the tests are equivalent between v2 and v3.

I am also going to add more tests for multipart requests, as mentioned in https://github.com/owasp-modsecurity/ModSecurity/pull/3483#issuecomment-3703576046

hnakamur:
I modified the URL-encoded parser and the multipart parser to align with the changes for v2: https://github.com/owasp-modsecurity/ModSecurity/pull/3483#issuecomment-3712886144

airween:
I'll review this PR soon -  could you rebase your V3 related PR branches with the current v3/master? I fixed some cppcheck issues in #3485.

hnakamur:
I merged v3/master to this branch and pushed https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/d15344f4cb925c869ce2ef33501f33e7795093f4 instead of rebasing since I think it is easier.

airween:
Let's clarify what do we want to achieve in #3483, then we can came back here.

airween:
Hi @hnakamur,

thanks for this cumulative PR - I reviewed the Sonar's issues and accepted most of them. I kept 2 opened, could you review them?

hnakamur:
Hi @airween,

I've fixed the two Sonar issues.

I think this PR now contains basically the same changes as:
[v2: Allow partial processing of XML, JSON, and multipart request body by hnakamur · Pull Request #3483 · owasp-modsecurity/ModSecurity](https://github.com/owasp-modsecurity/ModSecurity/pull/3483).

However, I'll need to adjust the tests in this PR once [v3: Remove deprecated REQBODY_PROCESSOR_ERROR and REQBODY_PROCESSOR_ERROR_MSG by hnakamur · Pull Request #3579 · owasp-modsecurity/ModSecurity](https://github.com/owasp-modsecurity/ModSecurity/pull/3579) is merged.

Issue #3450, opened 2025-10-20
Build fails because of missing `library/base64.c` when using Mbed TLS 4.x — Is support planned?


TheophileDiot:
Hello ModSecurity team,

First of all, thank you for your excellent work on ModSecurity (v3 / libmodsecurity) — it’s a great WAF library.

I attempted to build ModSecurity v3 against the [latest Mbed TLS release](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-4.0.0) (**4.0.0**), but the build fails right at the `./configure` step. The error message is as follows:

```
# Check for Mbed TLS
if ! test -f "${srcdir}/others/mbedtls/library/base64.c"; then
  AC_MSG_ERROR([
    Mbed TLS was not found within ModSecurity source directory.
    …
  ])
fi
```

The root cause appears to be **the removal or relocation of `library/base64.c` (and possibly other files) in Mbed TLS 4.x**, so the check in ModSecurity’s build system which expects that file fails.

According to the Mbed TLS 4.0.0 release notes, there are **“significant API changes that break backward compatibility”**.

---

### Questions

1. **Is support for Mbed TLS 4.x planned** in ModSecurity v3?
2. If yes:

   * Is there a rough timeline, release version, or roadmap entry for that?
   * Are there existing branches or PRs to track for this work?
3. If no or not yet:

   * Would the maintainers accept a community contribution / PR to enable building with Mbed TLS 4.x (including updating the configure script check and necessary code changes)?
   * Could documentation be updated to state clearly which Mbed TLS versions are supported (e.g., “Up to Mbed TLS 3.x only”)?

---

Thank you very much for your time and the continued maintenance of ModSecurity.
Best regards,

airween:
Hi @TheophileDiot,


> I attempted to build ModSecurity v3 against the [latest Mbed TLS release](https://github.com/Mbed-TLS/mbedtls/releases/tag/mbedtls-4.0.0) (**4.0.0**), but the build fails right at the `./configure` step. The error message is as follows:

Did you follow the [instructions](https://github.com/owasp-modsecurity/ModSecurity?tab=readme-ov-file#unix-linux-macos-freebsd-) and ran `git ... --recurse`?

MbedTLS is a submodule of the library.
 
> ```
> # Check for Mbed TLS
> if ! test -f "${srcdir}/others/mbedtls/library/base64.c"; then
>   AC_MSG_ERROR([
[...]

PR #3477, opened 2025-12-20
Add initial mbedTLS v4 support

Easton97-Jens:
what

Updated the project to support mbedTLS v4

Adapted to the new mbedTLS v4 structure, which is split into two separate parts

Integrated mbedTLS v4 as a submodule so it can be used by the main module

why

This pull request is based on the request and discussion in
https://github.com/owasp-modsecurity/ModSecurity/issues/3450

mbedTLS v4 introduces structural changes that require updates to the build and integration process

The goal is to provide an initial, working approach for mbedTLS v4 support

known issues / notes

Since mbedTLS v4 is split into two separate components, it is included as a submodule
→ To ensure the submodule is properly loaded by the main module, the following step is currently required:

git submodule update --init --recursive

The include setup is now working, and I can build the project locally.

Note: I am not a professional programmer, and this pull request represents my best effort
→ I am explicitly relying on feedback and guidance from the maintainers, especially regarding the Makefile and build system structure

references

Discussion / Request: https://github.com/owasp-modsecurity/ModSecurity/issues/3450
Repository: https://github.com/Easton97-Jens/ModSecurity

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,12 +3,13 @@ name: Quality Assurance
 on:
   push:
   pull_request:
-
+  
 jobs:
   build-linux:
     name: Linux (${{ matrix.platform.label }}, ${{ matrix.compiler.label }}, ${{ matrix.configure.label }})
     runs-on: ${{ matrix.os }}
     strategy:
+      fail-fast: false
       matrix:
         os: [ubuntu-22.04]
         platform:
@@ -33,6 +34,37 @@ jobs:
           - platform: {label: "x32"}
             configure: {label: "wo ssdeep"}
     steps:
+      - name: Detect latest Lua dev package
+        id: detect_lua
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          sudo apt-get update -y -qq
+
+          CANDIDATES="$(apt-cache pkgnames | grep -E '^liblua[0-9]+\.[0-9]+-dev$' || true)"
+
+          if [ -z "$CANDIDATES" ]; then
+            echo "No libluaX.Y-dev package found"
+            exit 1
+          fi
+
+          BEST_PKG="$(
+            printf '%s\n' "$CANDIDATES" \
+            | sed -E 's/^liblua([0-9]+\.[0-9]+)-dev$/\1 &/' \
+            | sort -V \
+            | tail -n1 \
+            | awk '{print $2}'
+          )"
+
+          if [ -z "$BEST_PKG" ]; then
+            echo "Failed to determine Lua package"
+            exit 1
+          fi
+
+          echo "lua_pkg=$BEST_PKG" >> "$GITHUB_OUTPUT"
+          echo "Using $BEST_PKG"
+
       - name: Setup Dependencies (common)
         run: |
           sudo dpkg --add-architecture ${{ matrix.platform.arch }}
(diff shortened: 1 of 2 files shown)


build.sh
```diff
@@ -24,12 +24,36 @@ ls -1 \
     *.h | tr "\012" " " >> headers.mk
 cd ../
 
+##############################################################################
+# NEU: psa_crypto_driver_wrappers.h automatisch generieren, Pfad wird gesucht
+##############################################################################
+echo "[+] Suche nach psa_crypto_driver_wrappers.h …"
```
airween: Please use English text in all of the source code.

others/Makefile.am
```diff
@@ -15,19 +15,99 @@ noinst_HEADERS = \
...
+
+MBED_PSA_CORE_SOURCES = \
+        mbedtls/tf-psa-crypto/core/psa_crypto.c \
+        mbedtls/tf-psa-crypto/core/psa_crypto_client.c \
+        mbedtls/tf-psa-crypto/core/psa_crypto_slot_management.c \
+        mbedtls/tf-psa-crypto/core/psa_crypto_storage.c \
+        mbedtls/tf-psa-crypto/core/psa_its_file.c \
+        mbedtls/tf-psa-crypto/core/psa_crypto_driver_wrappers_no_static.c
+
+MBED_BUILTIN_SOURCES = \
```
airween: Let me check the new version on `mbedtls`, and make sure that all of these sources are necessary.

src/parser/seclang-scanner.cc
```diff
@@ -1,5 +1,5 @@
 
-#line 2 "seclang-scanner.cc"
+#line 3 "seclang-scanner.cc"
```
airween: Why is this file has changed? `seclang-scanner.cc` (and `seclan-parser.cc` too) is a generated file by Bison, and it's necessary to change if there is any change in parser's source. But I don't see any changes there.

src/unique_id.cc
```diff
@@ -72,7 +73,7 @@ void UniqueId::fillUniqueId() {
 
     data = macAddress + name;
 
-    this->uniqueId_str = Utils::Sha1::hexdigest(data);
+    this->uniqueId_str = Utils::Sha256::hexdigest(data);
```
airween: Could you explain why did you changed the `sha1` by `sha256`? This will gives a longer ID (`sha1` length is 40 char hex, `sha256`'s 64 char hex) - are we sure users want to use that?

And I'm not sure here `sha256` gives any benefit (strong of cryptography is no matter here).

airween (changes requested):

src/utils/md5.h
```diff
@@ -1,32 +1,44 @@
...
- *
- * You may not use this file except in compliance with
- * the License.  You may obtain a copy of the License at
- *
- *     http://www.apache.org/licenses/LICENSE-2.0
- *
- * If any of the files related to licensing are missing or if you have any
- * other questions related to licensing please contact Trustwave Holdings, Inc.
- * directly using the email address security@modsecurity.org.
+ * Copyright (c) 2015 - 2021 Trustwave Holdings, Inc.
```
airween: As I know the Apache license recommends to use the version that you removed - see:

https://www.apache.org/licenses/LICENSE-2.0#apply

I don't insists this version, but I would insist to use the same syntax in all files.

Or we can use the [SPDX](https://spdx.org/licenses/) identifier.

airween (changes requested):

Easton97-Jens:
I have made all the necessary adjustments. Everything now works as expected with the new test and runs without errors

airween:
> I have made all the necessary adjustments. Everything now works as expected with the new test and runs without errors

Unfortunately none of the regression tests were success, all were failed. Please take a look at them.

Easton97-Jens:
Only the new test is expected to work.

airween:
> Only the new test is expected to work.

I meant all regression tests must be success, including [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/23705576195/job/69056777354?pr=3477#step:5:2478) one too.

[Here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/23705576195?pr=3477) you can see all tests.

Easton97-Jens:
That would require an adjustment to the ci.yml file. Should I correct it or edit the teama file in a new pull?

PR #3479, opened 2025-12-26
Refine SecRequestBodyNoFilesLimit test cases to cover edge scenarios

hnakamur:
## what

Extend and adjust existing tests for the `SecRequestBodyNoFilesLimit` directive in config-body_limits.json to ensure correct handling of edge boundary values.

Also fix the Content-Length of the requests. Note that an LF is added for each line in the body field. see https://github.com/owasp-modsecurity/ModSecurity/blob/cebcde3b9c4ee2620018955507943cb0023d103f/test/regression/regression_test.cc#L51

```
$ echo "param1=value1&param2=value2&param3=value3" | wc -c
42
```

## why

The existing tests do not cover edge boundary lengths. As a result, we cannot verify whether a request body whose length is exactly equal to `SecRequestBodyNoFilesLimit` is accepted.

## references

None.

### DIFF
--- a/test/test-cases/regression/config-body_limits.json
+++ b/test/test-cases/regression/config-body_limits.json
@@ -404,7 +404,7 @@
         "Host":"localhost",
         "User-Agent":"curl/7.38.0",
         "Accept":"*/*",
-	"Content-Length": "41",
+        "Content-Length": "42",
         "Content-Type": "application/x-www-form-urlencoded"
       },
       "uri":"/",
@@ -430,7 +430,7 @@
     "rules":[
       "SecRuleEngine On",
       "SecRequestBodyAccess On",
-      "SecRequestBodyNoFilesLimit 20",
+      "SecRequestBodyNoFilesLimit 41",
       "SecRule REQBODY_ERROR \"!@eq 0\" \"id:'200002', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2\""
     ]
   },
@@ -451,7 +451,7 @@
         "Host":"localhost",
         "User-Agent":"curl/7.38.0",
         "Accept":"*/*",
-	"Content-Length": "41",
+        "Content-Length": "42",
         "Content-Type": "application/x-www-form-urlencoded"
       },
       "uri":"/",
@@ -476,7 +476,7 @@
     "rules":[
       "SecRuleEngine On",
       "SecRequestBodyAccess On",
-      "SecRequestBodyNoFilesLimit 60",
+      "SecRequestBodyNoFilesLimit 42",
       "SecRule REQBODY_ERROR \"!@eq 0\" \"id:'200002', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2\""
     ]
   },


airween:
Hi @hnakamur,


thanks for this PR.

> Extend and adjust existing tests for the `SecRequestBodyNoFilesLimit` directive in config-body_limits.json to ensure correct handling of edge boundary values.
> 
> Also fix the Content-Length of the requests. Note that an LF is added for each line in the body field.

I think `Content-Length` field does not contain the LF in reality.

Also note, that libmodsecurity3 does not correspond to `Content-Length` value with length of request body.

It counts the added bytes and compares the current length with the given value: https://github.com/owasp-modsecurity/ModSecurity/blob/09fec93cb7621175a1631b630f482a87dca1e0b8/src/transaction.cc#L936-L938

> see:
> https://github.com/owasp-modsecurity/ModSecurity/blob/cebcde3b9c4ee2620018955507943cb0023d103f/test/regression/regression_test.cc#L51

you're right, but I think we should remove that "extra" `\n` at the end of the concatenation instead of increasing the (unused) `Content-Length` value.

> ```
> $ echo "param1=value1&param2=value2&param3=value3" | wc -c
> 42
> ```

Yes, but as I explained above, I think we should calculate the length without LF:

```bash
$ echo -n "param1=value1&param2=value2&param3=value3" | wc -c
41
```

> The existing tests do not cover edge boundary lengths. 

Do we need that edge boundary (probably you mean the "extra" `\n`)?

> As a result, we cannot verify whether a request body whose length is exactly equal to `SecRequestBodyNoFilesLimit` is accepted.

Well, to see the valid value could be helpfully, eg. if we put a valid number into CL header. But again: CL header does not have any effect, it's just there... To show the info about the request length is important, but I think that's only an informative thing.

What do you think?

hnakamur:
I agree with you about removing an extra `\n` at the end.

Actually I found it gets in the way for testing the edge cases, so I removed an extra `\n` at commit [400e871](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/400e871393c7c87e9eb84b703d82c5fb867d497e)  in  the https://github.com/owasp-modsecurity/ModSecurity/pull/3476 and add `\n` to test cases at commit [6382a3c](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/6382a3c85783b6bf961a1d2f884b5ea6be25760b).

For regression tests, the value of `Content-Length` header is not used. It is tedious for updating the values, so I think it is better to remove `Content-Length` headers which are not needed.

Or we can modify regression test to add the calculated `Content-Length` from the body specified in tests automatically when running if the `Content-Length` is not specified in tests. If a `Content-Length` header is specified in tests, those values are used without being overwritten by the calculated values.

Maybe I should close this pull request for now. I’d like to hear your thoughts.

airween:
> I agree with you about removing an extra `\n` at the end.

I think now I understand the idea why was it added... Probably the author wanted to substitute the empty lines in some special cases, but... I'm not sure the result **is** what they wanted.

> Actually I found it gets in the way for testing the edge cases, so I removed an extra `\n` at commit [400e871](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/400e871393c7c87e9eb84b703d82c5fb867d497e) in the #3476 and add `\n` to test cases at commit [6382a3c](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/6382a3c85783b6bf961a1d2f884b5ea6be25760b).

Thank you for this - this helped me to understand the original idea.

I'm not sure we should add `\n` to everywhere, eg [this one](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/6382a3c85783b6bf961a1d2f884b5ea6be25760b#diff-f79acd5e52015031df3c992c399f3b7be28df93c8f4ef57c67e586d5dcf3b5a8R31) is not necessary.

But in most cases, eg. where the payload contains more lines (XML, multipart), there it needs the `\n`.

> For regression tests, the value of `Content-Length` header is not used. It is tedious for updating the values, so I think it is better to remove `Content-Length` headers which are not needed.
> Or we can modify regression test to add the calculated `Content-Length` from the body specified in tests automatically when running if the `Content-Length` is not specified in tests. If a `Content-Length` header is specified in tests, those values are used without being overwritten by the calculated values.

Personally I don't like implicit use of values, and as we discussed it does not have any role. I think it's more informative than usable. Therefore I wouldn't support to remove or implicit calculation, let the developer to fill that with a correct value.
 
> Maybe I should close this pull request for now. I’d like to hear your thoughts.

What if we add extra `\n` only if the [node->u.array.len](https://github.com/owasp-modsecurity/ModSecurity/blob/400e871393c7c87e9eb84b703d82c5fb867d497e/test/regression/regression_test.cc#L48) is greater than 1? (Just a quick idea...)

airween:
Also, could you rebase your V3 related PR branches with the current v3/master? I fixed some cppcheck issues in #3485.

hnakamur:
> Personally I don't like implicit use of values, and as we discussed it does not have any role. I think it's more informative than usable. Therefore I wouldn't support to remove or implicit calculation, let the developer to fill that with a correct value.

I agree the explicit values are clearer. I added an option to show the request body length to regression test at https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/36c5c39778551b595a834200ef6850ef02f558de, so I think now it is easy to add Content-Length with a correct value.

> What if we add extra \n only if the [node->u.array.len](https://github.com/owasp-modsecurity/ModSecurity/blob/400e871393c7c87e9eb84b703d82c5fb867d497e/test/regression/regression_test.cc#L48) is greater than 1? (Just a quick idea...)

This prevents the request body with multiple lines but not `\n` at the end. For example,

```
[
  "a",
  "b"
]
```
becomes `"a\nb\n"`, and we cannot create `"a\nb"` test case.

Another idea is adding `\n` at end of each line except the last line. Actually I tried it at https://github.com/hnakamur/ModSecurity/commit/30f108d79323cde6d319c232017bc5d46bebe155.
However, with this version, we have to write
```
[
  "a\r",
  "b\r\n"
]
```
to create the body `"a\r\nb\r\n"`.

I think it is confusing to have implicit `\n` and explicit `\n` mixed.
So I think it is better all `\n` to be explicit. That is implemented at [400e871](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/400e871393c7c87e9eb84b703d82c5fb867d497e).

With this implementation
```
[
  "a\r\n",
  "b\r\n"
]
```
becomes `"a\r\nb\r\n"`, and
```
[
  "a\r\n",
  "b\r"
]
```
becomes `"a\r\nb\r"`. It is clearer since it is all `\n` are explicit and no implicit values.

> I'm not sure we should add \n to everywhere, eg [this one](https://github.com/owasp-modsecurity/ModSecurity/pull/3476/commits/6382a3c85783b6bf961a1d2f884b5ea6be25760b#diff-f79acd5e52015031df3c992c399f3b7be28df93c8f4ef57c67e586d5dcf3b5a8R31) is not necessary.

I agree a lot of newlines are not necessary. I added `\n` to everywhere after chaning the implementation of concatenating multiple body lines to keep tests unchanged. We can then modify test cases as needed.

airween:
Hi @hnakamur,

> I agree the explicit values are clearer. I added an option to show the request body length to regression test at [36c5c39](https://github.com/owasp-modsecurity/ModSecurity/commit/36c5c39778551b595a834200ef6850ef02f558de), so I think now it is easy to add Content-Length with a correct value.

great.
 
> > What if we add extra \n only if the [node->u.array.len](https://github.com/owasp-modsecurity/ModSecurity/blob/400e871393c7c87e9eb84b703d82c5fb867d497e/test/regression/regression_test.cc#L48) is greater than 1? (Just a quick idea...)
> 
> This prevents the request body with multiple lines but not `\n` at the end. For example,

ah, you're definitely right.

> I agree a lot of newlines are not necessary. I added `\n` to everywhere after chaning the implementation of concatenating multiple body lines to keep tests unchanged. We can then modify test cases as needed.

I like this idea.

But now the best thing would be that we close this PR and you open a new one with modified regression test (remove adding of extra `\n`) and modified test cases where it's necessary. I also like the idea to show the content length.

hnakamur:
Thank you for your comments!
I opened a new PR https://github.com/owasp-modsecurity/ModSecurity/pull/3486 and I am closing this PR now.

PR #3480, opened 2025-12-26
v2: Accept requests whose body length is exactly equal to SecRequestBodyNoFilesLimit

hnakamur:
## what

Accept requests whose body length is exactly equal to `SecRequestBodyNoFilesLimit`.

## why

The current implementation rejects requests when their body length is equal to `SecRequestBodyNoFilesLimit`.
However, according to the documented behavior, requests up to and including this limit should be accepted.

## references

- The documentation states: “Configures the maximum request body size ModSecurity will accept for buffering, excluding the size of any files being transported in the request.” https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit
- Tests verifying that requests with a body length equal to `SecRequestBodyNoFilesLimit` are accepted in v3:: https://github.com/owasp-modsecurity/ModSecurity/pull/3479

### DIFF
--- a/apache2/msc_reqbody.c
+++ b/apache2/msc_reqbody.c
@@ -406,7 +406,7 @@ apr_status_t modsecurity_request_body_store(modsec_rec *msr,
     }
 
     /* Check that we are not over the request body no files limit. */
-    if (msr->msc_reqbody_no_files_length >= (unsigned long) msr->txcfg->reqbody_no_files_limit) {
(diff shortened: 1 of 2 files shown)


airween (approved):

airween:
Hi @hnakamur,

nice catch, thank you! I'm surprised this operational anomaly has remained hidden until now.

Special thanks for new tests.

PR #3482, opened 2025-12-27
Fix multipart boundaries in regression tests

hnakamur:
## what

- Remove the redundant "--" prefix from boundaries.
- Add the missing Content-Type header for multipart bodies.

## why

- To make it clear that the tests assume the multipart body is correct and are intended to verify other aspects.

## references

None

### DIFF
--- a/test/test-cases/regression/config-body_limits.json
+++ b/test/test-cases/regression/config-body_limits.json
@@ -95,7 +95,8 @@
       "headers":{  
         "Host":"localhost",
         "User-Agent":"curl/7.38.0",
-        "Accept":"*/*"
+        "Accept":"*/*",
+        "Content-Type": "multipart/form-data; boundary=------------------------756b6d74fa1a8ee2"
       },
       "uri":"/?key=value&key=other_value",
       "method":"POST",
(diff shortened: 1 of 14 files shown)


airween:
Hi @hnakamur,

thanks for this PR - let me review all changes later, but after a quick review, it seems everything is okay.

test/test-cases/regression/config-body_limits.json
```diff
@@ -95,7 +95,8 @@
       "headers":{  
         "Host":"localhost",
         "User-Agent":"curl/7.38.0",
-        "Accept":"*/*"
+        "Accept":"*/*",
+        "Content-Type": "multipart/form-data; boundary=------------------------756b6d74fa1a8ee2"
```
airween, line 99: I was always sure (and still am) that these tests contain many failures. Nonetheless they are success because they don't test what the actual design of the test is - like this one. This is a multipart request, but the engine tests only the `SecRequestBodyLimit`.

Never mind, it's good to see that you caught that :smiley:.

airween (approved):

airween:
Thank you @hnakamur!

Issue #2146, opened 2019-08-14
REQUEST_BODY is set when it should not according to documentation


mirkodziadzka-avi:
I think there is a mismatch between modsec-3 implementation and modsec documentation.

According to https://github.com/SpiderLabs/ModSecurity/wiki/Reference-Manual-%28v2.x%29#request_body
the REQUEST_BODY variable should be set only under 2 conditions.

1. If the WWWFormURLEncoded body processor is used
2. If no body processor is used and ctl:forceRequestBodyVariable is used in phase:1

In current v3 code, this variables is always set.

It is also expected to be set in regression tests, for example test/test-cases/regression/variable-REQUEST_BODY.json. This test is using the Multipart processor (set from C code and not from modsec language)

However, the regression tests of the OWASP Modsec CoreRuleset team (https://github.com/SpiderLabs/owasp-modsecurity-crs/tree/v3.2/dev/util/regression-tests/tests) expect that REQUEST_BODY is not set in some cases. For example in regression test https://github.com/SpiderLabs/owasp-modsecurity-crs/blob/v3.2/dev/util/regression-tests/tests/REQUEST-944-APPLICATION-ATTACK-JAVA/944120.yaml#L128 against the rule

```
# Magic bytes detected and payload included possibly RCE vulnerable classess detected and process execution methods detected
# anomaly score set to critical as all conditions indicate the request try to perform RCE.
SecRule ARGS|ARGS_NAMES|REQUEST_COOKIES|!REQUEST_COOKIES:/__utm/|REQUEST_COOKIES_NAMES|REQUEST_BODY|REQUEST_HEADERS|XML:/*|XML://@* \
    "@rx (?:clonetransformer|forclosure|instantiatefactory|instantiatetransformer|invokertransformer|prototypeclonefactory|prototypeserializationfactory|whileclosure|getproperty|filewriter|xmldecoder)" \
    "id:944120,\
    phase:2,\
    block,\
    t:none,t:lowercase,\
    log,\
    msg:'Remote Command Execution: Java serialization (CVE-2015-5842)',\
    logdata:'Matched Data: %{MATCHED_VAR} found within %{MATCHED_VAR_NAME}',\
    tag:'application-multi',\
    tag:'language-java',\
    tag:'platform-multi',\
    tag:'attack-rce',\
    tag:'OWASP_CRS/WEB_ATTACK/COMMAND_INJECTION',\
    tag:'WASCTC/WASC-31',\
[...]

PR #3483, opened 2025-12-31
v2: Allow partial processing of XML, JSON, and multipart request body 

hnakamur:
## what

- This pull request prevents a parse error that occurs when `SecRequestBodyLimitAction` is set to `ProcessPartial` and the length of an XML or JSON request body exceeds the configured limit.

## why

- When `SecRequestBodyLimitAction` is `ProcessPartial`, the expected behavior is to process the partial XML, JSON, or multipart body up to the defined limit.
- The request should then be rejected if any invalid values are found within that processed part, or passed through if the processed part is valid.
- The previous implementation failed to handle the partial body correctly, leading to a parse error when the limit was exceeded.

## references

The same modification for v3 is at: https://github.com/owasp-modsecurity/ModSecurity/pull/3476


airween:
Hi @hnakamur,

thanks for this PR too.

> * When `SecRequestBodyLimitAction` is `ProcessPartial`, the expected behavior is to process the partial XML, JSON, or multipart body up to the defined limit.

I think `SecRequestBodyLimitAction` regards not only for XML, JSON or multipart bodies - but also to www-url-encoded body too. Consider this setup:
```
SecRequestBodyLimitAction ProcessPartial
SecRequestBodyNofilesLimit 10
```

and the payload:

```
aaa=123&bbb=123
```
then the `aaa` variable must have seen in `ARGS` with value `123`.

@dune73 what do you think?

> The request should then be rejected if any invalid values are found within that processed part, or passed through if the processed part is valid.

How can we be sure that the payload is valid, if the error is in the chunked part? Eg. consider the config above and this JSON:
```json
{"a":1,"b":2,[1]]}
```
(the second `]` makes the JSON non well-formed, but the key `a` must have seen in `ARGS`).

You helped me in #3456, which isn't merged yet (I'm waiting for a review and an approve). But could you add a new test case to the v2's source tree?

hnakamur:
Hello @airween. Thanks for your comment.

I added tests for url-encoded, JSON, and XML request bodies. It appears that the URL-encoded parser already works as expected with `ProcessPartial`.

By the way, I realized that it would be better to add more tests for multipart requests. I think it will take me a day or two to do so.

hnakamur:
> It appears that the URL-encoded parser already works as expected with ProcessPartial.

I was mistaken about that. I modified the URL-encoded parser to include names and values up to the delimiter before the limit when SecRequestBodyLimitAction is set to ProcessPartial and the request body exceeds SecRequestBodyLimit.

I also modified the multipart parser to include parts whose boundary appears before the limit under the same conditions.

In addition, I added tests to cover these changes.

tests/regression/config/10-request-directives.t
```diff
@@ -578,10 +578,10 @@
 		SecRequestBodyLimit 131072
 	),
 	match_log => {
-		error => [ qr/Multipart parsing error: Multipart: Final boundary missing./, 1],
+		-error => [ qr/Multipart parsing error: Multipart: Final boundary missing./, 1],
 	},
 	match_response => {
-		status => qr/^500$/,
+		status => qr/^200$/,
```
airween, line 618: It's a good question that we should allow to process a multipart request if there is no final boundary (because it's outside of the limited buffer length).

Consider we allow this - but what about the rules [200003](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/modsecurity.conf-recommended#L79-L93) and [200004](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/modsecurity.conf-recommended#L97-L98)?

At the moment I really don't know, just asking. It would be nice to discuss it.

tests/regression/config/10-request-directives.t
```diff
@@ -643,110 +643,1820 @@
...
+{
+	type => "config",
+	comment => "SecRequestBodyLimitAction ProcessPartial (multipart/bad_name before limit)",
+	conf => qq(
+		SecRuleEngine On
+		SecDebugLog $ENV{DEBUG_LOG}
+		SecDebugLogLevel 9
+		SecRequestBodyAccess On
+		SecRequestBodyLimitAction ProcessPartial
+		SecRequestBodyLimit 59
```
airween, line 690: To be clear, please add here and all other tests where you check the limit(s) the `SecRequestBodyNoFilesLimit`.

I think that would be useful to avoid the unwanted side effects (eg. not the relevant settings applied and a negative test is successed).

tests/regression/config/10-request-directives.t
```diff
@@ -643,110 +643,1820 @@
...
+{
+	type => "config",
+	comment => "SecRequestBodyLimitAction ProcessPartial (multipart/bad_name after limit)",
+	conf => qq(
+		SecRuleEngine On
+		SecDebugLog $ENV{DEBUG_LOG}
+		SecDebugLogLevel 9
+		SecRequestBodyAccess On
+		SecRequestBodyLimitAction ProcessPartial
+		SecRequestBodyLimit 58
```
airween, line 719: Please add the `SecRequestBodyNoFilesLimit` too with an exact value.

tests/regression/config/10-request-directives.t
```diff
@@ -643,110 +643,1820 @@
...
+{
+	type => "config",
+	comment => "SecRequestBodyLimitAction ProcessPartial (url-encoded/partial/bad_name without value without delimeter before limit)",
+	conf => qq(
+		SecRuleEngine On
+		SecDebugLog $ENV{DEBUG_LOG}
+		SecDebugLogLevel 9
+		SecRequestBodyAccess On
+		SecRequestBodyLimitAction ProcessPartial
+		SecRequestBodyLimit 8
```
airween, line 1888: Please add the `SecRequestBodyNoFilesLimit` with an exact value.

In these cases (where you check non-multipart requests) could you add more tests with higher `SecRequestBodyLimit` values?

airween (changes requested):

airween:
Hi @hnakamur,

thank you again for your work, I think this is a very important fix.

Please take a look at GH Copilot review, there are a logic change and a typo in the code.

I also added a few notices to tests.

apache2/apache2_io.c
```diff
@@ -351,11 +352,13 @@ apr_status_t read_request_body(modsec_rec *msr, char **error_msg) {
...
-    if (rcbe < 0) {
-        return HTTP_INTERNAL_SERVER_ERROR;
+    if (msr->txcfg->if_limit_action == REQUEST_BODY_LIMIT_ACTION_REJECT) {
+        if (rcbe == -5) {
+            return HTTP_REQUEST_ENTITY_TOO_LARGE;
+        }
+        if (rcbe < 0) {
+            return HTTP_INTERNAL_SERVER_ERROR;
+        }
     }
```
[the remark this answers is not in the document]
hnakamur: This change causes a parser error returns a response with status code 500, which is not wanted since this case is not an internal server error.
In case of a parse error, REQBODY_ERROR is set to 1, which can be used by a rule to return a 400 Bad Request response, for example:
```
SecRule REQBODY_ERROR "!@eq 0" "id:'200001', phase:2,t:none,log,deny,status:400,msg:'Failed to parse request body.',logdata:'%{reqbody_error_msg}',severity:2"
```
Side note: However, I found that no rule in [OWASP CRS 4.22.0](https://github.com/coreruleset/coreruleset/tree/v4.22.0) checks `REQBODY_ERROR`.

This is wanted behavior for both of `REQUEST_BODY_LIMIT_ACTION_REJECT` and `REQUEST_BODY_LIMIT_ACTION_PARTIAL`.

Therefore I think the right modification here is:

```
    if (rcbe == -5 && msr->txcfg->if_limit_action == REQUEST_BODY_LIMIT_ACTION_REJECT) {
        return HTTP_REQUEST_ENTITY_TOO_LARGE;
    }
```

tests/regression/config/10-request-directives.t
hnakamur: I think we should allow no final boundary for ProcessPartial consiering definition of it.

I added rules [200003](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/modsecurity.conf-recommended#L79-L93) and [200004](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/modsecurity.conf-recommended#L97-L98) to this test and Reject version of this test in commit [f679e11](https://github.com/owasp-modsecurity/ModSecurity/pull/3483/commits/f679e11d3238d78a71ce3b34f35af00699105603):

* `SecRequestBodyLimitAction ProcessPartial (multipart/greater - chunked)"`
* `SecRequestBodyLimitAction Reject (multipart/greater - chunked)`

I confirmed that both test pass.

This means that for ProcessPartial `MULTIPART_STRICT_ERROR` and `MULTIPART_UNMATCHED_BOUNDARY` is 0 when the body has no final boundary in the part before the limit and the body length exceeds the limit since the body may have the final boundary after the limit.
For Reject, the request whose body length exceeds the limit is rejected with status 413 before evaluating rules, so it does not matter what value `MULTIPART_STRICT_ERROR` and `MULTIPART_UNMATCHED_BOUNDARY`  are.

tests/regression/config/10-request-directives.t
hnakamur: Before adding tests for `SecRequestBodyNoFilesLimit`, let me understand it for `ProcessPartial`.

Though the document for [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit) does not mention for `ProcessPartial` case, I believe we should not reject a request whose length exceeds the `SecRequestBodyNoFilesLimit` for `ProcessPartial` case. Is this correct?

And I found the current implemention is handling the request whose length exceeds `SecRequestBodyNoFilesLimit` inconsistently between `modsecurity_request_body_store` and `modsecurity_request_body_end`.

I fixed it at https://github.com/hnakamur/ModSecurity/commit/96a8326ebdc6f1a71dd16289848c4d310290da6c.
Could you take a look?

Should I include this commit to this pull request or should I create another pull request for `SecRequestBodyNoFilesLimit`?

I'll add tests for `SecRequestBodyNoFilesLimit` after the above fix is settled.

apache2/apache2_io.c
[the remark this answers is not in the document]
airween: > This change causes a parser error returns a response with status code 500, which is not wanted since this case is not an internal server error.

okay, then we can make this as resolved. Thanks.

> Side note: However, I found that no rule in [OWASP CRS 4.22.0](https://github.com/coreruleset/coreruleset/tree/v4.22.0) checks `REQBODY_ERROR`.
> 
> This is wanted behavior for both of `REQUEST_BODY_LIMIT_ACTION_REJECT` and `REQUEST_BODY_LIMIT_ACTION_PARTIAL`.

Good advice - would you mind to open a PR or just an issue under [coreruleset](https://github.com/coreruleset/coreruleset)?
 
> Therefore I think the right modification here is:
> 
> ```
>     if (rcbe == -5 && msr->txcfg->if_limit_action == REQUEST_BODY_LIMIT_ACTION_REJECT) {
>         return HTTP_REQUEST_ENTITY_TOO_LARGE;
>     }
> ```

Do you want to add/change this one too? I'm waiting for your response, until that I don't resolve this comment.

tests/regression/config/10-request-directives.t
airween: > I added rules ...
> ...
> I confirmed that both test pass.

Could you fetch and merge the current v2/master? A few days ago I could merge the PR which uses regression tests in GH CI workflow. Then we could see all new tests here too. Thanks.

> This means that for ProcessPartial `MULTIPART_STRICT_ERROR` and `MULTIPART_UNMATCHED_BOUNDARY` is 0 when the body has no final boundary in the part before the limit and the body length exceeds the limit since the body may have the final boundary after the limit. For Reject, the request whose body length exceeds the limit is rejected with status 413 before evaluating rules, so it does not matter what value `MULTIPART_STRICT_ERROR` and `MULTIPART_UNMATCHED_BOUNDARY` are.

Great, thanks for the explanation and for the test too.

tests/regression/config/10-request-directives.t
airween: > Before adding tests for `SecRequestBodyNoFilesLimit`, let me understand it for `ProcessPartial`.

Right, thanks.

> Though the document for [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit) does not mention for `ProcessPartial` case, I believe we should not reject a request whose length exceeds the `SecRequestBodyNoFilesLimit` for `ProcessPartial` case. Is this correct?

Well, that's a good question.

@dune73 what do you think?
 
> And I found the current implemention is handling the request whose length exceeds `SecRequestBodyNoFilesLimit` inconsistently between `modsecurity_request_body_store` and `modsecurity_request_body_end`.
> 
> I fixed it at [hnakamur@96a8326](https://github.com/hnakamur/ModSecurity/commit/96a8326ebdc6f1a71dd16289848c4d310290da6c). Could you take a look?

Let me take a look at this soon.

> Should I include this commit to this pull request or should I create another pull request for `SecRequestBodyNoFilesLimit`?

It would be great if you merge the current v2/master branch into your working branch. Then after each commit the whole regression test will run, then we can see the tests are passed.

> I'll add tests for `SecRequestBodyNoFilesLimit` after the above fix is settled.

Okay. I'll let you know if I'm done. Thanks.

airween: @hnakamur,

> Before adding tests for `SecRequestBodyNoFilesLimit`, let me understand it for `ProcessPartial`.

I've reviewed this comment again and tried to understand the documentation too. I think that's a bit confused.
 
> Though the document for [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit) does not mention for `ProcessPartial` case, I believe we should not reject a request whose length exceeds the `SecRequestBodyNoFilesLimit` for `ProcessPartial` case. Is this correct?

I think the correct place is [SecRequestBodyLimitAction](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secrequestbodylimitaction) and not `SecRequestBodyNoFilesLimit` - this one does not mention limit action at all, but the `SecRequestBodyLimitAction` mentions the `SecRequestBodyLimit` (and only that one).

If we try to follow the explained behavior in the documentation, the we should apply the limit action method only for body limit (and not for no files limit).

But `SecRequestBodyNoFilesLimit` usually much lower than `SecRequestBodyLimit` (for eg. the default values in `modsecurity.conf` are 128k and 12.5MB). So a user who does not touch these values will never run into the limit action problem. If someone wants to use this feature, then they need to set the same value. But if they increase the no files limit, that's a potentially DoS cause. If they decrease the request body limit, then they won't be able to upload big files.

Furthermore: XML, JSON and urlencoded payloads belong to no files limit. Only multipart requests consider the whole body.

> And I found the current implemention is handling the request whose length exceeds `SecRequestBodyNoFilesLimit` inconsistently between `modsecurity_request_body_store` and `modsecurity_request_body_end`.
> 
> I fixed it at [hnakamur@96a8326](https://github.com/hnakamur/ModSecurity/commit/96a8326ebdc6f1a71dd16289848c4d310290da6c). Could you take a look?

I still wasn't able to review that, I hope I can do that this weekend.

rainerjung:
See also #2093

airween:
Hi @rainerjung,

> See also #2093

thanks for bringing up this one.

Actually, it would be nice to discuss the expected behavior.

The contradiction I feel is that we want to process the body even if it's bigger than the limit was set in `SecRequestBodyNoFilesLimit`. I think [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3483#discussion_r2719937790) comment from @hnakamur, especially this part highlights perfectly the problem:

> Though the document for [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit) does not mention for `ProcessPartial` case, I believe we should not reject a request whose length exceeds the `SecRequestBodyNoFilesLimit` for `ProcessPartial` case. Is this correct?

If we agree that `SecRequestBodyNoFilesLimit` does not have any relation to [SecRequestBodyLimitAction](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secrequestbodylimitaction). It has any effect only to `SecRequestBodyLimit`. This one holds the value of the whole request body's size, but it's usually much-much higher (10 times in the default `modsecurity.conf`) then `SecRequestBodyNoFilesLimit`.

Therefore `JSON`, `XML` and `URLENCODED` bodies reach this limit (`SecRequestBodyNoFilesLimit`) sooner than the limit which has an exemption (I mean `SecRequestBodyLimit` and `SecRequestBodyLimitAction`). If a `JSON` body reaches this limit, then we shouldn't continue the process, because there is no exemption - if we agree the documentation.

So, this is what I want to ask you (you guys, who uses ModSecurity): what would be the expected behavior?

hnakamur:
I have read #2093 and agree with the comment https://github.com/owasp-modsecurity/ModSecurity/issues/2093#issuecomment-639670880. And I think that adding `SkipProcess` should be handled in another issue if it is really needed. I do not need `SkipProcess` and all I need is that `ProcessPartial` processes the partial body as the name suggests.

As for `SecRequestBodyNoFilesLimit`, how about adding a new directive `SecRequestBodyProcessPartialLimitType LimitOnly|MinOfNoFilesLimitAndLimit` to change `ProcessPartial` to process the partial body up to the length of `min(SecRequestBodyNoFilesLimit, SecRequestBodyLimit)`?
The default value is `LimitOnly` (only use  `SecRequestBodyLimit`) for compatibility.

airween:
> I have read #2093 and agree with the comment [#2093 (comment)](https://github.com/owasp-modsecurity/ModSecurity/issues/2093#issuecomment-639670880). And I think that adding `SkipProcess` should be handled in another issue if it is really needed. I do not need `SkipProcess` and all I need is that `ProcessPartial` processes the partial body as the name suggests.

I also don't need `SkipProcess` mechanism, and honestly I don't see the advantage of that. But I don't want to close out the chance that we implement that.
 
> As for `SecRequestBodyNoFilesLimit`, how about adding a new directive `SecRequestBodyProcessPartialLimitType LimitOnly|MinOfNoFilesLimitAndLimit` to change `ProcessPartial` to process the partial body up to the length of `min(SecRequestBodyNoFilesLimit, SecRequestBodyLimit)`? The default value is `LimitOnly` (only use `SecRequestBodyLimit`) for compatibility.

For the first sight I think it would be a bit too complicated.

How about the new directive `SecRequestBodyNoFilesLimitAction`, which would correspond to `SecRequestBodyNoFilesLimit` - as `SecRequestBodyLimitAction` to `SecRequestBodyLimit`.

I think this is more clear, and probably we could implement this easier.

What do you think?

hnakamur:
Thanks for you comment! I agree with you that adding the new directive `SecRequestBodyNoFilesLimitAction` is simpler.

Let me clarify the actual behavior for the combination of `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction`.

Let the following two limit values be the default here.
```
SecRequestBodyNoFilesLimit   1048576 # 1MiB
SecRequestBodyLimit        134217728 # 128MiB
```

I suppose we expect the following behaviors. Could you check they are correct?

- Case 1: Both actions are `Reject`:
  ```
  SecRequestBodyNoFilesLimitAction Reject
  SecRequestBodyLimitAction        Reject
  ```

  | request length | body processing | 
  |----------------|-----------------|
  | x <= 1MiB      | full body check |
  | 1MiB < x       | reject          |


- Case 2: Both actions are `ProcessPartial`:
  ```
  SecRequestBodyNoFilesLimitAction ProcessPartial
  SecRequestBodyLimitAction        ProcessPartial
  ```

  | request length | body processing    | 
  |----------------|--------------------|
  | x <= 1MiB      | full body check    |
  | 1MiB < x       | partial body check |


- Case 3: `ProcessPartial` for `NoFilesLimit`, `Reject` for `Limit`:
  ```
  SecRequestBodyNoFilesLimitAction ProcessPartial
  SecRequestBodyLimitAction        Reject
  ```

  | request length     | body processing    | 
  |--------------------|--------------------|
  | x <= 1MiB          | full body check    |
  | 1MiB < x <= 128MiB | partial body check |
  | 128MiB < x         | reject             |


- Case 4: `Reject` for `NoFilesLimit`, `ProcessPartial` for `Limit`:
  ```
  SecRequestBodyNoFilesLimitAction Reject
  SecRequestBodyLimitAction        ProcessPartial
  ```

  | request length | body processing | 
  |----------------|-----------------|
  | x <= 1MiB      | full body check |
  | 1MiB < x       | reject          |

  I think `ProcessPartial` for `SecRequestBodyLimitAction` is meaningless for this case.
  Maybe we should reject this combination when loading rules, but I don't know about details on how to implement that yet.

I assume that `SecRequestBodyNoFilesLimit <= SecRequestBodyLimit` is an invariant and always holds true.
Is this correct? If so, maybe we should clarify this in the document.

airween:
Hi @hnakamur,

sorry for the late reply - I wanted to review the engine's behavior as well so I was immersed in the topic. I had a feeling that many users don't understand how this limitations work so I wrote a blog post, see:

https://modsecurity.org/20260222/how-big-is-too-big-a-deep-dive-into-modsecurity-request-body-limits/


> Let me clarify the actual behavior for the combination of `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction`.

I think your descriptions here are good all.
 
> Let the following two limit values be the default here.

See the blog post above.
 
> I assume that `SecRequestBodyNoFilesLimit <= SecRequestBodyLimit` is an invariant and always holds true. Is this correct? If so, maybe we should clarify this in the document.

No, it's not an invariant, anyone can set a different values, engine will accept that. But indeed, that makes no sense.

hnakamur:
Thank you for your blog post!

I think it is helpful for users to understand SecRequestBodyLimit, SecRequestBodyNoFilesLimit, and SecRequestBodyLimitAction.

> Back to PRs. The main concept of the open PRs is to align the “action” directives with the “limit” directives. The proposed new directive is SecRequestBodyNoFilesLimitAction and would follow the behavior of SecRequestBodyLimitAction, but another option is to extend the existing directive’s behavior to handle the case where SecRequestBodyNoFilesLimit is exceeded.

Regarding SecRequestBodyNoFilesLimitAction, I felt some inconsistency in the fact that both file and non-file data are still subject to rule evaluation, yet the behavior changes depending on whether the non-file data exceeds the size specified by SecRequestBodyNoFilesLimit.

airween:
> Regarding SecRequestBodyNoFilesLimitAction, I felt some inconsistency in the fact that both file and non-file data are still subject to rule evaluation, yet the behavior changes depending on whether the non-file data exceeds the size specified by SecRequestBodyNoFilesLimit.

I understand your concern - what do you suggest, how could we resolve this?

For eg. if the CT is multipart, then we ignore the `SecRequestBodyNoFilesLimitAction`?

hnakamur:
I realized that I had not fully understood SecRequestBodyNoFilesLimitAction.

Here is my updated understanding:

The current [SecRequestBodyNoFilesLimit documentation](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit
) states:

> Anything over the limit will be rejected with status code 413 (Request Entity Too Large).

However, there are cases where this behavior is not desirable. Therefore, we introduce a new directive, `SecRequestBodyNoFilesLimitAction`.

If the value is set to `Reject` (the default), a request whose total size of non-file data exceeds this limit will be rejected with status 413 (same as before).

If the value is set to `ProcessPartial`, a request whose total size of non-file data exceeds this limit will be processed only up to the configured limit, and the remaining part will not be processed.

> For eg. if the CT is multipart, then we ignore the SecRequestBodyNoFilesLimitAction?

Is this a typo? Should it be:

> If the Content-Type is NOT multipart, then we ignore SecRequestBodyNoFilesLimitAction?

Here is the rationale for this suggestion:

If the `Content-Type` is NOT multipart, `msc_reqbody_no_files_length` is always equal to `msc_reqbody_length`. Therefore, it does not make sense to configure `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction` differently (for example, `SecRequestBodyNoFilesLimitAction ProcessPartial` and `SecRequestBodyLimitAction Reject`).

On the other hand, if the Content-Type is multipart, msc_reqbody_no_files_length is less than msc_reqbody_length, since `msc_reqbody_no_files_length` excludes file data and multipart boundaries. It makes sense to configure `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction` differently (for example, `SecRequestBodyNoFilesLimitAction ProcessPartial` and `SecRequestBodyLimitAction Reject`).

Therefore:

* If `msc_reqbody_no_files_length` > `SecRequestBodyNoFilesLimit` and `SecRequestBodyNoFilesLimitAction` is set to `ProcessPartial`, only the portion up to the limit will be processed.
* If `msc_reqbody_length` > `SecRequestBodyLimit` and `SecRequestBodyLimitAction` is set to `ProcessPartial`, only the portion up to that limit will be processed.

airween:
> I realized that I had not fully understood SecRequestBodyNoFilesLimitAction.

No worries :)

> Here is my updated understanding:
> 
> The current [SecRequestBodyNoFilesLimit documentation](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyNoFilesLimit) states:
> 
> > Anything over the limit will be rejected with status code 413 (Request Entity Too Large).
> 
> However, there are cases where this behavior is not desirable. Therefore, we introduce a new directive, `SecRequestBodyNoFilesLimitAction`.

yes,
 
> If the value is set to `Reject` (the default), a request whose total size of non-file data exceeds this limit will be rejected with status 413 (same as before).

exactly,
 
> If the value is set to `ProcessPartial`, a request whose total size of non-file data exceeds this limit will be processed only up to the configured limit, and the remaining part will not be processed.

yes.
 
> > For eg. if the CT is multipart, then we ignore the SecRequestBodyNoFilesLimitAction?
> 
> Is this a typo? Should it be:
> 
> > If the Content-Type is NOT multipart, then we ignore SecRequestBodyNoFilesLimitAction?

Sorry, this was a question. I asked about the imagined behavior.

> Here is the rationale for this suggestion:
> 
> If the `Content-Type` is NOT multipart, `msc_reqbody_no_files_length` is always equal to `msc_reqbody_length`. 

this is almost true, but I wouldn't say _`msc_reqbody_no_files_length` is always equal to `msc_reqbody_length`_.

I would say here that `msc_reqbody_no_files_length` is usually lower, therefore even if the payload is lower than `msc_reqbody_length`, the other variable will block the request.

> Therefore, it does not make sense to configure `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction` differently (for example, `SecRequestBodyNoFilesLimitAction ProcessPartial` and `SecRequestBodyLimitAction Reject`).

I wouldn't say this here too. I mean _"it does not make sense to configure..."_. In the blog post, I tried to explain both variable's behavior, regarding the users' habit. Most users configure NoFilesLimit much bigger than the "regular" limit, because both limits apply to different cases.

> On the other hand, if the Content-Type is multipart, msc_reqbody_no_files_length is less than msc_reqbody_length, since `msc_reqbody_no_files_length` excludes file data and multipart boundaries. It makes sense to configure `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction` differently (for example, `SecRequestBodyNoFilesLimitAction ProcessPartial` and `SecRequestBodyLimitAction Reject`).

Exactly.

> Therefore:
> 
>     * If `msc_reqbody_no_files_length` > `SecRequestBodyNoFilesLimit` and `SecRequestBodyNoFilesLimitAction` is set to `ProcessPartial`, only the portion up to the limit will be processed.
> 
>     * If `msc_reqbody_length` > `SecRequestBodyLimit` and `SecRequestBodyLimitAction` is set to `ProcessPartial`, only the portion up to that limit will be processed.

I would just add that the first case applies to request if CT is NOT multipart, and the second case applies the second case if the CT IS multipart.

Consider the situation:
```
SecRequestBodyLimit 13107200
SecRequestBodyNoFilesLimit 131072
SecRequestBodyLimitAction ProcessPartial
SecRequestBodyLimitNoFilesAction ProcessPartial
```

CT is JSON, and payload length is 131073 bytes. In this case, the first 131072 bytes will be processed, (probably the last char will be removed from the JSON), other will be skipped. But this is okay.

Now the next request is a MP request. The `msc_reqbody_no_files_length` reaches the `SecRequestBodyNoFilesLimit` (because - for example - there are 1500 small files (which generates more than 128k "nofiles" part)).

What should we do? I mean from which position will we skip the processing in the multipart body?

hnakamur:
> > Here is the rationale for this suggestion:
> > If the `Content-Type` is NOT multipart, `msc_reqbody_no_files_length` is always equal to `msc_reqbody_length`.
> 
> this is almost true, but I wouldn't say _`msc_reqbody_no_files_length` is always equal to `msc_reqbody_length`_.
> 
> I would say here that `msc_reqbody_no_files_length` is usually lower, therefore even if the payload is lower than `msc_reqbody_length`, the other variable will block the request.

The first sentence of the description of [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secrequestbodynofileslimit) is "Configures the maximum request body size ModSecurity will accept for buffering, excluding the size of any files being transported in the request.".

It says it excludes the size of files. However files can exist only in muiltpart requests. JSON, XML, and URL encoded requests do not have files. So no size is excluded for requests whose `Content-Type` is not multipart.

Therefore `msc_reqbody_no_files_length == msc_reqbody_length` if the `Content-Type` is NOT multipart.

In the current implementation of [modsecurity_request_body_store](https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L318-L439), `msr->msc_reqbody_no_files_length += length;` is executed for XML, JSON, URLENCODED, and `ctl:forceRequestBodyVariable=On` request bodies (where the length is the chunk data length).

I made [a branch](https://github.com/hnakamur/ModSecurity/tree/v2/add_debug_log_and_test_cases_for_reqbody_no_files_length) to verify this by adding debug logs and test cases (commit: https://github.com/hnakamur/ModSecurity/commit/eb4cdb631d0afcaabefe4874240e23524d4fcc34).

In all four cases (XML, JSON, URLENCODED, and `ctl:forceRequestBodyVariable=On`) below, `msc_reqbody_no_files_length` is equal to `msc_reqbody_length`.

```
$ ls regression/misc/99-{forcebodybuf,json,urlencoded,xml}.t | xargs -n 1 ./my_run_test.sh 
=== regression/misc/99-forcebodybuf.t ===
reqbody_buffering: modsecurity_request_body_store added 6 to no_files_length
Input filter: Completed receiving request body (length 6).
=== regression/misc/99-json.t ===
JSON: modsecurity_request_body_store added 7 to no_files_length
Request body no files length: 7
Input filter: Completed receiving request body (length 7).
=== regression/misc/99-urlencoded.t ===
URLENCODED: modsecurity_request_body_store added 23 to no_files_length
Input filter: Completed receiving request body (length 23).
=== regression/misc/99-xml.t ===
XML: modsecurity_request_body_store added 8 to no_files_length
Request body no files length: 8
Input filter: Completed receiving request body (length 8).
```

> > ```
> > * If `msc_reqbody_no_files_length` > `SecRequestBodyNoFilesLimit` and `SecRequestBodyNoFilesLimitAction` is set to `ProcessPartial`, only the portion up to the limit will be processed.
> > 
> > * If `msc_reqbody_length` > `SecRequestBodyLimit` and `SecRequestBodyLimitAction` is set to `ProcessPartial`, only the portion up to that limit will be processed.
> > ```
> 
> I would just add that the first case applies to request if CT is NOT multipart, and the second case applies the second case if the CT IS multipart.
> 
> Consider the situation:
> 
> ```
> SecRequestBodyLimit 13107200
> SecRequestBodyNoFilesLimit 131072
> SecRequestBodyLimitAction ProcessPartial
> SecRequestBodyLimitNoFilesAction ProcessPartial
> ```
> 
> CT is JSON, and payload length is 131073 bytes. In this case, the first 131072 bytes will be processed, (probably the last char will be removed from the JSON), other will be skipped. But this is okay.

yes.

> 
> Now the next request is a MP request. The `msc_reqbody_no_files_length` reaches the `SecRequestBodyNoFilesLimit` (because - for example - there are 1500 small files (which generates more than 128k "nofiles" part)).
> 
> What should we do? I mean from which position will we skip the processing in the multipart body?

I think we should process the first `n` parts
if `msc_reqbody_no_files_length <= SecRequestBodyNoFilesLimit && msc_reqbody_length <= SecRequestBodyLimit` after parsing the `n`-th part
and `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit || msc_reqbody_length > SecRequestBodyLimit` after parsing the `(n+1)`-th part.

I also added multipart test cases at [the above branch](https://github.com/hnakamur/ModSecurity/tree/v2/add_debug_log_and_test_cases_for_reqbody_no_files_length) to understand `msc_reqbody_no_files_length`  (commit: https://github.com/hnakamur/ModSecurity/commit/eb4cdb631d0afcaabefe4874240e23524d4fcc34).

Here is my test script outputs.

form data part cases:

```
~/ghq/github.com/owasp-modsecurity/ModSecurity-v2/tests$ ls regression/misc/99-multipart-*form-data*.t | xargs -n 1 ./my_run_test.sh 
=== regression/misc/99-multipart-form-data.t ===
Multipart: multipart_process_part_header added 45 to no_files_length, data=<Content-Disposition: form-data; name="name"\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Multipart: multipart_process_part_data added 1 to no_files_length, data=<1>
Request body no files length: 48
Input filter: Completed receiving request body (length 68).
=== regression/misc/99-multipart-form-data2.t ===
Multipart: multipart_process_part_header added 45 to no_files_length, data=<Content-Disposition: form-data; name="name"\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Multipart: multipart_process_part_data added 1 to no_files_length, data=<1>
Multipart: multipart_process_part_header added 46 to no_files_length, data=<Content-Disposition: form-data; name="name2"\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Multipart: multipart_process_part_data added 1 to no_files_length, data=<2>
Request body no files length: 97
Input filter: Completed receiving request body (length 127).
=== regression/misc/99-multipart-large-form-data.t ===
Multipart: multipart_process_part_header added 45 to no_files_length, data=<Content-Disposition: form-data; name="name"\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Multipart: multipart_process_part_data added 4096 to no_files_length, data=<1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111>
Multipart: multipart_process_part_data added 4096 to no_files_length, data=<1111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111>
Multipart: multipart_process_part_data added 1 to no_files_length, data=<1>
Request body no files length: 8240
Input filter: Completed receiving request body (length 8260).
```

file part cases:
```
~/ghq/github.com/owasp-modsecurity/ModSecurity-v2/tests$ ls regression/misc/99-multipart-*file*.t | xargs -n 1 ./my_run_test.sh 
=== regression/misc/99-multipart-file.t ===
Multipart: multipart_process_part_header added 67 to no_files_length, data=<Content-Disposition: form-data; name="file1"; filename="name.txt"\r\n>
Multipart: multipart_process_part_header added 26 to no_files_length, data=<Content-Type: plain/text\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Request body no files length: 95
Input filter: Completed receiving request body (length 116).
=== regression/misc/99-multipart-large-file.t ===
Multipart: multipart_process_part_header added 74 to no_files_length, data=<Content-Disposition: form-data; name=; name="file1"; filename="name.txt"\r\n>
Multipart: multipart_process_part_header added 2 to no_files_length, data=<\r\n>
Request body no files length: 76
Input filter: Completed receiving request body (length 8289).
```

So I learned `msc_reqbody_no_files_length`  for multipart requests is equal to:

* form data part: combined size of part headers and form data values
* file part: part headers size.

The boundary sizes are excluded from `msc_reqbody_no_files_length` for both form data and file parts.

airween:
> The first sentence of the description of [SecRequestBodyNoFilesLimit](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#secrequestbodynofileslimit) is "Configures the maximum request body size ModSecurity will accept for buffering, excluding the size of any files being transported in the request.".

yes,
 
> It says it excludes the size of files. However files can exist only in muiltpart requests. JSON, XML, and URL encoded requests do not have files. So no size is excluded for requests whose `Content-Type` is not multipart.

no. If the CT is **NOT**  multipart, then the engine checks the no files limit too. See in my blog post, especially the first block, these:
```
SecRequestBodyLimit 130
SecRequestBodyNoFilesLimit 110

curl -v -H "Content-Type: application/json" -X POST --data @payloadmin4.json http://localhost

ModSecurity: Request body no files data length (120) is larger than the configured limit (110).
```
As you can see here, the no files limit is lower than the regular one, the payload is greater than that limit, and this is what the payload reaches.

> Therefore `msc_reqbody_no_files_length == msc_reqbody_length` if the `Content-Type` is NOT multipart.

the variable in the engine's inside, yes.
 
> In the current implementation of [modsecurity_request_body_store](https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L318-L439), `msr->msc_reqbody_no_files_length += length;` is executed for XML, JSON, URLENCODED, and `ctl:forceRequestBodyVariable=On` request bodies (where the length is the chunk data length).

yes, exactly. This is why I say no multipart requests' payload counts into the no files limit too.

> I made [a branch](https://github.com/hnakamur/ModSecurity/tree/v2/add_debug_log_and_test_cases_for_reqbody_no_files_length) to verify this by adding debug logs and test cases (commit: [hnakamur@eb4cdb6](https://github.com/hnakamur/ModSecurity/commit/eb4cdb631d0afcaabefe4874240e23524d4fcc34)).
> 
> In all four cases (XML, JSON, URLENCODED, and `ctl:forceRequestBodyVariable=On`) below, `msc_reqbody_no_files_length` is equal to `msc_reqbody_length`.

Yes, I think now I see what you mean when you write this, and yes, these are equals, because these both hold the current length.

I'm glad we could clarify this :)

> > Now the next request is a MP request. The `msc_reqbody_no_files_length` reaches the `SecRequestBodyNoFilesLimit` (because - for example - there are 1500 small files (which generates more than 128k "nofiles" part)).
> > What should we do? I mean from which position will we skip the processing in the multipart body?
> 
> I think we should process the first `n` parts if `msc_reqbody_no_files_length <= SecRequestBodyNoFilesLimit && msc_reqbody_length <= SecRequestBodyLimit` after parsing the `n`-th part and `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit || msc_reqbody_length > SecRequestBodyLimit` after parsing the `(n+1)`-th part.
> 
> I also added multipart test cases at [the above branch](https://github.com/hnakamur/ModSecurity/tree/v2/add_debug_log_and_test_cases_for_reqbody_no_files_length) to understand `msc_reqbody_no_files_length` (commit: [hnakamur@eb4cdb6](https://github.com/hnakamur/ModSecurity/commit/eb4cdb631d0afcaabefe4874240e23524d4fcc34)).
> 
> Here is my test script outputs.
> 
> form data part cases:
> 
> ```
> ~/ghq/github.com/owasp-modsecurity/ModSecurity-v2/tests$ ls regression/misc/99-multipart-*form-data*.t | xargs -n 1 ./my_run_test.sh 
> ...
> ```

thanks, I need some time to check this.

> So I learned `msc_reqbody_no_files_length` for multipart requests is equal to:
> 
>     * form data part: combined size of part headers and form data values
> 
>     * file part: part headers size.

exactly.

> The boundary sizes are excluded from `msc_reqbody_no_files_length` for both form data and file parts.

Yes, correct. Let me check your branch's result and come back to continue the discussion :).

hnakamur:
> > It says it excludes the size of files. However files can exist only in muiltpart requests. JSON, XML, and URL encoded requests do not have files. So no size is excluded for requests whose `Content-Type` is not multipart.
> 
> no. If the CT is **NOT** multipart, then the engine checks the no files limit too. See in my blog post, especially the first block, these:

I did not say that the ModSecurity does not check `SecRequestBodyNoFilesLimit` for non-multipart requests.
But I should have clarified I understand the ModSecurity does check `SecRequestBodyNoFilesLimit` for both non-multipart and multipart requests.

> > I made [a branch](https://github.com/hnakamur/ModSecurity/tree/v2/add_debug_log_and_test_cases_for_reqbody_no_files_length) to verify this by adding debug logs and test cases (commit: [hnakamur@eb4cdb6](https://github.com/hnakamur/ModSecurity/commit/eb4cdb631d0afcaabefe4874240e23524d4fcc34)).
> > In all four cases (XML, JSON, URLENCODED, and `ctl:forceRequestBodyVariable=On`) below, `msc_reqbody_no_files_length` is equal to `msc_reqbody_length`.
> 
> Yes, I think now I see what you mean when you write this, and yes, these are equals, because these both hold the current length.
> 
> I'm glad we could clarify this :)

Yes, I'm glad too. I should have made those concrete examples earlier since it explains the details better than my English skills as a non-native speaker.

Also those examples made my understanding of `msc_reqbody_no_files_length` clearer.

> > The boundary sizes are excluded from `msc_reqbody_no_files_length` for both form data and file parts.
> 
> Yes, correct. Let me check your branch's result and come back to continue the discussion :).

Thank your for taking your time.
Please note that the test cases in my branch are not meant to be real tests. The whole purpose is to show the actual value of `msc_reqbody_no_files_length` and `msc_reqbody_length` and which data length is added to `msc_reqbody_no_files_length`.

airween:
Hi @hnakamur,

I just could review the branch you mentioned (I want to try that too), but I caught a very good test:

https://github.com/hnakamur/ModSecurity/blob/eb4cdb631d0afcaabefe4874240e23524d4fcc34/tests/regression/misc/99-forcebodybuf.t

This highlights a new question: if there is no body processor, but a rule forces the request body variable (`REQUEST_BODY` appears and will be filled in v2 - unfortunately there is a know bug in v3, see #2045 and #2146), what is the expected behavior? (It's a theoretically question, and I assume we should consider both variable).

hnakamur:
When `ctl:forceRequestBodyVariable=On` is specified in a rule,
`msc_reqbody_no_files_length == msc_reqbody_length` is true.
https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L403-L406

Since `msc_reqbody_no_files_length == msc_reqbody_length` is also true for JSON, XML, and URLENCODED requests,
I think we can treat `ctl:forceRequestBodyVariable=On` case similarly as those cases, grouped as none multipart requests.

Now let's go back to the definitions of `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction`.
I think the expected behaviors are:

| `SecRequestBodyNoFilesLimitAction` | condition                                                   | processing                |
|------------------------------------|-------------------------------------------------------------|---------------------------|
| `Reject`                           | `msc_reqbody_no_files_length <= SecRequestBodyNoFilesLimit` | process whole body        |
| `Reject`                           | `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit`  | return status 413         |
| `ProcessPartial`                   | `msc_reqbody_no_files_length <= SecRequestBodyNoFilesLimit` | process whole body        |
| `ProcessPartial`                   | `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit`  | process partial body (*1) |

| `SecRequestBodyLimitAction`        | condition                                   | processing                |
|------------------------------------|---------------------------------------------|---------------------------|
| `Reject`                           | `msc_reqbody_length <= SecRequestBodyLimit` | process whole body        |
| `Reject`                           | `msc_reqbody_length > SecRequestBodyLimit`  | return status 413         |
| `ProcessPartial`                   | `msc_reqbody_length <= SecRequestBodyLimit` | process whole body        |
| `ProcessPartial`                   | `msc_reqbody_length > SecRequestBodyLimit`  | process partial body (*2) |

If we follow the definitions of `SecRequestBodyNoFilesLimitAction` and `SecRequestBodyLimitAction` one by one,
`(*1)` means process first `SecRequestBodyNoFilesLimit` bytes and
`(*2)` means process first `SecRequestBodyLimit` bytes.

However considering `SecRequestBodyNoFilesLimit` was added to prevent DoS attacks with large request bodies,
I think we might like to define both `(*1)` and `(*2)` mean process first `min(SecRequestBodyNoFilesLimit, SecRequestBodyLimit)` bytes. 

Let's combine the two tables above (I omit `SecRequestBody` and `msc_reqbody_` to save widths):

| `NoFilesLimitAction` | `LimitAction`    | condition                                            | processing                                          |
|----------------------|------------------|------------------------------------------------------|-----------------------------------------------------|
| `Reject`             | `Reject`         | `no_files_length <= NoFilesLimit && length <= Limit` | process whole body                                  |
| `Reject`             | `Reject`         | `no_files_length > NoFilesLimit \|\| length > Limit` | return status 413                                   |
| `ProcessPartial`     | `Reject`         | `no_files_length <= NoFilesLimit && length <= Limit` | process whole body                                  |
| `ProcessPartial`     | `Reject`         | `no_files_length > NoFilesLimit && length <= Limit`  | process first `NoFilesLimit` bytes                  |
| `ProcessPartial`     | `Reject`         | `length > Limit`                                     | return status 413                                   |
| `Reject`             | `ProcessPartial` | `no_files_length <= NoFilesLimit && length <= Limit` | process whole body                                  |
| `Reject`             | `ProcessPartial` | `no_files_length <= NoFilesLimit && length > Limit`  | process first `Limit` bytes                         |
| `Reject`             | `ProcessPartial` | `no_files_length > NoFilesLimit`                     | return status 413                                   |
| `ProcessPartial`     | `ProcessPartial` | `no_files_length <= NoFilesLimit && length <= Limit` | process whole body                                  |
| `ProcessPartial`     | `ProcessPartial` | `no_files_length > NoFilesLimit \|\| length > Limit` | process first `min(NoFilesLimit, Limit)` bytes (*3) |

(*3) for multipart, ignore parts after when one of `no_files_length > NoFilesLimit` or `length > Limit` becomes true.

The [modsecurity_request_body_end_raw](https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L547-L620) function is called from 
[modsecurity_request_body_end when `ctl:forceRequestBodyVariable=On` is set](https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L754-L757).
It is also [called from modsecurity_request_body_end_urlencoded](https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L632-L635).

My suggestion is to modify `modsecurity_request_body_end_raw` to set first `min(NoFilesLimit, Limit)` bytes of the request body to `msr->msc_reqbody_buffer` bytes instead of setting the whole body,
and modify the URLENCODED parser to support partial processing when the request body is cut off.

hnakamur:
Oh wait, I found the current implementation of the function `modsecurity_request_body_store` returns the status 413 when `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit` and `SecRequestBodyLimitAction` is set to `Reject`.
https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/msc_reqbody.c#L408-L424

On the other hand, `hook_request_late` and `read_request_body` return the status 413 when `msc_reqbody_length > SecRequestBodyLimit` and `SecRequestBodyLimitAction` is set to `Reject`.
https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/mod_security2.c#L1001-L1024

https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/apache2_io.c#L261-L289
https://github.com/owasp-modsecurity/ModSecurity/blob/7db03796f832eb702778223350e87eaee1c68c0d/apache2/apache2_io.c#L307-L334

So I think what is needed to fix the document of [SecRequestBodyLimitAction](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyLimitAction) to something like:
Description: Controls what happens once a request body limit (configured with SecRequestBodyLimit) or a request body limit excluding the content of multipart files (configured with SecRequestBodyNoFilesLimit) is encountered.

And we don't need to add `SecRequestBodyNoFilesLimitAction`.
What we need to modify the implementation is partial processing for the case when `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit || msc_reqbody_length > SecRequestBodyLimit` and `SecRequestBodyLimitAction` is `ProcessPartial`.

airween:
> So I think what is needed to fix the document of [SecRequestBodyLimitAction](https://github.com/owasp-modsecurity/ModSecurity/wiki/Reference-Manual-%28v2.x%29#user-content-SecRequestBodyLimitAction) to something like: Description: Controls what happens once a request body limit (configured with SecRequestBodyLimit) or a request body limit excluding the content of multipart files (configured with SecRequestBodyNoFilesLimit) is encountered.
> 
> And we don't need to add `SecRequestBodyNoFilesLimitAction`. What we need to modify the implementation is partial processing for the case when `msc_reqbody_no_files_length > SecRequestBodyNoFilesLimit || msc_reqbody_length > SecRequestBodyLimit` and `SecRequestBodyLimitAction` is `ProcessPartial`.

I think I agree with this - now what's your plan? Add some modifications to this PR, or close this one and open a completely new PR?

I think first you should create only for v2, with several new tests. If we agree that the result is good enough, then the v3 would be the next.

What do you think?

hnakamur:
Sorry for a late reply and thank you for your comment.
I would like to add modifications to this PR.
I just added more test cases for SecRequestBodyLimitAction.

I found segmentation faults occurs in `tests/regression/config/10-reqbody-limit-action-forcebodybuf.t`.
I am going to investigate the cause of the problem.

airween:
> Sorry for a late reply and thank you for your comment.

no worries!

> I would like to add modifications to this PR.

Okay, thank you.

hnakamur:
I added commits for partial processing of request body considering `SecRequestBodyNoFilesLength` as well as `SecRequestBodyLength`.
I also added more tests.

Basically, the values crossing the limit are not added to `ARGS` or `XML` variables.
For example, if the request body is `a=1&b=abcd` (content-length: 10) and `SecRequestBodyNoFilesLength 9` and `SecRequestBodyLimitAction ProcessPartial`,
the first 9 bytes (`a=1&b=abc`) is processed and only `a=1` is added to the `ARGS`.
If we add also `b=abc`, then `SecRule ARGS "abc$"` will match incorrectly, so we should not add it.

However, there is an exception with the XML request body processor.
If the text content is equal to or longer than 300 bytes, it will be added to `ARGS` or `XML` variables even when the content is crossing the limit.

When the partial processing is actually enabled (`SecRequestBodyLimitAction ProcessPartial` and the body length exceeds the limit),
`xmlParseChunk` is called from `xml_complete` with 0 as the `terminate` argument.
It was always 1 but is changed in the https://github.com/hnakamur/ModSecurity/commit/8ccfc6a6dea2ee2cdcf103c17a79db4d76139921 commit.
It is expected that the partial text content at the end is not added, however if the length of the content is equal to or longer than 300 bytes,
the content is processed and added by libxml2.
https://gitlab.gnome.org/GNOME/libxml2/-/blob/ca6a8cf94672e6f3c48c08d6af2201599788fc17/parser.c#L11061-11079

We could avoid this problem by parsing the XML twice.
Once for just to determine the length to the end of the last tag or content which fits in the limit,
Second for real parsing the cut-off request body.
However, I don't think this is viable option since it causes the longer latency of body processing.

airween:
Thanks @hnakamur - I try to review all pending PR's for v2 soon.

hnakamur:
I found a few bugs in my PR, and I'm currently fixing them.
Would you mind waiting until I've pushed the fixes before reviewing it?
Thanks!

airween:
> I found a few bugs in my PR, and I'm currently fixing them. Would you mind waiting until I've pushed the fixes before reviewing it? Thanks!

Sure, I have to review many other PR's for v2...

hnakamur:
I think I've fixed the bugs. Could you take another look when you have time? Thanks!

PR #3487, opened 2026-02-01
fix: cppcheck warnings in standalone/

umprayz:
## what

- This pull request fixes the `cppcheck` warnings in all `.c`, `.cpp` and `.h` files within the `standalone` directory.
- The `cppcheck` command has been connected into the workflow.

## why

- The changes were made based on the result of cppcheck command and the built in checks of the workflow.
- Previously, `cppcheck` could not find the necessary system and Apache headers, resulting in numerous "missing include" warnings and incorrect analysis.
- With these changes, running `cppcheck` on the standalone directory now produces zero warnings locally and in CI.

## references

- The `cppcheck` configuration logic in `Makefile.am` was based on the `v3/master` branch.

## notes

- I would like to continue this work on other directories too and make the whole source tree cppcheck clean.


umprayz:
Closing in favour of #3488, which is a cleaner version.

PR #3490, opened 2026-02-07
feat: add google benchmark

fzipi:
## what

- create a simple benchmark using google benchmark tooling

## why

- have a simple way to testing hypothesis on performance for specific rules based on requests

## todo

- include more examples?
- verify it can be easily integrated into the build
- only libmodsecurityv3, because it's easier to integrate as library


dune73:
I do not have time for this, but it looks really cool. Much needed.

PR #3494, opened 2026-02-12
refactor: autotools files

fzipi:
## what

- refactor autotools files
- add types definitions that can be used from `config.h`

## why

- simplify configure.ac logic


configure.ac
```diff
@@ -1,5 +1,6 @@
 # ModSecurity configure.ac
 
+AC_PREREQ([2.72])
```
airween: You bumped the required version to 2.72, therefore we got [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/21965640805/job/63454627588?pr=3494#step:6:14) error.

Ubuntu 22.04 [has](https://packages.ubuntu.com/jammy/autoconf) 2.71.

We can consider to upgrade the CI environment to Ubuntu 24.04, or use any version, but run CI in Docker, as we do it in case of mod_security2's [cppcheck](https://github.com/owasp-modsecurity/ModSecurity/pull/3488/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR143-R144) job.

airween (changes requested):

airween:
Thanks for this PR, it's nice to see that the library check methods are more simple now, that's awesome.

The only one thing I realized that the `_DISPLAY` variables are removed from the summary. For eg.

in case of YAJL the old and the new output at the end of the `./configure` script is this:
```
   + YAJL                                          ....found v2.1.0
      -lyajl,  -DWITH_YAJL -I/usr/include
```

but in case of CURL, this is the old one:
```
   + LibCURL                                       ....found v8.18.0 
      -lcurl,  -DWITH_CURL_SSLVERSION_TLSv1_2 -DWITH_CURL
```
and this is the new output:
```
   + LibCURL                                       ....found v8.18.0
      -lcurl, -I/usr/include/x86_64-linux-gnu -isystem /usr/include/mit-krb5 -isystem /usr/include/mit-krb5 -I/usr/include/p11-kit-1
```
so the macros for preprocessor  mostly disappeared.

I don't think it's important, but also don't know is could be useful for someone. What do you think?

fzipi:
I don't think people will base anything on the this specific output, other than "did this find my library or not?".

airween:
> I don't think people will base anything on the this specific output, other than "did this find my library or not?".

I didn't think about "did this", it's rather "what are the compilation macros". But well, if someone is on that level, probably they can figure out by themselves :smiley:.

Thank you again, merging into `v3`.

airween (approved):

Issue #3498, opened 2026-02-16
RFC: Let's kill mod_security2 for NGINX


fzipi:
## Description

Let's ☠️ mod_security2 for NGINX.

## Context

I don't know anyone that was brave/crazy enough to use mod_security2 for NGINX. The version is not feature complete, and the development was never finished.

## Alternatives

Of course, everyone should be using libmodsecurity3 + the nginx connector.

### 1. Build System Changes (configure.ac)

Remove or modify these sections:

- Lines 189-206: Remove the --enable-standalone-module option (or document that it no longer builds NGINX support)
- Line 300: Remove AC_CONFIG_FILES([doc/doxygen-nginx])
- Line 940: Remove AC_CONFIG_FILES([nginx/modsecurity/config])

### 2. Makefile Changes

👉 Makefile.am (top-level)

- Lines 38-39: Remove the test-regression-nginx target
- Line 62: Remove nginx/ from the cppcheck target

👉 tests/Makefile.am

- Lines 85-86: Remove the test-regression-nginx target

👉 doc/Makefile.am

- Lines 9-11: Remove the nginx target
- Line 23: Remove nginx from the all-local target dependencies
- Line 27: Remove nginx cleanup from clean-local

👉 standalone/Makefile.am

- Line 46: Remove or change -DVERSION_NGINX flag (note the FIXME comment on line 45 acknowledges this is misleading)

### 3. Directory and File Removal

👉 Delete the entire nginx/ directory tree:
```
nginx/
├── modsecurity/
│   ├── ngx_http_modsecurity.c
│   ├── apr_bucket_nginx.c
│   ├── apr_bucket_nginx.h
│   ├── ngx_pool_context.c
│   ├── ngx_pool_context.h
│   ├── config.in
│   └── config (generated)
└── TODO
```

👉 Delete test file:
- tests/run-regression-tests-nginx.pl

👉 Delete documentation config:
- doc/doxygen-nginx.in
- doc/doxygen-nginx (generated)

### 4. C Code Changes

Remove all VERSION_NGINX preprocessor conditionals:

👉 apache2/mod_security2.c

- Line 94: #if defined(WIN32) || defined(VERSION_NGINX) → Change to #if defined(WIN32)
- Line 238: #if !(defined(VERSION_IIS)) && !(defined(VERSION_NGINX)) && !(defined(VERSION_STANDALONE)) → Remove && !(defined(VERSION_NGINX))
[...]

PR #3502, opened 2026-02-19
refactor: remove NGINX support from v2 (#3498)

sanjib2006:
### files and folders deleted/modified other than mentioned directly in the issue: 
- standalone/server.c: Changed ap_server_argv0 value from "nginx" to "standalone".
- tests/regression/action/00-disruptive-actions.t: Removed NGINX-specific test expectations (nginx => ...).
- tests/regression/config/00-load-modsec.t: Removed NGINX-specific test expectations.
- tests/regression/nginx/: Deleted directory containing NGINX-specific regression test configurations 

## why

- The NGINX module for v2 is not feature-complete, and I removed the related files as mentioned in the issue

## references

- #3498


fzipi (commented):
For now it looks good. I'll grep to see if we need to cleanup additional stuff.

fzipi:
Maybe adding on top of CHANGES:
```
DD MMM YYYY - 2.9.13
--------------------

 * chore: removed support for nginx in v2.x
```

So it is documented there also.

sanjib2006:
These were commented out so I did not touch them in this commit
- Lines: 613-644, tests/regression/config/10-request-directives.t 
- Lines: 110-145, tests/regression/config/10-audit-directives.t

airween:
> So it is documented there also.

Just fyi: I don't expect any maintainers to fill up the CHANGELOG. Usually I can collect all changes from PR's. But as you want...

sanjib2006:
So should I edit the CHANGELOG??

airween:
> So should I edit the CHANGELOG??

You don't need to edit, but you can (through a regular commit) :smiley:.

As I wrote, I always collect all PR's and fill the changelog before releases.

sanjib2006:
ok I won't edit that :smile:

apache2/msc_release.h
```diff
@@ -52,16 +52,12 @@
 #ifdef	VERSION_IIS
```
airween: Here you should try to avoid the current nested macro chain and use a "flat" structure:
```suggestion
#if defined(VERSION_IIS)
    #define MODSEC_MODULE_NAME "ModSecurity for IIS (STABLE)"
#elif defined(VERSION_STANDALONE)
    #define MODSEC_MODULE_NAME "ModSecurity Standalone (STABLE)"
#else
    #define MODSEC_MODULE_NAME "ModSecurity for Apache"
#endif

#define MODSEC_MODULE_VERSION MODSEC_VERSION
#define MODSEC_MODULE_NAME_FULL MODSEC_MODULE_NAME "/" MODSEC_MODULE_VERSION " (http://www.modsecurity.org/)"
```

standalone/Makefile.am
```diff
@@ -53,7 +52,7 @@ standalone_la_CFLAGS = -DVERSION_NGINX \
     @MODSEC_EXTRA_CFLAGS@ \
     @PCRE_CFLAGS@ \
     @YAJL_CFLAGS@ \
-    @SSDEEP_CFLAGS@
+    @SSDEEP_CFLAGS@
```
airween, line 55: I'm wondering why GH mark this line as new...

The [current](https://github.com/owasp-modsecurity/ModSecurity/blob/9b836b652a269bde9a91bfbbbd2cf5c0c79d5a68/standalone/Makefile.am#L56) version already contains it.

Do you have any idea why is that here?

airween (changes requested):

airween:
@sanjib2006 thanks for this PR.

I've made a review, please take a look at my comments.

And here are some more staff that you might review too:
* please run command `grep -riI nginx doc/*` in your source root, and remove `nginx` patterns from all occurrences
* there are two tests (you mentioned them previously), in `tests/regression/config/10-request-directives.t` and `tests/regression/config/10-audit-directives.t`; just remove them, those are not relevant anymore

Thanks again!

standalone/Makefile.am
fzipi: Probably newline differences?

airween: Ah, thanks, you're right. @sanjib2006 added (correctly) a `\r` to the end of all lines.

airween (approved):
LGTM!

airween:
Thank you @sanjib2006!

PR #3503, opened 2026-02-20
fix: strlen related variables should be of type size_t

fzipi:
## what

- `strlen` returns a `size_t` variable, that depends on the architecture and environment

## why

- fix warnings is Windows
- it is the correct thing to do ™️ 

## references

- it appears as warnings in Windows related pipelines. See for example https://github.com/owasp-modsecurity/ModSecurity/pull/3502/changes#diff-bed0e89713f19f2d99baeb2645c2d8c15dc9172d400dcba1d844f73193d1775fL199


fzipi:
@copilot apply changes based on the comments in [this thread](https://github.com/owasp-modsecurity/ModSecurity/pull/3503#pullrequestreview-4156113027)

PR #3504, opened 2026-02-25
CI: Align workflow with configure.ac (recursive submodules, v2 Linux job, drop x32)

Easton97-Jens:
### Summary

This PR introduces a new separate CI workflow (`ci_new.yml`).
The existing workflow remains unchanged.

The primary goal is to align CI behavior with the expectations defined in configure.ac, while modernizing the Linux environment and simplifying architecture handling.

---

#### 1. Alignment with `configure.ac`

The build system assumes that required components provided via git submodules are fully and recursively initialized.

`configure.ac` explicitly documents this expectation (e.g., requiring recursive submodule initialization for bundled modules).

To ensure CI matches this requirement, the new workflow:

Upgrades actions/checkout from v4 to v6

Explicitly runs:
```
git submodule update --init --recursive
```

This guarantees:

* Deterministic submodule state
* Full availability of bundled modules
* Consistency with configure.ac
* Reduced risk for upcoming submodule updates (including the pending Mbed TLS update PR)

### This is the primary structural change introduced by this workflow.
---

#### 2. Integration of v2 Linux Static Analysis Flow

The workflow adds a dedicated Linux `cppcheck` job using a `debian:sid` container.

This structure is derived from the previous v2 CI logic and restores Linux-based static analysis in addition to the macOS job.

Benefits:

* Closer alignment with the Linux target environment
* Broader static analysis coverage
* Continuity with earlier CI design
---

#### 3. Removal of x32 (i386)

The new workflow drops 32-bit (x32/i386) builds:

Ubuntu upgraded from 22.04 to 24.04

Ubuntu 24.04 no longer provides native i386 runner support

Removes multilib setup and architecture-specific dependency branches

Simplifies the matrix and dependency logic

This reduces CI complexity and aligns the build matrix with current runner support.
---

### Additional Updates

macOS 14 → macOS 15

Windows 2022 → Windows 2025

Minor matrix cleanup and clearer job structure

---

### Rationale for Separate Workflow

The existing workflow is intentionally preserved to:

* Avoid disrupting the current CI baseline
* Allow isolated validation of structural changes
* Enable safe comparison before potential consolidation
* Reduce migration risk

---


airween:
Hi @Easton97-Jens,

many thanks for this PR.

The only one thing that I've done that a bit reformatted your description for the better visibility. Please take a look at that, I hope I put lines (lists, headings) to the right place.

If you think everything is okay, just let me know, or feel free to correct the description. I can merge this one soon.

Beside of that, I like this PR. If everything will be okay, we can remove the existing one later - what do you think?

Easton97-Jens:
You can go ahead and merge the PR.
I’ll leave it up to you when to remove the old CI workflow – we can keep it for a short observation period if you prefer.

.github/workflows/ci_new.yml
```diff
@@ -0,0 +1,295 @@
...
+          ./build.sh
+          ./configure
+
+      - name: Run cppcheck
+        run: make check-static
+
+  cppcheck-linux:
+    name: Static analysis (cppcheck, Linux, debian:sid)
+    runs-on: ubuntu-latest
+    container: debian:sid
```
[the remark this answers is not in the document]
Easton97-Jens: Good idea, I've implemented it, although it then differs from v2/master.

.github/workflows/ci_new.yml
```diff
@@ -0,0 +1,295 @@
...
+
+    steps:
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Ensure submodules are fully initialized (recursive)
+        run: git submodule update --init --recursive
+
```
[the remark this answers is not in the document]
Easton97-Jens: Good idea, I checked the behavior and it behaves the same as the other one.

.github/workflows/ci_new.yml
```diff
@@ -0,0 +1,295 @@
...
+            libtool \
+            pkg-config \
+            cppcheck \
+            libyajl-dev \
+            libcurl4-openssl-dev \
+            liblmdb-dev \
+            liblua5.2-dev \
+            libmaxminddb-dev \
+            libpcre2-dev \
+            libgeoip-dev \
```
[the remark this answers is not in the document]
Easton97-Jens: It's removed, he's right about that, I also completely removed it on macOS.

.github/workflows/ci_new.yml
```diff
@@ -0,0 +1,295 @@
...
+        run: |
+          sudo apt-get update -y -qq
+          sudo apt-get install -y \
+            libyajl-dev \
+            libcurl4-openssl-dev \
+            liblmdb-dev \
+            liblua5.2-dev \
+            libmaxminddb-dev \
+            libpcre2-dev \
+            libpcre3-dev \
```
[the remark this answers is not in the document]
Easton97-Jens: is removed

.github/workflows/ci_new.yml
```diff
@@ -243,24 +216,22 @@ jobs:
   cppcheck-linux:
     name: Static analysis (cppcheck, Linux, debian:sid)
     runs-on: ubuntu-latest
-    container: debian:sid
+    container: debian:trixie
```
airween: `trixie` is the current stable release of Debian (Debian 13). It contains 2.17.1, see the [packages](https://packages.debian.org/trixie/cppcheck) site.

Sid contains [2.19.0](https://packages.debian.org/sid/cppcheck), and here that's the point: we definitely want to use the last `cppcheck` version, as soon as possible.

I saw Copilot suggestion, but I'm afraid in this case you shouldn't follow it :).

.github/workflows/ci_new.yml
```diff
@@ -97,17 +91,13 @@ jobs:
           - { label: "without ssdeep",  opt: "--without-ssdeep" }
           - { label: "with lmdb",       opt: "--with-lmdb" }
           - { label: "with pcre2 (default)", opt: "" }
-          - { label: "with pcre",       opt: "--with-pcre" }
```
airween: You removed the old PCRE library's check, but I think while the used OS supports PCRE3, we must check that too.

It seems you use [ubuntu24.04](https://github.com/Easton97-Jens/ModSecurity/blob/f9d884ce030e8107f1ee1500a546d65142a61d64/.github/workflows/ci_new.yml#L13), which still contains [pcre3](https://packages.ubuntu.com/noble/libpcre3), so please keep this case too.

(Unfortunately we don't have any information about how many uses uses the old and the new PCRE lib)

.github/workflows/ci_new.yml
```diff
@@ -278,7 +249,6 @@ jobs:
             liblua5.2-dev \
             libmaxminddb-dev \
             libpcre2-dev \
-            libgeoip-dev \
```
airween: `libgeoip-dev` was removed, but there is not a replacement, but a new library which can be used: [maxminddb](https://packages.debian.org/trixie/libmaxminddb-dev). I see that's also installed (two lines above), but for some reason the `configure` script does not recongise it. You can see what happens during the job (check the output, for eg. like [this](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/22770912452/job/66080347106?pr=3504#step:4:266) one).

Also, if you want to use `LMDB`, then you should pass that to `configure` script: `--with-lmdb`. Beside of that, you could try to pass `--with-maxmind` too.

airween (changes requested):
Please see my comments.

.github/workflows/ci_new.yml
Easton97-Jens: ci: add missing dependencies for cppcheck static analysis

Added the required dependencies so cppcheck can properly detect
and analyze libraries such as maxminddb and lmdb during the
static analysis job.

.github/workflows/ci_new.yml
Easton97-Jens: Good point, thanks for the clarification.

I originally switched to "trixie" because it is the current stable Debian release, but you're right that "sid" provides the newer cppcheck version (2.19). Since the goal of this job is to run static analysis with the most recent cppcheck, it makes sense to keep using "debian:sid".

I'll update the workflow accordingly.

.github/workflows/ci_new.yml
Easton97-Jens: Good point, thank you for the clarification.

The old PCRE case was removed while simplifying the CI matrix. Since Ubuntu 24.04 still provides PCRE3, it makes sense to keep this configuration covered by the tests.

The "--with-pcre" configuration option has now been restored to ensure both PCRE and PCRE2 setups remain tested. The same change has also been applied to the macOS workflow.

airween:
@Easton97-Jens,

just a quick and interesting note: look at those two `cppcheck` jobs.

[Here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/22790837280/job/66135568725?pr=3504#step:3:6) is the failed one's [result](https://productionresultssa4.blob.core.windows.net/actions-results/1cc6c50f-c303-4c2f-882f-73d0cc6cc517/workflow-job-run-9f82dc9b-2349-508a-b2f3-89156679e340/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-03-08T16%3A35%3A49Z&sig=CTuux1dBkTEDMkdxg9R479GSUjrg4%2FL%2FUkY%2Bn%2F%2FcS28%3D&ske=2026-03-08T18%3A54%3A31Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-03-08T14%3A54%3A31Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-03-08T16%3A25%3A44Z&sv=2025-11-05) which uses `cppcheck-2.20.0`
```
2026-03-07T11:20:25.3147910Z warning: src/variables/variable.h,157,style,functionConst,Unless it overrides a base class member, the member function 'modsecurity::variables::KeyExclusions::toOmit' can be made a const function. Making this function 'const' should not cause compiler errors. Even though the function can be made const function technically it may not make sense conceptually. Think about your design and the task of the function first - is it a function that must not change object internal state?
```

And [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/22790837269/job/66135568726?pr=3504#step:2:9) is the success one which uses `cppcheck-2.19.0`. This version of `cppcheck` can't find the issue that the newer found.

This is why it's important to use the latest release.

(Please do not misunderstand this, I know this issue did not caused by your PR)

Easton97-Jens:
Thanks for the clarification! Do you think the PR is good to merge now, or is there anything else I should adjust?

.github/workflows/ci_new.yml
```diff
@@ -0,0 +1,295 @@
...
+          apt-get install -y git
+
+      - uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          submodules: true
+
+      - name: Mark repo as safe for git
+        run: git config --global --add safe.directory $GITHUB_WORKSPACE
+
```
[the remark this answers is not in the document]
airween: @Easton97-Jens I think this makes sense.

airween (changes requested):

airween:
> Thanks for the clarification! Do you think the PR is good to merge now, or is there anything else I should adjust?

There is only one thing as I see - please check the reviews.

Easton97-Jens:
Thanks again for the review! The change has already been included in my latest update and is part of the commits from March 6, 2026. I also updated the submodules checkout configuration. Could you please take another look at the file, especially around lines 230–237?

airween (approved):

airween:
> Thanks again for the review! The change has already been included in my latest update and is part of the commits from March 6, 2026. I also updated the submodules checkout configuration. Could you please take another look at the file, especially around lines 230–237?

Sorry, I just checked the modified file and there Copilot still mentioned that.

Thank you for this PR, I'm going to merge it.

PR #3510, opened 2026-03-18
Fix memory leak

JonathanBerrew:
This is a Marc Stern modification, I don't have much more insight on the code he made. To be reviewed with caution and check if this is still relevant

### DIFF
--- a/apache2/apache2_config.c
+++ b/apache2/apache2_config.c
@@ -34,7 +34,7 @@
 # define ap_log_perror(a, b, c, d, fmt, ...) fprintf(stderr, fmt "\n", __VA_ARGS__)
 #endif
 
-const char* id_log(msre_rule* rule, apr_pool_t* pool) {
+const char* id_log(const msre_rule* rule, apr_pool_t* pool) {
 	assert(rule != NULL);
 	assert(rule->actionset != NULL);
 	assert(rule->ruleset != NULL);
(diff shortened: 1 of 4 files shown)


airween:
Hi @JonathanBerrew,

thanks for this PR.

Please take a look at the CI tests. It looks like none of them could produce a runnable code. For eg. [here](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/23236608104/job/67542963250?pr=3510#step:7:105) you can see the reason:
```
Error: msc_util.c:2730:13: error: conflicting types for 'id_log'; have 'const char *(msre_rule *)'
 2730 | const char* id_log(msre_rule* rule) {
      |             ^~~~~~
In file included from modsecurity.h:40,
                 from msc_util.c:15:
msc_util.h:169:13: note: previous declaration of 'id_log' with type 'const char *(msre_rule *, apr_pool_t *)'
  169 | const char* id_log(msre_rule* rule, apr_pool_t* pool);
      |             ^~~~~~
```

Please before you send a PR, try to build the module and make some tests in a clean environment.

After it built and you restarted the server with the new module, please send a few request there and try to emulate some attack which must be caught by the WAF and you should see them in logs.

Also after the build process you should run these commands:
```
make test
make test-regression
make check-static
```
Note, that the last static analysis does not cover the whole source tree, but for eg. if you add some code to `IIS/` directory, it can be useful.

Also note, that `make test-regression` takes a bit more time (it depends on your HW, but usually it can be 4-5 mins).

JonathanBerrew:
I indeed forgot to remove the id_log from msc_util.c
It does build on my side and it has been tested for long time now since we're using this change Marc made more than over a year.
For the test, I'm not sure how to fix that... My knowledge of c/c++ isn't that good sadly

Issue #2807, opened 2022-10-05
Bug: Premature EOF in JSON leads to status 500 instead of REQBODY_ERROR


dune73:
**Describe the bug**

Instead of failing safely, ModSec triggers a status 500.

```
$ cat example.json
{ "id" : "123"
$ curl -v http://localhost -H "Content-Type: application/json" -d @example.json
...
< HTTP/1.1 500 Internal Server Error
...
```

**Logs and dumps**

```
[2022-10-05 11:17:58.926606] [-:error] 127.0.0.1:39976 Yz1Lxuf_TBfAeTSTcu5oFwAAAAk [client 127.0.0.1] ModSecurity: JSON parser error: parse error: premature EOF\n [hostname "localhost"] [uri "/"] [unique_id "Yz1Lxuf_TBfAeTSTcu5oFwAAAAk"]
[2022-10-05 11:17:58.926624] [-:error] 127.0.0.1:39976 Yz1Lxuf_TBfAeTSTcu5oFwAAAAk [client 127.0.0.1] ModSecurity: JSON parser error: parse error: premature EOF\n [hostname "localhost"] [uri "/"] [unique_id "Yz1Lxuf_TBfAeTSTcu5oFwAAAAk"]
[2022-10-05 11:17:58.926629] [core:trace3] 127.0.0.1:39976 Yz1Lxuf_TBfAeTSTcu5oFwAAAAk fixups hook gave 500: /
```

**Server (please complete the following information):**
 - ModSecurity version (and connector): ModSec 2.9.6 
 - WebServer: Apache 2.5.54

**Additional information:**

I played around with broken JSON payloads following a customer request. This was the only case where I could make ModSecurity stumble.

martinhsv:
Hi @dune73 ,

At first glance I would probably agree that it makes more sense for this use case to signal REQBODY_ERROR instead of triggering a status code of 500. Indeed, that is the behaviour in ModSecurity v3.

A slight clarification: with ```SecRuleEngine On```, the outcome is as you describe. But with ```DetectionOnly``` REQBODY_ERROR does get set (with appropriate rule 200002 output in the audit log).

[...]

PR #3515, opened 2026-03-18
Return 400 in case of parsing error

JonathanBerrew:
This is a Marc Stern modification, I don't have much more insight on the code he made. To be reviewed with caution and check if this is still relevant


JonathanBerrew:
https://github.com/owasp-modsecurity/ModSecurity/issues/2807
When Marc made this modification, he mentioned this issue

PR #3522, opened 2026-03-27
Update SQLi/XSS operators for libinjection v4.0.0

Easton97-Jens:
## what

- Updated the SQLi and XSS detection operators to support the newer `libinjection` return codes (`injection_result_t`).
- Added explicit handling for `TRUE`, `FALSE`, and `ERROR` results from `libinjection_sqli` and `libinjection_xss`.
- Treated `LIBINJECTION_RESULT_ERROR` as a fail-safe match to avoid missing potentially malicious input.
- Preserved diagnostic context by storing raw input in `TX.0` when `capture` is enabled, even on parser errors.
- Maintained previous behavior for positive detections (fingerprints for SQLi, input storage for XSS).
- Added regression tests for benign inputs to ensure no false positives are introduced.

## why

- Newer versions of `libinjection` introduced `injection_result_t`, requiring explicit handling in ModSecurity operators.
- Without this update, parser errors could be handled inconsistently and weaken fail-safe detection behavior.
- Treating parser errors as matches ensures conservative and security-focused handling when input cannot be reliably parsed.
- Preserving captured input during error cases improves debugging and visibility for rule authors.
- Regression tests ensure the updated logic does not introduce false positives for safe inputs.

## references

- Related to adapting ModSecurity to newer `libinjection` APIs.
- Adds regression coverage for parser error handling and safe input validation.

### DIFF
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -59,10 +59,10 @@ typedef struct Rules_t RulesSet;
       if (m_rules && m_rules->m_debugLog && m_rules->m_debugLog->m_debugLevel >= b) { \
           m_rules->debug(b, m_id, m_uri, c); \
       } \
-  } while (0)
+  } while (0);
 #else
 #define ms_dbg(b, c) \
-  do { } while (0)
+  do { } while (0);
 #endif
 
 #ifndef NO_LOGS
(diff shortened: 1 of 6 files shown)


airween:
Hi @Easton97-Jens,

thanks for this PR - could you take a look at the failed tests on macOS and also the SonarCloud reports?

Easton97-Jens:
Hi,

thanks for pointing that out!

I had a look at the failed macOS tests and the SonarCloud report. The Lua-related error does not seem to be caused by the changes in this PR, but rather appears to be a general issue in the project. Therefore, I’d suggest addressing it separately from this PR.

Let me know if you’d still like me to investigate it further here.

Best regards

fzipi:
The macos build should be fixed by #3524. Once it is merged, you can rebase.

Easton97-Jens:
can be checked and tested

Issue #3523, opened 2026-03-27
Build failure with Lua 5.5.0: LUA_ERRGCMM undeclared


Easton97-Jens:
Description:
I noticed a build failure when compiling ModSecurity with Lua 5.5.0 on macOS.
I'm not entirely sure where the issue originates, but the build fails during compilation.

Error message:

engine/lua.cc:158:18: error: use of undeclared identifier 'LUA_ERRGCMM'
case LUA_ERRGCMM:
     ^

Observation:
It seems related to LUA_ERRGCMM not being defined.
However, I'm not sure whether this is due to changes in Lua (e.g. version 5.5.0) or an issue in the code.

Expected behavior:
The build should either succeed or clearly indicate if a specific Lua version is unsupported.

Actual behavior:
The build fails during compilation.

Steps to reproduce:

Install Lua 5.5.0
Build ModSecurity with Lua support enabled
Run the build
Compilation fails in engine/lua.cc

Additional context:
I noticed this in CI (macOS job with Lua). Other jobs do not seem to be affected.

airween:
Hi @Easton97-Jens,

seems like this is a duplicated issue, see
* #2406
* #2403
* #2389

I try to solve the workflow and force using Lua < 5.5.

airween:
#3524 will be merged soon and (partially) will solve this issue.

Please note that this is just a workaround, do not close this issue until we don't fix the root cause (namely align the code to Lua 5.5).

airween:
@Easton97-Jens please pull down the modifications from upstream's v3/master branch.

PR #3524, opened 2026-03-27
fix(ci): pin Lua version on 5.4

airween:
## what

This PR pins Lua version in OSX workflows on 5.4.

## why

It seems like the default Lua version is 5.5 if someone installs it on OSX, and the library is not ready to work with it.

This is a workaround, we should align the Lua related code.

## references

#3523

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -99,13 +99,14 @@ jobs:
                        libtool \
                        yajl \
                        lmdb \
-                       lua \
+                       lua@5.4 \
                        libmaxminddb \
                        libxml2 \
                        ssdeep \
                        pcre \
                        bison \
                        flex
+          echo "/usr/local/opt/lua@5.4/bin" >> $GITHUB_PATH
       - uses: actions/checkout@v4
         with:
           submodules: true
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,6 +4,9 @@ on:
   push:
   pull_request:
 
+env:
+  LUA_VERSION: "5.4"
+
 jobs:
   build-linux:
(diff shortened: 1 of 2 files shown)


fzipi (commented):
Two suggestions:

## 1. Use a workflow-level `env` variable

The string `5.4` is hardcoded in 6 places across the two files. If we ever bump the version, all of them need updating. Each workflow should define it once at the top:

```yaml
env:
  LUA_VERSION: "5.4"
```

Then reference it consistently everywhere — macOS and Linux alike:

```yaml
# macOS
brew install ... lua@${{ env.LUA_VERSION }} ...
echo "/usr/local/opt/lua@${{ env.LUA_VERSION }}/bin" >> $GITHUB_PATH

# Linux
sudo apt-get install -y liblua${{ env.LUA_VERSION }}-dev
```

## 2. Lua version mismatch across platforms

After this PR, macOS jobs will use Lua **5.4** while all Linux jobs still use `liblua5.2-dev` (Lua **5.2**):

| File | Job | Lua version |
|------|-----|-------------|
| `ci.yml` | `build-linux` | 5.2 (`liblua5.2-dev`) |
| `ci.yml` | `build-macos` | 5.4 (after this PR) |
| `ci_new.yml` | `build-linux` | 5.2 (`liblua5.2-dev`) |
| `ci_new.yml` | `build-macos` | 5.4 (after this PR) |
| `ci_new.yml` | `cppcheck` (macOS) | 5.4 (after this PR) |
| `ci_new.yml` | `cppcheck-linux` | 5.2 (`liblua5.2-dev`) |

This makes CI *more* inconsistent than before — we'd be testing against different Lua versions per platform. The Linux jobs should also use `liblua${{ env.LUA_VERSION }}-dev` so there's a single source of truth for the Lua version across all platforms and jobs.

fzipi (approved):
Perfect! 🎉

airween:
Issue #3523 was fixed in #3531, closing this one.

PR #3529, opened 2026-03-29
ci: update workflows (checkout v6, recursive submodules), prep for mbedTLS v4, Windows fixes

Easton97-Jens:
### **What**

* Based on the approach from PR #3524, but implemented with adjusted behavior to better fit the current CI structure
* Reworked and unified CI workflows so that all jobs run consistently
* Removed GeoIP-related steps from CI to reduce complexity
* Updated `actions/checkout` from v4 to v6
* Switched submodule handling to `recursive`
* Added dynamic Lua version detection on Linux (no more manual version maintenance required)
* Improved CI configuration as preparation for mbedTLS v4 compatibility
* Fixed Windows workflow so tests now pass successfully

---

### **Why**

* PR #3524 introduces important Lua version pinning changes, which are taken into account here but implemented in a more flexible way
* The previous CI setup was not fully aligned and harder to maintain
* Dynamic Lua detection reduces maintenance effort and avoids hardcoded versions
* Removing GeoIP simplifies CI and reduces potential failure points
* Unified workflows improve maintainability and consistency
* `actions/checkout@v6` is the current recommended version
* Recursive submodules ensure all dependencies are properly fetched
* Changes serve as a preparation step towards mbedTLS v4 compatibility
* Windows CI was failing previously and is now fixed

---

### **Notes**

* This PR incorporates the core ideas from #3524 but with a different structure and behavior
* After merging this PR, #3524 may no longer be needed in its current form
* Feedback on whether #3524 should be closed would be appreciated

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,6 @@ on:
   push:
   pull_request:
   
-env:
-  LUA_VERSION: "5.4"
-
 jobs:
   build-linux:
     name: Linux (${{ matrix.platform.label }}, ${{ matrix.compiler.label }}, ${{ matrix.configure.label }})
@@ -37,6 +34,45 @@ jobs:
           - platform: {label: "x32"}
             configure: {label: "wo ssdeep"}
     steps:
+      - name: Detect latest Lua dev package
+        id: detect_lua
+        shell: bash
+        run: |
+          set -euo pipefail
+
+          sudo dpkg --add-architecture ${{ matrix.platform.arch }}
+          sudo apt-get update -y -qq
+
+          CANDIDATES="$(apt-cache search '^liblua[0-9]+\.[0-9]+-dev$' | awk '{print $1}')"
+
+          if [ -z "$CANDIDATES" ]; then
+            echo "No libluaX.Y-dev package found"
+            exit 1
+          fi
+
+          VALID_PACKAGES=""
+          for pkg in $CANDIDATES; do
+            if apt-cache show "${pkg}:${{ matrix.platform.arch }}" >/dev/null 2>&1; then
+              VALID_PACKAGES="${VALID_PACKAGES}${pkg}"$'\n'
+            fi
+          done
+
+          if [ -z "$VALID_PACKAGES" ]; then
+            echo "No Lua package available for architecture ${{ matrix.platform.arch }}"
+            exit 1
+          fi
+
+          BEST_PKG="$(
+            printf '%s\n' "$VALID_PACKAGES" \
+            | sed -E 's/^liblua([0-9]+\.[0-9]+)-dev$/\1 &/' \
+            | sort -V \
+            | tail -n1 \
+            | awk '{print $2}'
+          )"
+
+          echo "lua_pkg=$BEST_PKG" >> "$GITHUB_OUTPUT"
+          echo "Using $BEST_PKG for ${{ matrix.platform.arch }}"
+
       - name: Setup Dependencies (common)
         run: |
           sudo dpkg --add-architecture ${{ matrix.platform.arch }}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -3,12 +3,16 @@ name: Quality Assurance
 on:
   push:
   pull_request:
+  
+env:
+  LUA_VERSION: "5.4"
```
airween: Please remove all hardcoded Lua versions from all workflow. Let the system to choose the right version. I think it's a good feedback if we need to align that in the code (and the build system).

airween (changes requested):

.github/workflows/ci.yml
airween: Note here, we already fixed the Lua version issue in #3525.

Easton97-Jens: Should the Lua version be selected dynamically by the system going forward, or do we want to keep a fixed version (at least for certain environments like Linux)?

airween: I think we should let the system to choose the right version. Usually, macOS uses the recent releases, which is good, because we can get notification about that.

If we use a fixed version, then we can forget to update the code.

Easton97-Jens: macOS now follows the previous behavior (using the default system setup), while Linux has been updated to dynamically detect and use the currently available Lua version.

This should give us more flexibility on Linux while keeping macOS stable and aligned with its typical environment.

airween:
@Easton97-Jens,

please consider to approve Copilot suggestions'.

I think none of them are exactly true (especially [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3529#discussion_r3007933639) one: `brew install python3` is likely to fail; use `brew install python`), because all tests were passed, but I think those are useful, and you should accept them. But let's discuss, if you have any question.

Easton97-Jens:
Thanks for the suggestion! This is not strictly required, as Python 3 is already available in the environment, so no additional installation is needed. Therefore, I have removed the installation entirely.

.github/workflows/ci_new.yml
```diff
@@ -31,33 +31,57 @@ jobs:
           - { label: "without geoip",   opt: "--without-geoip" }
           - { label: "without ssdeep",  opt: "--without-ssdeep" }
           - { label: "with lmdb",       opt: "--with-lmdb" }
-          - { label: "with pcre2 (default)", opt: "" }
-          - { label: "with pcre",        opt: "--with-pcre" }
+          - { label: "with pcre2 (default)", opt: "--with-pcre2" }
```
airween: Please rollback these changes. There was some issue with `--with-pcre2` configure option, and that's the default now, but we still need the old PCRE (just for the case). And until Ubuntu supports the old PCRE library (24.04 [contains](https://packages.ubuntu.com/noble/libpcre3) it), I suggest to use that.

```suggestion
          - { label: "with pcre2 (default)", opt: "" }
          - { label: "with pcre",        opt: "--with-pcre" }
```

.github/workflows/ci_new.yml
```diff
@@ -115,7 +140,7 @@ jobs:
                        ssdeep \
                        pcre \
                        bison \
-                       flex
+                       flex
```
airween: Please remove trailing whitespaces.
```suggestion
                       flex
```

.github/workflows/ci_new.yml
```diff
@@ -206,7 +231,7 @@ jobs:
 
       - name: Install cppcheck
         run: |
-          brew install autoconf automake libtool cppcheck libmaxminddb yajl lua lmdb ssdeep
+          brew install autoconf automake libtool cppcheck libmaxminddb yajl lua lmdb ssdeep
```
airween: Please remove trailing whitespaces.
```suggestion
          brew install autoconf automake libtool cppcheck libmaxminddb yajl lua lmdb ssdeep
```

.github/workflows/ci_new.yml
```diff
@@ -234,11 +259,47 @@ jobs:
...
+          echo "lua_dev_pkg=$BEST_PKG" >> "$GITHUB_OUTPUT"
+          echo "lua_pkg=$LUA_PKG" >> "$GITHUB_OUTPUT"
+
+          echo "Using dev package: $BEST_PKG"
+          echo "Using interpreter: $LUA_PKG"
 
       - name: Install dependencies (v2 style)
         run: |
-          apt-get update
           apt-get install -y \
```
airween, line 305: Why did you remove `apt-get update`? I don't believe it was unnecessary.

airween (changes requested):

.github/workflows/ci_new.yml
Easton97-Jens: This is done when determining the Lua version, which is why it's omitted at this point.

airween: Ah, I see - sorry, and thanks.

airween (approved):

airween:
Thanks @Easton97-Jens, merging now.

Issue #3308, opened 2024-11-26
Drop YAJL dependency


mikelolasagasti:
`yajl` library has been unmaintained upstream[1] since 2015. Last published release contians multiple CVEs (CVE-2023-33460, CVE-2022-24795, CVE-2017-16516)  and fixes have had to be carried by downstream distributions and third parties. This is not an ideal situation. While a fork exists[2], it is unclear how widely adopted or blessed it is downstream.

As the maintainer of `libmodsecurity` for Fedora[3] and EPEL, I recently found that the `yajl` library will not be shipped with RHEL 10. This means a new maintainer would be required to add it to EPEL. The previous maintainer for RHEL & EPEL has recommended moving away from `yajl`[4].

Currently, libmodsecurity can be built without `yajl`, but I understand that making it mandatory is considered desirable, as per #3144 and #3151.

Given the security concerns and the upstream status of `yajl`, I recommend opening a discussion on dropping `yajl` as a dependency and exploring alternative JSON libraries, such as `JSON-C`, which is actively maintained and more widely adopted.

[1] https://github.com/lloyd/yajl
[2] https://github.com/robohack/yajl/
[3] https://src.fedoraproject.org/rpms/libmodsecurity
[4] https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/message/YPFHPOKAND3RZR7ZKWTDHUQEESG6IUJ3/

airween:
Agree, we should change.

What do you think about the "official" YAML [library](https://github.com/yaml/libyaml)?

airween:
Note, I added label [2.x] too, because mod_security2 is also affected.

dune73:
This does not sound good. Thank you for bringing this to the attention @mikelolasagasti.

berrange:
> What do you think about the "official" YAML [library](https://github.com/yaml/libyaml)?

I'm the Fedora `yajl` maintainer quoted in the link above. While YAML may be a superset of JSON, I'd not recommend using a YAML parser instead of a native JSON parser. It'll perform worse, and potentially open the code up to attacks that are impossible with pure JSON (aka the "Billion laughs attack" in YAML context). I can vouch for using `json-c` as an alternative to `yajl`, as that's what was successfully adopted in projects I work on.  Be slightly cautious about `jansson` - when we previously tried it, we found it was not able to cope with the full numeric range of the `uint64` type - if that's important to modsecurity's needs.

airween:
[...]

PR #3539, opened 2026-04-08
Prototype: Evaluate alternatives to YAJL for JSON processing

Easton97-Jens:
### Description

This pull request introduces a **prototype** exploring the potential **replacement of YAJL** or the use of alternative JSON libraries.

See related discussion in the issue:
👉 [https://github.com/owasp-modsecurity/ModSecurity/issues/3308](https://github.com/owasp-modsecurity/ModSecurity/issues/3308)

The goal is to evaluate the current JSON handling and investigate possible improvements in terms of:

* maintainability
* performance
* security
* modernization of dependencies

As part of this prototype, the following JSON libraries are being considered:

* RAPIDJSON
* NLOHMANNJSON
* JSONC (json-c)
* JANSSON
* CJSON
* JSONCPP
* JSONCONS
* SIMDJSON
* YYJSON
* GLAZE

This implementation should be considered **experimental** and primarily serves as a starting point for discussion.

### Notes

* This is **not a final solution**, but an initial draft
* API/behavior may change
* Focus is on feasibility and early insights

### Feedback

I’m explicitly open to:

* suggestions for improvement
* preferred libraries from the list
* compatibility constraints within the project
* performance/security considerations
* general feedback on the direction

Thanks for taking the time to review 🙌


airween:
@Easton97-Jens,

first of all, thanks for this PR.

Unfortunately for some reason (probably because you completely re-organized the test cases) no regression tests were ran. That's very bad.

The other think is that I don't think we need as many different JSON parser as you've added here.

Here is my idea about this topic (add more JSON parser):
* under #3308 I mentioned the [jsonbench](https://github.com/digitalwave/jsonbench) tool
* I think we should extend that tool with more JSON parsers
* **important**: we should implement the possible solutions for both engine (v2 and v3) and make the tool available to choose the necessary one (for a user)
* then we (not just you and me, but the community) should start testing them thoroughly with so many different input
* we need to gather experience, like
  * performance
  * availability (how can we ensure that users can use any parser? As an operation system package? through a git submodule?)
  * knowledge, like
    * stream parsing (SAX) - this is mandatory
    * DOM parsing - this is necessary for tests, but we can replace it
    * **this is a new expectation**, but it might be I mentioned somewhere else: If the parser could perform schema validation, we could create a new operator: `@validateJsonSchema`

Based on these, we could choose 2 or 3 new JSON parsers, and that would be enough to add as a solution.

I know this is a hard work and not trivial, but I'm sure this would be a very big step forward in both engines.

Easton97-Jens:
Thanks for your feedback. I'm working on the second prototype where I'm implementing two JSON modules: jsoncons and simdjson, as they're more closely related to Sax. Rapidjson is omitted because the last release is too old. You can find it at https://github.com/Easton97-Jens/ModSecurity branch v3/master-json2 see. It's not finished yet.

airween:
> Thanks for your feedback. I'm working on the second prototype where I'm implementing two JSON modules: jsoncons and simdjson, as they're more closely related to Sax. Rapidjson is omitted because the last release is too old. You can find it at https://github.com/Easton97-Jens/ModSecurity branch v3/master-json2 see. It's not finished yet.

Thanks.

Could you add them first to [jsonbench](https://github.com/digitalwave/jsonbench)?

Easton97-Jens:
I opened a PR
https://github.com/digitalwave/jsonbench/pull/1

PR #3544, opened 2026-04-12
fix: heap buffer overflow in acmp pm

airween:
## what

This PR fixes a  possible heap buffer overflow in ACMP (Aho Corasick) pattern matching function.

## why

There is a bug report, received in email from @fumfel and his team. Also they provided this fix.

## references

The original report:
```
Root Cause
----------

The function acmp_add_pattern() receives a pattern and its length as
separate parameters: (const char *pattern, size_t length). However, the
code uses strlen(pattern) instead of length to determine buffer sizes.

When pattern contains embedded null bytes (which is valid -- the length
parameter accounts for them), strlen() returns the offset of the first
null byte, which is shorter than the actual length. The buffer allocated
is too small, and subsequent operations (the for loop copying text, and
strcpy copying pattern) write past the end of the heap allocation.
```

## other notes

The bug can only be exploited if the admin puts a `\0` character into an argument of any `@pm` (or similar) operator.

### DIFF
--- a/src/utils/acmp.cc
+++ b/src/utils/acmp.cc
@@ -387,7 +387,7 @@ if (parser->is_active != 0) return -1;
             child->pattern = (char *)"";
             child->letter = letter;
             child->depth = i;
-            child->text = (char *)calloc(1, length + 2);
+            child->text = (char *)calloc(1, i + 2);
             /* ENH: Check alloc succeded */
             for (j = 0; j <= i; j++) child->text[j] = pattern[j];
         }
@@ -395,7 +395,7 @@ if (parser->is_active != 0) return -1;
             if (child->is_last == 0) {
                 parser->dict_count++;
                 child->is_last = 1;
-                child->pattern = (char *)calloc(1, length + 2);
+                child->pattern = (char *)calloc(1, length + 1);
                 /* ENH: Check alloc succeded */
                 memcpy(child->pattern, pattern, length);
                 child->pattern[length] = '\0';
--- a/src/utils/acmp.cc
+++ b/src/utils/acmp.cc
@@ -389,7 +389,9 @@ if (parser->is_active != 0) return -1;
             child->depth = i;
             child->text = (char *)calloc(1, i + 2);
             /* ENH: Check alloc succeded */
-            for (j = 0; j <= i; j++) child->text[j] = pattern[j];
+            for (j = 0; j <= i; j++) {
+                child->text[j] = pattern[j];
+            }
         }
         if (i == length - 1) {
             if (child->is_last == 0) {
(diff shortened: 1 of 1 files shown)


src/utils/acmp.cc
```diff
@@ -387,17 +387,18 @@ if (parser->is_active != 0) return -1;
             child->pattern = (char *)"";
             child->letter = letter;
             child->depth = i;
-            child->text = (char *)calloc(1, strlen(pattern) + 2);
+            child->text = (char *)calloc(1, i + 2);
             /* ENH: Check alloc succeded */
             for (j = 0; j <= i; j++) child->text[j] = pattern[j];
```
theseion:
```suggestion
            for (j = 0; j <= i; j++) {
                child->text[j] = pattern[j];
            }
```

src/utils/acmp.cc
```diff
@@ -387,17 +387,18 @@ if (parser->is_active != 0) return -1;
             child->pattern = (char *)"";
             child->letter = letter;
             child->depth = i;
-            child->text = (char *)calloc(1, strlen(pattern) + 2);
+            child->text = (char *)calloc(1, i + 2);
```
theseion, line 390: What's the `+ 2` for? The terminating null byte only requires one slot...

src/utils/acmp.cc
theseion: Why not use `memcpy` here as well?

theseion (changes requested):

src/utils/acmp.cc
airween: The code was implemented as you can see in this form. I just have some thoughts why is the +2:
* `i` means the depth of the pattern, so based on `i`'s length it needs `i+1` byte to store, and +1 for the `\0`
* other safety reason

But I don't want to touch this if it's not necessary.

src/utils/acmp.cc
airween: This is also a legacy code, if it was not necessary I wouldn't change that. But here is some idea why did they use this form:
* you should see the "big picture" in that implementation, eg. see the variable `long letter` in line [379](https://github.com/airween/ModSecurity/blob/66c338e1fce5a6294bdd790aa1b7312a9f4bd19c/src/utils/acmp.cc#L379)
* `child->text` is a `char *`
* if we used `memcpy` then it would make a binary copy
* probably this is also a safety solution to force the type cast

theseion (commented):
As this issue doesn't appear to depend on special compiler flags, could we add a test?

airween:
> As this issue doesn't appear to depend on special compiler flags, could we add a test?

Unfortunately I was not able. @fzipi also tried to help me with this, but all tests were passed always...

theseion (approved):

PR #3551, opened 2026-04-22
Add C API to retrieve matched rules from a transaction

meirdev:
## references

Required for: https://github.com/owasp-modsecurity/ModSecurity-nginx/pull/374

### DIFF
--- a/src/transaction.cc
+++ b/src/transaction.cc
@@ -28,7 +28,6 @@
 #include <fstream>
 #include <iomanip>
 #include <iostream>
-#include <iterator>
 #include <set>
 #include <unordered_map>
 #include <vector>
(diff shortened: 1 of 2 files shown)


src/transaction.cc
```diff
@@ -2325,5 +2326,47 @@ extern "C" int msc_set_request_hostname(Transaction *transaction,
...
+ *
+ * @param transaction ModSecurity transaction.
+ * @param index       Zero-based index into the matched rule records.
+ *
+ * @returns The rule id at the given index.
+ * @retval >0 Rule id of the matched rule at the given index.
+ * @retval  0 Index is out of range.
+ *
+ */
+extern "C" int64_t msc_get_matched_rule_id(Transaction *transaction,
```
airween, line 2360: This function returns only with one rule id, right?

In case of CRS, if it uses anomaly scoring model, this will return always with rule 949110, or similar ID (depends on intervention's phase).

Does it make sense?

meirdev: Not exactly - the function returns a single rule ID, but I'm using `msc_get_matched_rules_count` to iterate through *all* matched rules.

Example of log from my [tests](https://github.com/meirdev/modsec-dashboard):

```
["2026-04-22T06:44:46+00:00","172.20.0.1","36770","","178","localhost","0.000","POST","/etc/nginx/html/","","HTTP/1.1","http","1df077358d8536cd96324e994abf6256","","403","308","text/html","curl/8.5.0","","","1", "920273,941100,941110,941160,941390,941320,942131,942370,942431,942511,942432,949110"]
```

airween:
Tests on Windows were failed, could you check them?

airween:
@meirdev - thanks for the changes. There are three new issues in Sonarcloud, could you take a look at them?

Also, please pick up the recent changes from v3/master.

airween:
@meirdev: and also please take a look at the Copilot suggestions.

meirdev:
> There are three new issues in Sonarcloud, could you take a look at them?

Regarding the two remaining issues in SonarCloud: all the other `extern "C"` functions are inside the namespace, and I don't want to write these two differently.

airween:
> Regarding the two remaining issues in SonarCloud: all the other `extern "C"` functions are inside the namespace, and I don't want to write these two differently.

Ah, you're totally right - thank you.

airween (approved):

airween:
@meirdev thank you, merging now.

PR #3553, opened 2026-04-26
Add is_interrupted flag in the audit JSON log

meirdev:
## what

Add an `is_interrupted` flag to the audit JSON log.

## why

From the audit logs, it's not possible to determine whether ModSecurity actually interfered with the request.

Neither `producer.secrules_engine` nor `messages.*.details.ruleId` reliably indicate this, since rule actions can be modified dynamically.

For example:

```
SecRule REMOTE_ADDR "@streq 1.1.1.1" \
    "id:1000,\
    phase:1,\
    log,\
    msg:'test'"

SecRuleUpdateActionById 1000 "deny,status:403"
```

## references

Coraza has introduced a similar flag:
https://github.com/corazawaf/coraza/blob/8b28b72ec16d01ed5b9b664699d1f69386f67029/internal/auditlog/auditlog.go#L89

### DIFF
--- a/headers/modsecurity/transaction.h
+++ b/headers/modsecurity/transaction.h
@@ -594,6 +594,14 @@ class Transaction : public TransactionAnchoredVariables, public TransactionSecMa
     ModSecurityIntervention m_it;
 
     /**
+     * Sticky flag: set to true the first time the connector consumes a
+     * disruptive intervention via Transaction::intervention(). Unlike
+     * m_it.disruptive, this is NOT cleared by intervention::reset(), so
+     * it remains a reliable signal at audit-log time.
+     */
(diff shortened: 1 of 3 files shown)


airween:
Hi @meirdev,

thanks for this PR - I think it's a great idea.

I've been thinking about where could we put this information in `Native` format, but unfortunately many parts are unimplemented yet in case of v3, so we need to skip to add this info there.

Anyway, we should take it back again if someone wants to extend the auditlog parts.

airween:
@meirdev: could you pick up the recent changes from upstream's `v3/master`?

airween (approved):

airween:
@meirdev - thank you, merging into v3/master.

PR #3555, opened 2026-05-04
Add indicator to negative operator

### LINKED ISSUE #3554: Log message should indicate if operator is negative
The rule:

```
SecRule ARGS:var "!@streq hello" "id:100,phase:1,deny,log"
```

Example of denied request:

```
curl http://localhost:8080/?var=hi
```

Log message:

```
Access denied with code 403 (phase 1). Matched "Operator `StrEq' with parameter `hello' against variable `ARGS:var' (Value: `hi' )
```

Reading the log message alone is confusing, why does `@StrEq hello` have a match for `hi`?

---

Also, I think the use of backticks (`) and single quotes (') in the log message should be fixed.

meirdev:
Fix #3554 

In the issue, we mention that we can't change the backticks (`) and single quotes (') in the message to avoid breaking log parsers. Could this change also cause problems?

### DIFF
--- a/src/operators/operator.cc
+++ b/src/operators/operator.cc
@@ -109,8 +109,9 @@ std::string Operator::resolveMatchMessage(Transaction *t,
     std::string ret = m_match_message;
 
     if (ret.empty() == true) {
+        const std::string opName = (m_negation ? "!" : "") + m_op;
         if (m_couldContainsMacro == false) {
-            ret = "Matched \"Operator `" + m_op + "' with parameter `" +
+            ret = "Matched \"Operator `" + opName + "' with parameter `" +
                 utils::string::limitTo(200, m_param) +
                 "' against variable `" + utils::string::toHexIfNeeded(key) + "' (Value: `" +
                 utils::string::limitTo(100,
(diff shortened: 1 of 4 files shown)


airween:
Hi @meirdev,

thanks for this PR - seems the code is good, but cppcheck complains about this new part:

```
2026-05-04T07:08:26.5186100Z warning: src/rule_with_operator.cc,230,style,unreadVariable,Variable 'opName' is assigned a value that is never used.
```

probably because of [this](https://github.com/meirdev/ModSecurity/blob/ac89a7bdbc1b726edfc83a22c57a3f1f908028ad/src/rule_with_operator.cc#L230C23-L230C29) new variable.

Feel free to add an inline suppression to that line.

dune73:
Could we get an example of a log message before and one after the change?

meirdev:
Rule:

```
SecRule REQUEST_URI "!@contains /hello" "id:1,phase:1,deny,status:403,log,msg:'blocked'"
```

Request:

```
curl http://localhost/admin
```

Before:

```
Matched "Operator `Contains' with parameter `/hello' against variable `REQUEST_URI' (Value: `/admin' )
```

After:

```
Matched "Operator `!Contains' with parameter `/hello' against variable `REQUEST_URI' (Value: `/admin' )
```

We add `!` before the operator name

dune73:
Got you. Thanks. I think this makes sense.

dune73:
I think ModSec2 is much clearer than this with the "required" clause in the logfile. But this is probably even better.

Issue #3082, opened 2024-02-14
Segmentation fault during collection cleanup on possibly corrupted dbm data


twouters:
**Describe the bug**

Apache segfaults during processing of requests when the persistent collection is being processed to remove stale records.

**Logs and dumps**

Note: IP addresses have been obfuscated

```
(gdb) bt full
#0  0x00007236e618c456 in ?? () from /lib/x86_64-linux-gnu/libc.so.6
No symbol table info available.
#1  0x00007236e65bb2f8 in memcpy (__len=4294967295, __src=0x7236bc01deae, __dest=<optimized out>)
    at /usr/include/x86_64-linux-gnu/bits/string_fortified.h:34
No locals.
#2  apr_pstrmemdup (a=<optimized out>, s=s@entry=0x7236bc01deae "\017\061\070\061.123.123.12", 
    n=4294967295) at strings/apr_strings.c:107
        res = <optimized out>
#3  0x00007236e5fb4664 in collection_unpack (msr=msr@entry=0x7236bc020eb0, blob=0x7236bc01de76 "7", 
    blob_size=228, log_vars=log_vars@entry=0) at persist_dbm.c:71
        var = 0x7236bc0b2e60
        col = 0x7236bc0b2a38
        blob_offset = 56
#4  0x00007236e5fb5c96 in collections_remove_stale (msr=msr@entry=0x7236bc020eb0, 
    col_name=0x7236bc009020 "ip") at persist_dbm.c:781
        col = 0x0
        var = 0x0
        dbm_filename = 0x7236bc02d578 "/var/cache/httpd/www-ip"
        key = {dptr = 0x7236bc030ea0 "127.12.12.123", dsize = 14}
        value = {dptr = 0x7236bc01de76 "7", dsize = 228}
        dbm = <optimized out>
        rc = <optimized out>
        keys_arr = 0x7236bc02d700
        keys = 0x7236bc031ae8
        now = <optimized out>
        i = 314
        userinfo = 0x7236bc02d570 "www"
[snip]
```
**To Reproduce**

Not sure, I guess: have corrupt persistent collection data files (I made backups, if you're interested) and perform some specific requests to trigger a segfault (don't known exactly when it triggers).

**Expected behavior**

Webserver doesn't crash

**Server (please complete the following information):**
 - ModSecurity version: 2.9.7
 - WebServer: Apache 2.4.57
 - OS (and distro): Debian bullseye/bookworm

**Additional context**

[...]

PR #3560, opened 2026-05-10
Re-apply lost fix 4f33f5b: collection_unpack value_len underflow regression

g4mm4-VCF:
## Summary

Re-applies commit 4f33f5b ("Fix possible segfault in collection_unpack",
@twouters, 2024-03-01) which was silently dropped during merge 649aea7
on 2024-04-04, and adds one further bound check that the review
process surfaced. The drop has been confirmed by:

- `git show 4f33f5b:apache2/persist_dbm.c | sed -n '60,75p'` — has the guards
- `git show 649aea7:apache2/persist_dbm.c | sed -n '60,75p'` — guards gone
- `git show v2.9.13:apache2/persist_dbm.c | sed -n '60,75p'` — still gone

Affected releases: v2.9.10, v2.9.11, v2.9.12, v2.9.13.

## What this PR does

`collection_unpack()` in `apache2/persist_dbm.c` had two missing
bound checks that this PR restores / adds:

1. **Re-introduce the `var->name_len < 1 ||` and
   `var->value_len < 1 ||` guards** from 4f33f5b. Without them, a
   blob containing `0x00 0x00` in the `value_len` field underflows
   `value_len - 1` to `0xFFFFFFFF` and `apr_pstrmemdup()` is invoked
   with size = 4 GiB.

2. **Strengthen the name-bound check from
   `blob_offset + name_len > blob_size` to
   `blob_offset + name_len + 2 > blob_size`** to also cover the
   2-byte `value_len` header read site (caught during Copilot
   review on this PR). The previous shape allowed an OOB read on a
   truncated blob whose tail landed exactly at the end of the name
   body.

Both changes apply to the same function and are exercised by the same
crash primitive (a malformed SDBM entry under SecDataDir), so they
are landed together.

The final diff against `v2/master` is four lines in
`apache2/persist_dbm.c` plus one explanatory comment — no additional
files, no new dependencies, no behaviour change for well-formed
blobs.

## Test plan

- [x] standalone harness (replicating the `collection_unpack`
      arithmetic byte-for-byte) confirms `copy_len = UINT_MAX` on
      the buggy code path and `rc = 0` (NULL return) on the patched
      code path for a `value_len = 0` blob
- [ ] **manual**: feed an 8-byte SDBM blob (header + `name_len = 1`
      + `name = ""` + `value_len = 0`) into a `SecDataDir`-backed
      collection; observe the previously-crashing worker now ignore
      the malformed entry
- [ ] **manual (truncated-blob path, Copilot #1)**: feed a blob
      whose tail lands exactly at `blob_offset + name_len`;
      observe the patched code returns `NULL` instead of reading
      `value_len` past the buffer

The standalone unit-test harness is **not** included in-tree — the
existing `tests/regression/` is a Perl HTTP integration harness that
doesn't fit a unit test of a `static` function, and wiring a new
`check_PROGRAMS` into `tests/Makefile.am` (where `msc_test` lives)
would require restructuring out of scope for this fix. The harness
is preserved in the disclosure PoC archive for anyone who wants to
verify offline.

## References

- Original issue:  https://github.com/owasp-modsecurity/ModSecurity/issues/3082
- Original fix:    https://github.com/owasp-modsecurity/ModSecurity/commit/4f33f5b
- Reverting merge: https://github.com/owasp-modsecurity/ModSecurity/commit/649aea7
- Affected tag:    https://github.com/owasp-modsecurity/ModSecurity/releases/tag/v2.9.13

Coordinated with @airween via private security channel.


airween:
Hi @g4mm4-VCF,

thanks for this patch.

Please take a look at the SonarCLoud report and fix those issues.

g4mm4-VCF:
Hi there,

I have fixed all the issues, and everything is now working properly.
~g4mm4

airween:
> Hi there, I have fixed all the issues, and everything is now working properly.

Thanks - now please take a look at Copilot suggestions. I think all of them make sense.

g4mm4-VCF:
Thanks for running Copilot. Pushed f239b02e to address all 5 suggestions. Summary:

**#1 (`apache2/persist_dbm.c:72`) — fixed.** Copilot is right, this is a real adjacent bug. After consuming the name field, `blob_offset` can land exactly at `blob_size`, and the 16-bit read of `value_len` from `blob[blob_offset]` / `blob[blob_offset+1]` would OOB-read on a truncated blob. Added `if (blob_offset + 1 >= blob_size) return NULL;` before the read. Same defensive pattern as the existing checks for `name_len`.

**#2–#5 (`tests/regression/persist_dbm/`) — dropped.** Copilot was right that the directory wasn't wired into Autotools (no `AC_CONFIG_FILES`, no parent `SUBDIRS`), and the sanitizer flags weren't portable. Looking at the existing test layout (`tests/regression/` is the Perl HTTP regression harness, `tests/` is the `msc_test` binary that links all the `apache2/` sources), a unit test of a `static` function inside `persist_dbm.c` doesn't slot cleanly into either: it's not an HTTP-level test for the Perl harness, and adding a separate `check_PROGRAMS` to `tests/Makefile.am` would require non-trivial restructuring of the existing pattern.

The cleanest move was to drop the in-tree test directory entirely. The standalone test harness is preserved outside the upstream tree (in the disclosure PoC archive) for anyone who wants to verify the fix offline:

```c
// snippet from the standalone test
unsigned char blob[] = {
    0x49, 0x52, 0x01,   // SDBM header
    0x00, 0x01,         // name_len = 1
    0x00,               // name = ""
    0x00, 0x00,         // value_len = 0  ← BUG TRIGGER
};
// vulnerable build → apr_pstrmemdup(..., 0xFFFFFFFF) → SIGSEGV
// patched build    → second bound check returns NULL early
```

Happy to revisit the in-tree test if you'd prefer a Perl-level integration test that drops a malformed SDBM file into `SecDataDir` and asserts no segfault on retrieve. That'd be more work but fits your existing harness cleanly. Let me know.

Diff vs `upstream/v2/master` is now just the three checks in `apache2/persist_dbm.c`:

- the original `name_len < 1` and `value_len < 1` guards from 4f33f5b, plus
- the new `blob_offset + 1 >= blob_size` check that Copilot caught.

CI / SonarCloud should pass on the new commit. Ping me if anything else surfaces.

g4mm4-VCF:
SonarCloud caught a follow-up: cognitive complexity tipped to 26/25 because of the new bound check. Force-pushed `fbdd35a4` that folds Copilot's catch into the existing name-bound check instead of adding a new line:

```diff
- if (blob_offset + var->name_len > blob_size) return NULL;
+ /* Need name_len bytes for the name body plus 2 more for the value_len header. */
+ if (var->name_len < 1 || blob_offset + var->name_len + 2 > blob_size) return NULL;
```

The strengthened predicate now requires room for *both* the name body *and* the 2-byte `value_len` header. That's a strict superset of the original check, covers the truncated-blob OOB read Copilot flagged, and adds zero new branches → cognitive complexity stays at the previous baseline.

Final diff vs `upstream/v2/master` is now exactly three semantic changes + one comment line in `apache2/persist_dbm.c`:

1. `name_len < 1 ||` — the original 4f33f5b guard that 649aea7 dropped
2. `+ 2` on the upper bound — covers the new `value_len`-header read site
3. `value_len < 1 ||` — the matching 4f33f5b guard for the value path

CI/SonarCloud should be green again on `fbdd35a4`.

g4mm4-VCF:
Good catch — the PR description was stale after I dropped `tests/regression/persist_dbm/` in `fbdd35a4`. Just updated the body to match the actual contents of the branch:

- removed the "adds a standalone regression test" claim
- removed the `cd tests/regression/persist_dbm && make check` checkbox from the test plan
- documented why the in-tree test was dropped (Perl harness doesn't fit a unit test of a `static` function; wiring a new `check_PROGRAMS` is out of scope for a security fix) and pointed at the standalone harness preserved in the disclosure PoC archive
- added the new Copilot-flagged manual test step (truncated blob whose tail lands at `blob_offset + name_len`)
- expanded the "what this PR does" to call out the strengthened bound check explicitly

The branch itself is unchanged; only the description was out of sync.

g4mm4-VCF:
Two follow-ups from the latest Copilot pass, both on `apache2/persist_dbm.c`:

**Copilot — "`var->name_len < 1` is unreachable" — agreed, fixed in `2d2c0b0a`.**

Verified by walking the control flow:

| line | condition | exit |
|---|---|---|
| 38 | reads `var->name_len = (blob<<8)+blob` — value range 0..65535 |
| 39 | `if (var->name_len == 0)` | `break` (exits `while`) |
| 52 | `else if (var->name_len > 65536)` | `break` (mathematically unreachable, but harmless) |
| 65 | reachable iff `var->name_len ∈ [1, 65536]` | — |

So `var->name_len < 1` at line 65 is provably unreachable. Removed and replaced with a comment documenting that `name_len == 0` is handled upstream by the early break:

```c
blob_offset += 2;
/* Need name_len bytes for the name body plus 2 more for the value_len header.
 * name_len == 0 is already handled by the early break above, so no zero-check
 * is required at this point. */
if (blob_offset + var->name_len + 2 > blob_size) return NULL;
```

The `value_len < 1` clause on the value-side check is left in — there is no earlier zero-check for `value_len`, so that one is the live guard.

The original `4f33f5b` patch by @twouters added the `name_len < 1` clause for symmetry with the value side; my removal here deviates from that, but the dead-code observation is correct and the asymmetry is now documented in the comment. Happy to revert to the symmetric form if you'd prefer to mirror the original commit exactly.

---

**Copilot — "the new bound check is not covered by an in-tree regression test" — push-back, requesting maintainer guidance.**

I want to flag the framework limitation explicitly so we can decide together rather than play whack-a-mole with Copilot.

The fix lives inside `collection_unpack()`, which is `static` in `apache2/persist_dbm.c`. The existing v2 test surfaces are:

1. **`tests/regression/`** — Perl-driven HTTP integration harness. Tests rule execution against a running Apache; it doesn't have a slot for invoking internal `static` functions directly. A test in this format would have to: (a) drop a malformed SDBM file into `SecDataDir` from a fixture, (b) issue an HTTP request that triggers `collection_retrieve_ex`, (c) assert the worker doesn't crash. Possible but requires new fixture machinery that doesn't exist today.

2. **`tests/msc_test.c`** — JSON-driven harness for `tfn` / `op` / `action` plugin types only. There's no slot for invoking persistent-collection deserialisation, and `collection_unpack` would need to be exposed (made non-static or wrapped) before it could be reached from this harness.

Three options I see, in increasing scope:

- **(A)** Land the security fix without an in-tree test. Standalone harness is preserved in the disclosure PoC archive; CI keeps relying on the existing Perl regression suite for end-to-end coverage. **Smallest patch.**
- **(B)** Add a `tests/regression/persist_dbm/*.t` Perl test that drops a fixture SDBM file containing the 8-byte malicious blob into a temporary `SecDataDir`, then issues a request that resolves a persistent collection and asserts no segfault. This is the most idiomatic fit with the existing v2 framework. ~50 LoC, no API changes. **Recommended if you want in-tree coverage.**
- **(C)** Expose `collection_unpack` (drop `static` or add a `__attribute__((visibility))` wrapper) and add unit-test cases inside `tests/msc_test.c`. Strictest coverage but adds a public-ish symbol and requires extending the JSON harness. **Largest patch.**

Default position is (A), which is what the current branch reflects. If you'd prefer (B), I can put together the Perl fixture in a follow-up commit on this branch — happy to do the work, just want to confirm scope before adding test machinery that the harness doesn't currently have. (C) feels disproportionate for a 2-line bound-check fix.

Let me know which way you'd like to go.

g4mm4-VCF:
@airween — SonarCloud passed on the latest commit. All Copilot suggestions addressed. Ready to merge when you have time.

airween:
@g4mm4-VCF - thanks again, and sorry for late reply. Could you update your branch with the recent added commits?

g4mm4-VCF:
@airween No problem at all — thanks for getting back to it!

Done: I've rebased the branch onto the current `v2/master` (now at #3593 / `d27c982`), so it now includes all the recent commits. The rebase was clean — none of the new upstream commits touch `apache2/persist_dbm.c`, so there were no conflicts, and the net change is still the same isolated `+5/-2` guard fix in `collection_unpack()`.

The branch is up to date and green on SonarCloud. Ready for another look / merge whenever you have time.

airween:
> @airween No problem at all — thanks for getting back to it!


thank you - and sorry, meanwhile there was a new PR (that's important for newest CRS), could you update again?

> Done: I've rebased the branch onto the current `v2/master` (now at #3593 / `d27c982`), so it now includes all the recent commits. The rebase was clean — none of the new upstream commits touch `apache2/persist_dbm.c`, so there were no conflicts, and the net change is still the same isolated `+5/-2` guard fix in `collection_unpack()`.

thanks!
 
> The branch is up to date and green on SonarCloud. Ready for another look / merge whenever you have time.

Wait for the update, then if everything is okay, I'll merge this one too - new release is soon, that will include this PR too. Thank you!

g4mm4-VCF:
@airween Updated — thanks for the heads up on #3592, that's the right one to have in for CRS.

Two things in this push:

1. **Rebased onto the current `v2/master` (`df082e7`)**, so the branch now includes #3592 (`ctl:ruleRemoveTargetByTag` XML-attribute fix) and #3578. Clean rebase — none of the new commits touch `apache2/persist_dbm.c`, so no conflicts.

2. **Addressed Copilot's new overflow note.** It correctly pointed out that both guards used `blob_offset + len > blob_size`, and since `blob_offset`, `name_len`, `value_len` and `blob_size` are all `unsigned int`, that sum could wrap on a hostile `blob_size`. I rewrote both as remaining-byte comparisons (`len > blob_size - blob_offset`). `blob_offset <= blob_size` is guaranteed at each check by the loop guard, so the subtraction can't underflow, and `name_len` is capped at 65536 so `name_len + 2` can't overflow. I verified exhaustively that the new form is identical to the old one for every `offset <= size` in range — it only differs on the wraparound case the old form mishandled, so no behaviour change for real blobs.

On the remaining Copilot note about an in-tree Perl regression under `SecDataDir`: as discussed earlier the standalone harness lives in the advisory PoC archive; happy to wire up a `.pag`/`.dir` corruption case in `tests/regression/` if you'd like it in-tree, just say the word.

SonarCloud should re-run on the new commit. Ready to merge whenever it's green. Thanks!

airween:
Thanks @g4mm4-VCF!

airween (approved):

PR #3562, opened 2026-05-11
Exlude .gitignore from release tarballs

### LINKED ISSUE #3559: Exclude .gitignore files from release tarballs
We import upstream releases in our own git repository, much like what happens with the debian packages on salsa.debian.org

Just noticed that we were missing some files due to `.gitignore` being shipped in the release files, which I don't think is actually needed.

twouters:
The `.gitignore` files should not be included in the release tarballs because they hide required files and can conflict with existing `.gitignore` files when the tarball is extracted into a git repository.

Closes: https://github.com/owasp-modsecurity/ModSecurity/issues/3559
Fixes: 04f70099 ("Adds a simple release script")

### DIFF
--- a/build/release.sh
+++ b/build/release.sh
@@ -11,7 +11,7 @@ MY_DIR=${PWD##*/}
 ./build.sh
 
 cd ..
-tar --transform "s/^$MY_DIR/$DIR_NAME/" -cvzf $TAR_NAME --exclude .git $MY_DIR
+tar --transform "s/^$MY_DIR/$DIR_NAME/" -cvzf $TAR_NAME --exclude .git --exclude .gitignore $MY_DIR
 
 sha256sum $TAR_NAME > $TAR_NAME.sha256
 gpg --detach-sign -a $TAR_NAME


build/release.sh
```diff
@@ -11,7 +11,7 @@ MY_DIR=${PWD##*/}
 ./build.sh
 
 cd ..
-tar --transform "s/^$MY_DIR/$DIR_NAME/" -cvzf $TAR_NAME --exclude .git $MY_DIR
+tar --transform "s/^$MY_DIR/$DIR_NAME/" -cvzf $TAR_NAME --exclude .git --exclude .gitignore $MY_DIR
```
airween, line 14: Shouldn't we ignore only `libinjection/.gitignore`?

twouters: I don't think so. The `/.gitignore` file belongs to *this* git repository and it has no real meaning outside of it.

The same applies to every `.gitignore` file scattered around the repository and its submodules.

If someone wants to maintain their own `.gitignore` file in their own repository (clones excluded) they should be able to without these conflicting with it.

Ideally, they'd even have `export-ignore` gitattributes to not be included in git generated archives.

twouters:
> ❌ The last analysis has failed.
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=owasp-modsecurity_ModSecurity&pullRequest=3562)

Besides the apparently opinionated nature of the PR, I don't understand what failed here :)

airween:
> Besides the apparently opinionated nature of the PR, I don't understand what failed here :)

You might try:
```
git commit --allow-empty -m "Trigger SonarCloud"
git push
```

airween:
@twouters,

thanks - would you mind to check Copilot's suggestions?

Anyway, let me check the effect of this modification on Salsa's repository.

build/release.sh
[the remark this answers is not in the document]
twouters: I noticed this too, but it's not relevant to the proposed change and can be a fixed in separate PR

PR #3566, opened 2026-05-14
Fix lexer rejection of '@' in ctl:ruleRemoveTarget actions (#3565)

### LINKED ISSUE #3565: ctl:ruleRemoveTarget* rejects any target containing `@` (e.g. XPath attribute selectors)
## Summary

The action parser refuses to load a rule when the value of `ctl:ruleRemoveTargetByTag` (or `ctl:ruleRemoveTargetById`) contains a `@` character. The same rule loads cleanly under ModSecurity v2.

The cause appears to be in the scanner: `@` is tokenized as the operator prefix (as in `@rx`, `@detectXSS`) and that interpretation leaks into action-value tokenization. Likely in [`src/parser/seclang-scanner.ll`](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/parser/seclang-scanner.ll#L460).

## Affected

- libmodsecurity 3.0.15 (`owasp/modsecurity-crs:nginx` image, current pinned digest)
- ModSecurity-nginx connector

## Reproducer

```apache
SecRule REQUEST_URI "@unconditionalMatch" \
    "id:999990,\
    phase:2,\
    pass,\
    nolog,\
    ctl:ruleRemoveTargetByTag=attack-xss; XML://@*"
```

## Expected

Rule loads. `XML://@*` is a valid ModSecurity XPath target and is accepted as a `SecRule` variable in the same engine.

## Actual

```
[emerg] "modsecurity_rules_file" directive Rules error.
Line: 6. Column: 47. Expecting an action, got:  @*"
```

The error column always points at the `@`.

## Bisect 
| Value passed to `ctl:ruleRemoveTargetByTag` | Result |
|---|---|
| `attack-xss; XML:/*` | ✓ loads |
| `attack-xss; XML://foo` | ✓ loads |
| `attack-xss; XML://node` | ✓ loads |
| `attack-xss; XML:/element` | ✓ loads |
| `attack-xss; REQUEST_HEADERS:Foo` | ✓ loads |
| `attack-xss; XML://@*` | ✗ fails at `@` |
| `attack-xss; XML://@foo` | ✗ fails at `@` |
| `attack-xss; XML://@` | ✗ fails at `@` |
| `attack-xss; ARGS:@foo` | ✗ fails at `@` |
| `attack-xss; FOO@BAR` | ✗ fails at `@` |

So the problematic character is `@`, regardless of its position or what follows it. Quoting (`'…'`) and backslash escaping (`\@`) are not honored.

## ModSecurity v2 comparison

The same rule loads and runs correctly under Apache + `mod_security2` 2.9.x.

## Impact

`ctl:ruleRemoveTargetByTag` / `ctl:ruleRemoveTargetById` cannot be used to runtime-toggle any target that includes a `@` — which includes the entire XPath attribute axis (`XML://@*`, `XML://@name`, etc.). The only workaround is `ctl:ruleRemoveById`, which disables the whole rule rather than a single target.

## Asks

1. Allow `@` (and ideally the full target/XPath grammar) inside `ctl:ruleRemoveTarget*` values.
2. If a scanner change is non-trivial, document the limitation in the v3 reference manual.

Jitterx69:
## what

- Expands the `REMOVE_RULE_BY` regular expression macro in the Flex scanner (`src/parser/seclang-scanner.ll`) to include characters essential for XPath selectors.
- Specifically adds `@`, `=`, `(`, `)`, and `'` to the allowed character class.

## why

- Previously, the lexer rejected rule targets containing the `@` character (such as `XML://@*` or `ARGS:@foo`) because the macro lacked these characters.
- This premature termination of the match caused the scanner to fall back and generate the syntax error `Expecting an action, got: @*`.
- By explicitly including these characters, the lexer can now accurately tokenize XPath attribute axes and predicates without inadvertently consuming action list delimiters (such as commas, spaces, or double-quotes).
- This restores parity with ModSecurity v2 functionality, enabling granular rule exclusions on specific XML attributes.

## references

- closes #3565


airween:
@Jitterx69,

thanks - please add test cases to prove the correct behavior of this patch.

See regression tests [this](https://github.com/owasp-modsecurity/ModSecurity/blob/9b49187452b973009a1e51d62ee0a5b91c07dd94/test/test-cases/regression/action-ctl_rule_remove_target_by_id.json) or [this one](https://github.com/owasp-modsecurity/ModSecurity/blob/9b49187452b973009a1e51d62ee0a5b91c07dd94/test/test-cases/regression/action-ctl_rule_remove_target_by_tag.json).

airween:
Thanks - sorry, I forgot to mention that you should regenerate Bison related files. For this, please install the latest Bison and Flex versions, run `./configure` with your options, plus add `--enable-parser-generation`. Then you will see that there will be new files under `src/parser` with `.cc` and with `.hh` extensions - you must add them too to the commit.

Without this, the new tests will be failed.

You can find a similar solution [here](https://github.com/owasp-modsecurity/ModSecurity/pull/3558#issuecomment-4391333501).

Jitterx69:
> here

Yeah...I just realized that before you qoute...I will imply as you say.

src/parser/seclang-scanner.ll
```diff
@@ -457,7 +457,7 @@ FREE_TEXT_COMMA_QUOTE                   [^,\"\\n\\r]+
 NEW_LINE_FREE_TEXT                      [^, \t\"\n\r]+
 NOT !
 FREE_TEXT                               ([^\"]|([^\\]\\\"))+
-REMOVE_RULE_BY                          [0-9A-Za-z_\/\.\-\*\:\;\]\[\$]+
+REMOVE_RULE_BY                          [0-9A-Za-z_\/\.\-\*\:\;\]\[\$@\=\(\)\']+
```
theseion: Which valid target specs do the additional characters match? Please add tests for all of them.

theseion (changes requested):

src/parser/seclang-scanner.ll
Jitterx69: I have added more comprehensive regression tests for the additional characters in test/test-cases/regression/.

Rationale for characters =, (, ), and ': These are primarily required to support XPath predicates and functions in XML:// targets, which were supported in ModSecurity v2 but resulted in parser syntax errors in v3. Specifically:

- ( and ): Required for XPath functions like local-name() or namespace-uri(). They are also used in regex-based target exclusions (e.g., ARGS:/foo(bar)?/).
- = and ': Required for XPath predicates and string literals (e.g., XML://*[@id='123']).

Regarding SonarCloud Failures in seclang-scanner.cc: The current failures and the large diff in the generated C++ file are due to regenerating the scanner using Apple Flex (v2.6.4) on macOS, which introduces type casts (e.g., yy_size_t) and table layout changes not present in the original Linux-generated file.

Once the logic in seclang-scanner.ll and the new tests are approved, I recommend that a maintainer regenerates the parser files on the standard Linux CI environment to ensure a clean diff and adherence to the project's preferred formatting.

theseion: Thanks

theseion (approved):

airween:
@Jitterx69,

thanks for this PR again.

I saw you added syntax handling, which is definitely a very good step:

* [ruleRemoveById](https://github.com/owasp-modsecurity/ModSecurity/pull/3566/changes#diff-0783e2185501ce790d4d664c5d312df25937a3bb4143d7eed48a0e8a0833c272R30)
* [ruleRemoveByTag](https://github.com/owasp-modsecurity/ModSecurity/pull/3566/changes#diff-8288f78ffc63c6594e94b6f61901c93744d52592c93a6f8d62cd627623f8aa86R29)
* [ruleRemoveTargetById](https://github.com/owasp-modsecurity/ModSecurity/pull/3566/changes#diff-c64865f680f57d87b4925d4bb8c6664d60b7c3b52e08b4cfe3c9e54699b05151R33)
* [ruleRemoveTargetByTag](https://github.com/owasp-modsecurity/ModSecurity/pull/3566/changes#diff-5ed1614926d92b41b2f40c005c543b8ea33c6d42fdbc50912826e8c003a190ddR33)

Would you mind to add more tests to check these conditions too?

Please note that our regression test framework handles the parser error too, eg. you can add a wrong syntax in a test and expect a parser error. For eg. see [this](https://github.com/owasp-modsecurity/ModSecurity/blob/9b49187452b973009a1e51d62ee0a5b91c07dd94/test/test-cases/regression/action-block.json#L37-L45) block: here the `SecAction` does not have a disruptive action.

In the planned "negative" tests you could construct invalid exclusions that you handle in those conditions.

Unfortunately my [update merge](https://github.com/owasp-modsecurity/ModSecurity/pull/3566/changes/14cf56fa96f2e9555afc625f8d6acf3894beb503) [failed](https://github.com/owasp-modsecurity/ModSecurity/actions/runs/27639421890/job/81735787856?pr=3566#step:11:81) you PR, could you rebase it?

But may be you might wait to merge of [this](https://github.com/owasp-modsecurity/ModSecurity/pull/3577) PR.

Let us know if you need any help.

airween:
Also please take a look at the Copilot suggestions.

airween:
@Jitterx69 - could you take a look at this?

airween:
Thanks again, @Jitterx69!

PR #3567, opened 2026-05-15
Fix GCC -Wall warnings on Fedora/RHEL builds

mikelolasagasti:
## what

- Fix compiler warnings reported when building with GCC and Fedora/RHEL-style flags (`-Wall`, hardening, ...).
- Each commit targets one warning class / location where practical.
- Build log warnings reduced from ~1270 to ~57 (same Fedora 45 / 3.0.15 build); remainder left for follow-up PRs.

## why

- Clean build logs are just better.
- Fewer noisy warnings makes real regressions easier to spot in CI/packaging builds.

## references

- Fedora 45 build log without the patches: https://kojipkgs.fedoraproject.org//packages/libmodsecurity/3.0.15/1.fc45/data/logs/x86_64/build.log
- Fedora 45 build log with these patches (temp logs): https://kojipkgs.fedoraproject.org//work/tasks/231/145490231/build.log

### DIFF
--- a/headers/modsecurity/anchored_set_variable_translation_proxy.h
+++ b/headers/modsecurity/anchored_set_variable_translation_proxy.h
@@ -43,7 +43,8 @@ class AnchoredSetVariableTranslationProxy {
         m_fount(fount)
     {
         m_translate = [](const std::string *name, std::vector<const VariableValue *> *l) {
-            for (int i = 0; i < l->size(); ++i) {
+            for (std::vector<const VariableValue *>::size_type i = 0;
+                i < l->size(); ++i) {
                 VariableValue *newVariableValue = new VariableValue(name, &l->at(i)->getKey(), &l->at(i)->getKey());
                 const VariableValue *oldVariableValue = l->at(i);
                 l->at(i) = newVariableValue;
--- a/headers/modsecurity/intervention.h
+++ b/headers/modsecurity/intervention.h
@@ -30,33 +30,33 @@ typedef struct ModSecurityIntervention_t {
 
 #ifdef __cplusplus
 namespace intervention {
-    static void reset(ModSecurityIntervention_t *i) {
+    inline void reset(ModSecurityIntervention_t *i) {
         i->status = 200;
         i->pause = 0;
         i->disruptive = 0;
     }
 
-    static void clean(ModSecurityIntervention_t *i) {
+    inline void clean(ModSecurityIntervention_t *i) {
         i->url = NULL;
         i->log = NULL;
         reset(i);
     }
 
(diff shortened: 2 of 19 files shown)


mikelolasagasti:
Fixed 3 SonarCloud warnings.

The "10.9% Duplication on New Code (required ≤ 3%)" Quality Gate is out of the scope of this PR.

airween:
Hi @mikelolasagasti,

could you pick [these](https://github.com/owasp-modsecurity/ModSecurity/pull/3577) changes or update your branch?

Thanks!

mikelolasagasti:
@airween rebased against current master. Also added fixes for CoPilot findings (commits 21 & 22) and fixed the githab AI findings with with previous approach.

airween:
@mikelolasagasti thanks - meanwhile there were some changes recently (see 3.0.16), could you pick up those modifications?

For eg. [utf8_to_unicode.cc](https://github.com/mikelolasagasti/ModSecurity/blob/a8136523520505a57e8af0dab648374e6d434218/src/actions/transformations/utf8_to_unicode.cc) is completely new.

mikelolasagasti:
@airween the PR is rebased against current `v3/master`, so it includes all the recent changes.

For `utf8_to_unicode.cc` I had to make a few changes in my patch on top of your recent changes: https://github.com/owasp-modsecurity/ModSecurity/pull/3567/changes/311e04aa809870a1e5b7719ea85ad4d0cb6f0c20

airween:
> For `utf8_to_unicode.cc` I had to make a few changes in my patch on top of your recent changes: [311e04a](https://github.com/owasp-modsecurity/ModSecurity/commit/311e04aa809870a1e5b7719ea85ad4d0cb6f0c20)

Yes - I think you should use the mentioned fix.

Also please upgrade your git submodules - I've added a test case against that bug to [secrules-language-tests](https://github.com/owasp-modsecurity/secrules-language-tests/commit/ae4258dbb6285bddfe089517354d69295b878e6f#diff-6cb26f81f0a2347790334f595b771f1a31bb80c6cff2ad7cef495c68bdfa16b4) repository too (which is a submodule - it was suspicious that i386 tests were success in the CI workflow).

airween:
> Yes - I think you should use the mentioned fix.

sorry, now I reviewed your changes and it seems it built on my previous changes, but yours are more strict - which is better. So I would accept your changes in case of file `utf8_to_unicode.cc`, but it would be good to eliminate the Sonar's "Duplicated Lines (%) on New Code 31.6%" report.

mikelolasagasti:
Added an extra commit to address the duplicated code issue that was already present in the code. With this SonarQube shows all green.

mikelolasagasti:
@airween should be all green now

PR #3570, opened 2026-06-08
ci: pin Windows version in workflow

airween:
## what

This PR pins the Windows version in CI workflow.

## why

It seems like GH upgraded the machines and we need to align the workflow for the new environments. Until then, I just pinned the last stable version.


airween:
A failed test:

https://github.com/airween/ModSecurity/actions/runs/27101560510/job/79983113005

fzipi (approved):

Issue #3394, opened 2025-05-29
ip collection does not seem to work properly with latest ModSecurity


ne20002:
I am using `@rbl xbl.spamhaus.org.` rule for protecting my Fediverse server and I face the issue that with latest version of either ModSecurity or the docker.io/owasp/modsecurity-crs:nginx container the collection handling for the IP collection seems to not longer work properly.

In my pi-hole I see a huge number of queries to .xbl.spamhaus.org for all the ip addresses checked and with queries to the same ip based query many times where I would expect that this is not the case due to how I set up the ModSecurity rules.

- The issue is happening with the latest version of the docker.io/owasp/modsecurity-crs:nginx container.
- It is far less a occuring with docker.io/owasp/modsecurity-crs:4.10-nginx-202501050801 even when mounting ruleset 4.14 to this container.
- I don't believe it is a problem with the CRS.

Based on what I see I would see it as if
- the expirevar for the IP collection is not working properly
- maybe the persistent store is causing the problem (as I don't see any problems with the TX collection which is also a collection but not persistet). This might be an docker image problem. 

According to modsecurity.conf the SecTmpDir is set to /tmp/modsecurity/tmp (as data and upload directory are also located in /tmp/modsecurity).
This directory is empty. Wouldn't this be the directory for persistent storage and shouldn't there be a file for the IP collection database?

This is the rules I use:

```
# xbl.spamhaus.org to block malicious/infected ips
SecAction \
    "phase:1,id:1100,\
    t:none,pass,nolog,\
    tag:'COLLECTIONS',\
    initcol:ip=%{remote_addr},\
    setvar:'tx.real_ip=%{remote_addr}'"

SecRule IP:SPAMMER "@eq 1" \
    "phase:1,id:1101,\
    t:none,deny,log,auditlog,\
    msg:'Request from Known SPAM Source (Previous RBL Match)',\
    tag:'AUTOMATION/MALICIOUS',\
    severity:'CRITICAL',\
    setvar:'tx.msg=%{rule.msg}'"

SecRule IP:PREVIOUS_RBL_CHECK "@eq 1" "phase:1,id:1102,t:none,pass,nolog,skipAfter:END_RBL_LOOKUP"

SecAction "phase:1,id:1103,\
    t:none,pass,nolog,\
    setvar:ip.previous_rbl_check=1,\
    expirevar:ip.previous_rbl_check=3600"

[...]

PR #3597, opened 2026-07-10
fix(collections): Honour SecDataDir for LMDB, fix literal setvar writ…

20vikash:
## what

- SecDataDir is now honoured by the LMDB backend. The path is stored in
  RulesSetProperties and used by the MDBEnvProvider singleton to open the LMDB
  environment (data.mdb/lock.mdb) inside that directory. Falls back to the
  legacy "./modsec-shared-collections" file when SecDataDir is unset.

- SecDataDir misconfiguration is now loud. The parser validates the path
  (exists, is a directory, writable) and fails configuration load on error,
  surfacing the message through getParserError to the connector error log.

- LMDB open failures are now logged to stderr instead of being silently
  swallowed.

- Literal setvar targets now persist. set_var.cc only dispatched writes to
  the dynamic variable subclasses. Variables parsed as literal dict elements
  fell through an empty else branch and were never stored. This adds name
  based dispatch for the literal case.

- Collection variable names are now case-insensitive. The composite key
  compared variable names case-sensitively, so setvar:ip.counter and
  IP:COUNTER resolved to different entries. All Collection wrapper methods now
  lowercase the variable name term.

- Bison and flex outputs regenerated from the edited grammar.

## why

- SecDataDir is a documented directive shipped in modsecurity.conf-recommended.
  Users expect it to control where persistent collection data lives. In v3
  with LMDB it was a no-op: the value was discarded and the file always
  appeared in the worker current working directory, with no error in the log.

- The silent setvar no-op was a data loss bug. Any rule writing a literal
  collection variable lost the value with no error. The 2018 RunTimeString
  refactor replaced the original generic dispatcher with per type dynamic
  casts that omitted the literal case by oversight, not by design.

- Case insensitive variable names are a SecLang property: the scanner matches
  collection prefixes case insensitively. The case sensitive key broke the
  natural pattern of a lowercase write and uppercase read, silently producing
  empty match values. Fixing it at the Collection wrapper boundary covers both
  LMDB and in memory backends.

## references

- Hardcoded LMDB path: lmdb.cc mdb_env_open of "./modsec-shared-collections".
- setvar literal no-op: set_var.cc dynamic cast dispatch, introduced in
  commit f17af957 "Using RunTimeString on setvar action".
- Case sensitive keys: collection.h composite key wrappers, lmdb.cc strncmp.
- modsecurity.conf-recommended line 203 ships "SecDataDir /tmp/".
- Tests that informed the fix: action-expirevar.json (lowercase write plus
  uppercase read asserted the value), collection-regular_expression_selection
  .json (uppercase setvar only asserted the debug log line, masking the no-op).

## related issue
[#3394](https://github.com/owasp-modsecurity/ModSecurity/issues/3394)

## changed files

Hand edited:
- headers/modsecurity/rules_set_properties.h
- headers/modsecurity/collection/collection.h
- src/parser/seclang-parser.yy
- src/collection/backend/lmdb.h
- src/collection/backend/lmdb.cc
- src/actions/set_var.cc

Regenerated:
- src/parser/seclang-parser.cc
- src/parser/seclang-parser.hh
- src/parser/seclang-scanner.cc

### DIFF
--- a/headers/modsecurity/collection/collection.h
+++ b/headers/modsecurity/collection/collection.h
@@ -22,6 +22,7 @@
 #include <algorithm>
 #include <memory>
 #include <cstdint>
+#include <cctype>
 #endif
 
 
@@ -46,6 +47,23 @@ class Collection {
  public:
     explicit Collection(const std::string &a) : m_name(a) { }
     virtual ~Collection() { }
+
+    /*
+     * Collection variable names are case-insensitive in SecLang (e.g.
+     * `setvar:ip.counter` and `IP:COUNTER` refer to the same key). Normalise
+     * the variable-name term of the composite LMDB key to lowercase so that
+     * writes and reads performed with different casing hit the same entry.
+     * The compartment values (collection key, web app id) are intentionally
+     * left untouched since they may be opaque, case-sensitive identifiers.
+     */
+    static std::string normKey(const std::string &k) {
+        std::string out;
+        out.reserve(k.size());
+        for (char c : k) {
+            out.push_back(static_cast<char>(std::tolower(static_cast<unsigned char>(c))));
+        }
+        return out;
+    }
 
     virtual bool storeOrUpdateFirst(const std::string &key,
         const std::string &value) = 0;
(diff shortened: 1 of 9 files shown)


airween:
Hi @20vikash,

thanks for this PR.

To be honest, I've been working on this solution for a while (well, I'm almost done), but I want to add that to 4.0, not to 3.0.

I hope I can finish that work soon and can send my PR - I'll let you know when I'm done.

20vikash:
@airween Hey. Thanks for the reply. Hope to see the full fix in V4. I encountered this silent failure while implementing a basic rate limiter. So, decided to make a quick fix in V3.

Issue #3295, opened 2024-11-07
Operator @rx has different flags in two engines


airween:
**Describe the bug**

It seems like the `@rx` operator has a different behavior in two engines (mod_security2 and libmodsecurity3)

mod_security2 has these PCRE flags:

[PCRE2](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/re_operators.c#L990), [PCRE](https://github.com/owasp-modsecurity/ModSecurity/blob/v2/master/apache2/re_operators.c#L992)

libmodsecurity3 has these ones:

[PCRE2](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/utils/regex.cc#L100), [PCRE](https://github.com/owasp-modsecurity/ModSecurity/blob/v3/master/src/utils/regex.cc#L106)

**To Reproduce**

https://github.com/coreruleset/coreruleset/issues/3277

We should discuss:
* do we want to resolve this issue?
* how?

airween:
I think [this](https://github.com/corazawaf/coraza/pull/876#issuecomment-2462004088) is a good idea, I mean add a build flag to libmodsecurity3 (mod_security2 is not affected) which changes the flag, and not in the next release but after that we make it mandatory (and we can add an optional build flag to keep the old (current) one).

marcstern:
PCRE2_DOTALL & PCRE2_DOLLAR_ENDONLY look the right way for me as we need to check multiline ARGS

fzipi:
@airween I opened #3600 implementing this using the build-flag approach you proposed, mirroring what Coraza shipped in corazawaf/coraza#876 (`--enable-regex-dollar-endonly`, off by default). Only `@rx`/`@rxGlobal` are affected; other operators keep their current regex flags. Would appreciate a look when you have a chance.

airween:
@fzipi thanks both for put this again to the table and also for the PR (#3600).

Let me review the whole situation. In 2024's CRS retreat we discussed about this with @M4tteoP, but I have a memory that we agreed libmodsecurity3's behavior is the correct (but may be I'm wrong...). After a quick review now I think mod_security2's behavior should be the reference.

M4tteoP:
[...]

PR #3600, opened 2026-07-19
feat: add opt-in build flag to match v2's @rx anchor semantics (#3295)

fzipi:
## Summary

Closes the behavioral gap described in #3295: `@rx` in libmodsecurity (v3) compiles patterns with `PCRE(2)_DOTALL | PCRE(2)_MULTILINE`, while ModSecurity v2's `@rx` (`msre_op_rx_param_init`) uses `PCRE(2)_DOTALL | PCRE(2)_DOLLAR_ENDONLY`. This means `^`/`$` anchors behave differently against multi-line values (e.g. multi-line `ARGS`) between the two engines, which is what surfaced as a CRS false-positive/negative report in coreruleset/coreruleset#3277.

Both commenters on #3295 agreed the target behavior should be v2's flags; the open question was only the rollout mechanism. This PR follows airween's proposal to use a build flag, mirroring the approach Coraza already shipped for the same issue (corazawaf/coraza#876, merged, gated behind `coraza.rule.no_regex_multiline`, off by default).

- `configure.ac`: adds `--enable-regex-dollar-endonly` (off by default), defining `MODSEC_REGEX_DOLLAR_ENDONLY` when enabled.
- `src/utils/regex.h` / `regex.cc`: `Regex` gains a `multiLine` constructor parameter, defaulting to `true` (current behavior). All existing callers (`verifyCC`, `verifySSN`, `verifyCPF`, internal chain/audit-log/collection regexes, etc.) are unaffected since they don't pass the new parameter.
- `src/operators/rx.cc` / `rx_global.cc`: only `@rx` and `@rxGlobal` pass `multiLine=false` when the build flag is set, compiling with `DOTALL | DOLLAR_ENDONLY` instead of `DOTALL | MULTILINE`. This mirrors v2, where only `@rx` uses `DOLLAR_ENDONLY` while other operators (`verifyCC`, `verifySSN`, `gsbLookup`) keep `MULTILINE` unconditionally.
- `README.md`: documents the new flag next to the existing PCRE2/PCRE section.

## Test plan

- [x] Standalone harness exercising `Utils::Regex` directly against both the PCRE2 and legacy PCRE1 backends, confirming `multiLine=false` reproduces v2's dollar-endonly anchor semantics (no match across internal line breaks; trailing-newline `$` behavior matches PCRE's `DOLLAR_ENDONLY` docs) while `multiLine=true` preserves current behavior.
- [x] Full `./configure && make` succeeds with the flag both off (default) and on (`--enable-regex-dollar-endonly=yes`).
- [x] End-to-end: built a small program against the compiled library with `SecRule ARGS:param1 "^hello.*world$" "deny"` evaluated against a multi-line `ARGS` value. Default build blocks (reproducing the current v3 behavior described in #3295); with the flag enabled it does not block (matching v2).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


README.md
```diff
@@ -115,6 +115,11 @@ Windows build information can be found [here](build/win32/README.md).
   * **Default:** PCRE2 is detected and used.
   * **Fallback:** legacy PCRE can be used if `--with-pcre` is explicitly provided (`WITH_PCRE`).
 * In other words, current builds expect PCRE2 unless explicitly configured otherwise.
+* `@rx` and `@rxGlobal` compile patterns with `DOTALL | MULTILINE` by default. Pass
+  `--enable-regex-dollar-endonly=yes` to compile them with `DOTALL | DOLLAR_ENDONLY`
+  instead, matching ModSecurity v2's `@rx` behavior (`^`/`$` anchor only at the
+  start/end of the whole subject, not at internal line breaks). This flag is
+  disabled by default; see [issue #3295](https://github.com/owasp-modsecurity/ModSecurity/issues/3295).
```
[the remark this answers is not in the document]
fzipi: Fixed in d595832f — reworded to separate the two effects precisely: `MULTILINE` controls where `^`/`$` can anchor (per-line vs. whole-subject), while `DOLLAR_ENDONLY` only further restricts `$` against a trailing newline. `DOTALL` is unchanged either way, so `.` still matches across internal line breaks in both configurations. Thanks for catching the imprecise wording.

src/utils/regex.cc
```diff
@@ -87,11 +87,12 @@ bool crlfIsNewline() {
...
-Regex::Regex(const std::string& pattern_, bool ignoreCase)
+Regex::Regex(const std::string& pattern_, bool ignoreCase, bool multiLine)
     : pattern(pattern_.empty() ? ".*" : pattern_) {
 #ifndef WITH_PCRE
     PCRE2_SPTR pcre2_pattern = reinterpret_cast<PCRE2_SPTR>(pattern.c_str());
-    uint32_t pcre2_options = (PCRE2_DOTALL|PCRE2_MULTILINE);
+    uint32_t pcre2_options = PCRE2_DOTALL |
+        (multiLine ? PCRE2_MULTILINE : PCRE2_DOLLAR_ENDONLY);
     if (ignoreCase) {
         pcre2_options |= PCRE2_CASELESS;
```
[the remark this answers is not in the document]
fzipi: Added in d595832f — `test/unit/regex_multiline_test.cc` exercises `Utils::Regex(..., multiLine=false)` directly (cross-line anchoring, single-line sanity, and the $-before-trailing-newline case), alongside the multiLine=true default. The existing test dispatcher (`test-suite.sh`) only knew how to run JSON test-case files against `unit_tests`/`regression_tests`, so I wired this in as its own small binary via `test/Makefile.am` + `test-suite.in` and added a dispatch branch in `test-suite.sh` for non-JSON test entries. Verified it builds and passes, and that the existing `@rx` JSON regression suite is unaffected.

test/unit/regex_multiline_test.cc
```diff
@@ -0,0 +1,82 @@
+/*
```
airween: This is a new file, probably to check the new feature.

libmodsecurity3's regression test has a great feature: a test case can contain a key `resource`, see how does it handle it [here](https://github.com/fzipi/ModSecurity/blob/d595832f91bd153a4104ddc45d3a91260766d415/test/regression/regression.cc#L412-L432) and an example in a real test is [here](https://github.com/fzipi/ModSecurity/blob/d595832f91bd153a4104ddc45d3a91260766d415/test/test-cases/regression/action-ctl_request_body_processor.json#L6).

I suggest you to check this feature and consider to use that, instead of introduce a new test file.

test/test-suite.sh
airween, line 23: Seems like you didn't change this line, right? If not, I'm going to flag this as an accepted issue, you don't need to change this.

test/unit/regex_multiline_test.cc
fzipi: Good call, thanks — much better fit for the project's conventions. Replaced the standalone binary in 08d0afc0 with a JSON regression test using the `resource` gating mechanism: registered `regex-dollar-endonly` in `regression.cc` under `MODSEC_REGEX_DOLLAR_ENDONLY`, and added `test/test-cases/regression/operator-rx-dollar-endonly.json` covering both a single-line match (still blocks) and a multi-line ARGS value (anchors no longer span line breaks). Verified: skips cleanly on a default build, passes on a build with `--enable-regex-dollar-endonly=yes`, and the existing `operator-rx.json` suite is unaffected.

test/test-suite.sh
fzipi: Correct, that line's content was unchanged — it only shifted because of the branch added above it. Moot now anyway: after switching to the resource-gated JSON regression test approach (08d0afc0), test-suite.sh no longer needs any changes at all and is back to being byte-identical to upstream.

theMiddleBlue:
Hey 👋 nice work on this one! (I was working on a similar issue and I found this PR)

From a bit of poking around, this doesn't seem to fully close things on the security side (though I'm not sure whether you're already aware of it). Didn't want to drop specifics in a public thread, so I'm happy to share more if that's useful. Thanks 🙏

airween:
> From a bit of poking around, this doesn't seem to fully close things on the security side (though I'm not sure whether you're already aware of it). Didn't want to drop specifics in a public thread, so I'm happy to share more if that's useful. Thanks 🙏

Feel free to open a security advisory, there we can discuss the details.

Thank you.

fzipi:
Hey @theMiddleBlue, nice to see you around! 👋

fzipi:
JFYI: this PR is scoped for closing the gap between v2 and v3.

PR #3607, opened 2026-08-01
test/fuzzer: fix AFL harness input handling and restore it to a compiling state

shotintoeternity:
## what

Restores `test/fuzzer/afl_fuzzer.cc` to a compiling state and makes it fuzz the
actual input it is given.

- Builds the input string from `buf` instead of the fill constructor, and skips
  iterations where `read` returns nothing.
- Updates `op_test` to the current 3-argument `Operator::evaluate`.
- Updates the 37 generated transformation calls to the current
  `Transformation::transform(std::string&, const Transaction*)`.
- Adds a single-shot `__AFL_LOOP` fallback so the harness compiles without afl++
  installed.

One file, no build-system or CI changes.

## why

**The harness does not compile.** On current `v3/master`:

```
$ c++ -fsyntax-only -std=c++17 -D'__AFL_LOOP(x)=1' \
      -I. -Iheaders -Isrc -Iothers test/fuzzer/afl_fuzzer.cc
38 errors, all "no matching member function for call to 'evaluate'"
```

37 are the generated transformation lines and 1 is `op_test`. `__AFL_LOOP` is
defined on the command line above because `afl-clang-fast` supplies it, so it is
not one of the errors.

Two API changes caused this, neither of which touched the harness:

- **`5d398907`** (merged 2024-08-27) renamed `Transformation::evaluate` to
  `transform(std::string &value, const Transaction *trans)` and removed
  `Action::evaluate(const std::string&, Transaction*)`.
- **`4df297b5`** (merged 2024-10-07) changed the fourth parameter of
  `Operator::evaluate` from `std::shared_ptr<RuleMessage>` to `RuleMessage&`, so
  `nullptr` no longer binds.

Syntax-checking `5d398907^` gives 0 errors, so the harness last built in August
2024. It is gated behind `--enable-afl-fuzz`, which defaults to false
(`configure.ac:284`) and only then generates `test/fuzzer/Makefile`
(`configure.ac:386`), so no CI job surfaces the breakage.

**The input was never the fuzzer's input.** Independently of the above,
`std::string(read_bytes, 128)` selects the fill constructor
`basic_string(size_type count, CharT ch)`, giving `read_bytes` copies of the
character `128`. Overload resolution is not ambiguous, which is likely why it
survived: the pointer-and-length constructor is not viable because `ssize_t` does
not convert to `const char*`, and the iterator-pair constructor is not viable
because deduction conflicts. `buf` appears three times in the file, in its
declaration, its `memset` and its `read`, and is never read from. The compiler has
been saying so all along:

```
warning: implicit conversion from 'int' to 'char' changes value
         from 128 to -128 [-Wconstant-conversion]
```

`git log -L145,145:test/fuzzer/afl_fuzzer.cc` returns a single commit, `c2d9a153`
("Adds support to afl fuzzer in the build system", 2015-12-22), so the line is
unchanged since the harness was written.

The practical effect is that the harness's input domain was a run of `0x80` bytes
whose only attacker-controlled property was its length, 0 to 128, which is 129
distinct inputs. `t:hexDecode` never saw a hex digit, `Base64Decode` never saw
base64, `DetectSQLi` and `DetectXSS` never saw a quote or an angle bracket.

## on the transformation calls

The 37 transformation lines are ported to a helper that mirrors the existing
`op_test`:

```cpp
template <typename T>
inline void tfn_test(const std::string &tfnName, const std::string &s,
    const Transaction *t) {
    T tfn(tfnName);
    // transform() rewrites in place, so each transformation gets its own copy
    // of the fuzzer input rather than the output of the previous one.
    std::string value(s);
    tfn.transform(value, t);
}
```

so each generated line becomes `tfn_test<Base64Decode>("Base64Decode", s, t);`.

This constructs the classes directly, as the current file already does, rather than
routing through `Transformation::instantiate`. That is deliberate.
`instantiate` matches with

```c
#define IF_MATCH(b) \
    if (a.compare(2, std::strlen(#b), #b) == 0)
```

The compare starts at offset 2 because it expects the rule-language spelling with
the `t:` prefix, as in `t:base64Decode`. Class-cased names match nothing, fall
through every branch, and reach the final `return new Transformation(a)`, which is
the base class, whose `transform()` returns `false` without touching the value.
Passing the 37 class names to `instantiate` and checking the dynamic type:

```
Base64Decode         dynamic_type=Transformation  transform_changed=NO
Base64DecodeExt      dynamic_type=Transformation  transform_changed=NO
HexDecode            dynamic_type=Transformation  transform_changed=NO
...
base-class no-ops: 37   real transformations: 0

control, rule-language spelling with the t: prefix:
t:hexDecode          dynamic_type=<subclass>      transform_changed=yes
t:lowercase          dynamic_type=<subclass>      transform_changed=yes
t:urlDecode          dynamic_type=<subclass>      transform_changed=yes
```

Direct construction sidesteps the prefix convention entirely and keeps the name
list compile-checked, so a wrong name is a build error rather than a silent no-op.

## on the __AFL_LOOP fallback

`afl-clang-fast` defines `__AFL_LOOP`. Without it the file does not compile, which
is part of why this rot went unnoticed for two years. The fallback here is
single-shot rather than a literal `1`, so a non-AFL build reads stdin once and
exits instead of spinning at EOF against the new `read_bytes <= 0` guard.

## verification

Built `libmodsecurity` from this branch and compiled the harness against it with a
plain compiler:

- 0 errors. The only warnings are the 4 pre-existing `-Wcomment` ones from the `/*`
  sequences inside the "generated by" block comments, which are untouched here.
- Runs on real stdin input and exits 0.
- Exits 0 on empty input rather than looping.

## credit

The input-handling bug and the `op_test` signature fix are @Easton97-Jens's find,
from #3556. This PR carries the `afl_fuzzer.cc` portion of that work on its own, at
their suggestion, so it can be reviewed without the CI and workflow changes bundled
there. The CI additions in #3556 are unaffected by this and remain for separate
discussion.

## Summary by CodeRabbit

* **Tests**
  * Improved fuzz-testing reliability across environments where AFL loop support is unavailable.
  * Added safe handling for empty, invalid, or nonpositive input reads.
  * Ensured each transformation is tested with independent input data.
  * Updated test execution to use the current evaluation interface.
  * Improved generated test setup for more consistent fuzzing across supported environments.


shotintoeternity:
Hi, can a reviewer on the team please take a look at this proposed fix? Thanks!

PR #3613, opened 2026-08-14
fix: record real response status for IIS audit log F part

### LINKED ISSUE #3612: IIS connector: audit log F part always shows '500 Internal Server Error' for normal traffic
## Summary

In the IIS connector (`iis/mymodule.cpp`), the audit log's F part (response status line) renders as `HTTP/1.1 500 Internal Server Error` for **every** transaction, even for normal requests that actually return e.g. `200 OK` or `404`.

## Root cause

The IIS connector never copies the real HTTP response status into the `request_rec`. In `CMyHttpModule::OnSendResponse`, `r->status` is left at its initial value `0` (the request_rec is `apr_pcalloc`'d). When `hookfn_log_transaction` later builds the F part it calls `ap_get_status_line(r->status)`; for `r->status == 0` the standalone `ap_index_of_response()` maps anything `< 100` to the `LEVEL_500` bucket, producing `500 Internal Server Error`.

## Affected output

```
--A--
[13/Aug/2026:16:16:40.333548] ...
--F--
HTTP/1.1 500 Internal Server Error   <- should be the real status (e.g. 200 OK)
```

## Proposed fix

In `CMyHttpModule::OnSendResponse` (`iis/mymodule.cpp`), transfer `pRawHttpResponse->StatusCode` into `r->status` (and build `r->status_line` from the reason phrase) before the rest of the response handling:

```cpp
if(pRawHttpResponse->StatusCode > 0)
{
    r->status = pRawHttpResponse->StatusCode;

    if(pRawHttpResponse->pReason != NULL && pRawHttpResponse->ReasonLength > 0)
    {
        r->status_line = apr_psprintf(r->pool, "%d %s", r->status,
            ZeroTerminate(pRawHttpResponse->pReason, pRawHttpResponse->ReasonLength, r->pool));
    }
}
```

This makes the audit log F part and relevant-status checks use the real response code. Verified against a local IIS (default site + OWASP CRS): a normal `GET /` now logs `HTTP/1.1 200 OK`, blocked requests log `HTTP/1.1 403 ModSecurity Action`.

A13501350:
## Summary

Fixes the IIS connector so the audit log's F part records the real HTTP response status instead of a bogus `500 Internal Server Error`.

## Problem

In the IIS connector, `r->status` was never populated from the actual HTTP response. `request_rec` is `apr_pcalloc`'d, so `r->status` stayed `0`. When the logging hook builds the F part it calls `ap_get_status_line(r->status)`; for `r->status == 0` the standalone `ap_index_of_response()` maps any value `< 100` to `LEVEL_500`, so **every** transaction logged `HTTP/1.1 500 Internal Server Error` regardless of the true response code.

```
--F--
HTTP/1.1 500 Internal Server Error   <- wrong, e.g. actual response is 200 OK
```

There was a second, related source of bogus `500` records: the IIS default-document rewrite (e.g. `GET /` internally redirected to `/iisstart.htm`) supersedes the original request before it ever reaches `OnSendResponse`, so its `r->status` stays `0` and `hook_log_transaction` wrote a phantom `500` record for it.

## Fix

1. In `CMyHttpModule::OnSendResponse` (`iis/mymodule.cpp`), transfer the raw response status into `r->status` (and build `r->status_line` from the reason phrase) before the rest of the response handling:

```cpp
if(pRawHttpResponse->StatusCode > 0)
{
    r->status = pRawHttpResponse->StatusCode;

    if(pRawHttpResponse->pReason != NULL && pRawHttpResponse->ReasonLength > 0)
    {
        r->status_line = apr_psprintf(r->pool, "%d %s", r->status,
            ZeroTerminate(pRawHttpResponse->pReason, pRawHttpResponse->ReasonLength, r->pool));
    }
}
```

2. In `hook_log_transaction` (`apache2/mod_security2.c`, under `VERSION_IIS`), skip audit logging for transactions that never received a response status **and** were not intercepted. In the IIS connector `r->status` is only ever set by `OnSendResponse`, so `r->status == 0 && !msr->was_intercepted` uniquely identifies requests superseded by an internal redirect:

```c
#if defined(VERSION_IIS)
    if (r->status == 0 && msr->was_intercepted == 0) {
        return DECLINED;
    }
#endif
```

A blocked request always reaches `OnSendResponse` with `r->status` set (e.g. `403`), so intercepted transactions are unaffected.

## Verification

Verified against a local IIS default site with OWASP CRS (and in the module's Windows CI, which enables the audit log and asserts the F-part status lines):

- Normal `GET /` now logs `HTTP/1.1 200 OK` (single record; the phantom `500` records for the internal redirect are gone)
- Blocked request (`SQLi`/`XSS`) logs `HTTP/1.1 403 ModSecurity Action`

Closes #3612

### DIFF
--- a/iis/mymodule.cpp
+++ b/iis/mymodule.cpp
@@ -450,6 +450,22 @@ CMyHttpModule::OnSendResponse(
 	pHttpResponse = pHttpContext->GetResponse();
 	pRawHttpResponse = pHttpResponse->GetRawHttpResponse();
 
+	// here we must transfer response status
+	// otherwise r->status stays 0 and the audit log records a
+	// bogus "500 Internal Server Error" (ap_get_status_line(0))
+	// for every transaction
+	//
+	if(pRawHttpResponse->StatusCode > 0)
+	{
+		r->status = pRawHttpResponse->StatusCode;
+
+		if(pRawHttpResponse->pReason != NULL && pRawHttpResponse->ReasonLength > 0)
+		{
+			r->status_line = apr_psprintf(r->pool, "%d %s", r->status,
+				ZeroTerminate(pRawHttpResponse->pReason, pRawHttpResponse->ReasonLength, r->pool));
+		}
+	}
+
 	// here we must add handling of chunked response
 	// apparently IIS 7 calls this handler once per chunk
 	// see: http://stackoverflow.com/questions/4385249/how-to-buffer-and-process-chunked-data-before-sending-headers-in-iis7-native-mod
--- a/iis/tests/web.config
+++ b/iis/tests/web.config
@@ -1,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <system.webServer>
+    <rewrite>
+      <rules>
+        <rule name="ReverseProxyToAlbedo" stopProcessing="true">
+          <match url="(.*)" />
+          <action type="Rewrite" url="http://localhost:8080/{R:1}" />
+        </rule>
+      </rules>
+    </rewrite>
+  </system.webServer>
+</configuration>
--- a/.github/workflows/test-ci-windows.yml
+++ b/.github/workflows/test-ci-windows.yml
@@ -229,8 +229,12 @@ jobs:
               $newName = $_.Name.Replace(".example", "")
               Rename-Item -Path $_.FullName -NewName $newName
           }
-          
+
+          $testConfigDir = "${{ github.workspace }}\iis\tests"
+
           $modSecurityConfigFile = "$modSecurityConfigDir\modsecurity_iis.conf"
+
+          Add-Content -Path "$crsDir\crs-setup.conf" -Value (Get-Content -Path "$testConfigDir\crs-900005.conf")
 
           $crsRules = @(
             "Include coreruleset/crs-setup.conf",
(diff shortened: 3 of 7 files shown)


iis/tests/web.config
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<configuration>
+  <system.webServer>
```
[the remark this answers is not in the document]
A13501350: `web.config` exists solely for the CI test environment and never serves real traffic, so I'd consider this ignorable.

iis/mymodule.cpp
A13501350, line 1: Like I said before, legacy code is usually not convenient to refactor.

iis/mymodule.cpp
```diff
@@ -450,6 +450,22 @@ CMyHttpModule::OnSendResponse(
...
+	// here we must transfer response status
+	// otherwise r->status stays 0 and the audit log records a
+	// bogus "500 Internal Server Error" (ap_get_status_line(0))
+	// for every transaction
+	//
+	if(pRawHttpResponse->StatusCode > 0)
+	{
+		r->status = pRawHttpResponse->StatusCode;
+
+		if(pRawHttpResponse->pReason != NULL && pRawHttpResponse->ReasonLength > 0)
```
airween: Seems like this is a new block - would you mind to consider to accept SonarCloud's suggestion here?

PR #3620, opened 2026-08-27
IIS: fix Content-Length formatting (correct printf specifier + full buffer size)

### LINKED ISSUE #3619: IIS: Content-Length header corrupted/truncated for large responses (wrong printf format + off-by-one buffer size)
## Summary
In `iis/mymodule.cpp`, `StringCchPrintfA` is used to build the `Content-Length` header value, but it has two defects that corrupt the header for large responses:

1. **Wrong format specifier for 64-bit value.** At the response path, `ulTotalLength` is a `ULONGLONG`, but it is printed with `"%d"` (which expects a 32-bit `int`). `printf` only reads the low 32 bits, so the resulting `Content-Length` is wrong whenever the response body exceeds ~2 GiB. The other two call sites format an `unsigned int` `length` with `"%d"` as well (should be `%u`).
2. **Off-by-one destination size.** `StringCchPrintfA`'s `cchDest` argument is passed as `sizeof(szLength)/sizeof(CHAR) - 1` (= 20). Since this argument must include the null terminator, the buffer can only hold 19 digits. A 64-bit value can be 20 decimal digits (e.g. `18446744073709551615`), so it is silently truncated / the call fails.

## Impact
When ModSecurity for IIS must synthesize the `Content-Length` header (the only-response, non-chunked case), a wrong value causes clients to hang or error because the body length does not match the advertised header.

## Affected locations (v2/master)
- `iis/mymodule.cpp:642` — `ulTotalLength` (ULONGLONG) -> `"%llu"`
- `iis/mymodule.cpp:1140` — `length` (unsigned int) -> `"%u"`
- `iis/mymodule.cpp:1228` — `length` (unsigned int) -> `"%u"`

All three should also pass the full buffer size `sizeof(szLength)/sizeof(CHAR)` (21) instead of `... - 1`.

## Suggested fix
```c
CHAR szLength[21]; // Max length for a 64-bit int is 20 digits + null
ZeroMemory(szLength, sizeof(szLength));

HRESULT hr = StringCchPrintfA(
    szLength,
    sizeof(szLength) / sizeof(CHAR), // includes null terminator
    "%llu",                          // ULONGLONG
    ulTotalLength);
```

A13501350:
## Summary
Fixes the `Content-Length` header value being corrupted/truncated in the IIS module when ModSecurity has to synthesize it (only-response, non-chunked case).

Two defects were fixed in `iis/mymodule.cpp`:
- `ulTotalLength` is a `ULONGLONG`, but was printed with `"%d"` (32-bit `int`). `printf` only consumed the low 32 bits, producing a wrong header for response bodies larger than ~2 GiB. Now uses `"%llu"`.
- The two other call sites format an `unsigned int` `length` with `"%d"`; changed to `"%u"`.
- `StringCchPrintfA`'s `cchDest` was passed as `sizeof(szLength)/sizeof(CHAR) - 1` (20), which can only hold 19 digits + null and would truncate a 20-digit 64-bit value. Now passes the full buffer size (21, which includes the null terminator).

## Fixes
Closes #3619

## Changed locations
- `iis/mymodule.cpp:644` — `ulTotalLength` (ULONGLONG) -> `"%llu"`
- `iis/mymodule.cpp:1142` — `length` (unsigned int) -> `"%u"`
- `iis/mymodule.cpp:1230` — `length` (unsigned int) -> `"%u"`

All three now pass `sizeof(szLength) / sizeof(CHAR)` as the destination size.

### DIFF
--- a/iis/mymodule.cpp
+++ b/iis/mymodule.cpp
@@ -639,9 +639,9 @@ CMyHttpModule::OnSendResponse(
 
          ZeroMemory(szLength, sizeof(szLength));
 
-         hr = StringCchPrintfA(
+          hr = StringCchPrintfA(
                     szLength, 
-                    sizeof(szLength) / sizeof(CHAR) - 1, "%d", 
+                    sizeof(szLength) / sizeof(CHAR), "%llu", 
                     ulTotalLength);
 
         if(FAILED(hr))
@@ -1139,7 +1139,7 @@ apr_status_t WriteBodyCallback(request_rec *r, char *buf, unsigned int length)
 
     HRESULT hr = StringCchPrintfA(
             szLength, 
-            sizeof(szLength) / sizeof(CHAR) - 1, "%d", 
+            sizeof(szLength) / sizeof(CHAR), "%u", 
             length);
 
     if(FAILED(hr))
@@ -1227,7 +1227,7 @@ apr_status_t WriteResponseCallback(request_rec *r, char *buf, unsigned int lengt
 
     HRESULT hr = StringCchPrintfA(
             szLength, 
-            sizeof(szLength) / sizeof(CHAR) - 1, "%d", 
+            sizeof(szLength) / sizeof(CHAR), "%u", 
             length);
 
     if(FAILED(hr))


airween:
Hi @A13501350,

thank you for this PR!

Could you check [SonarCloud](https://sonarcloud.io/project/issues?id=owasp-modsecurity_ModSecurity&open=AaBCXTyM2e7bGHf3GgaY&pullRequest=3620&s=IMPACT_RANK&sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&types=CODE_SMELL) issues, and consider to change the suggestions?

A13501350:
Hi @airween, thank you for the review. The warning is about legacy code, unrelated to this fix. I'd like to keep the change minimal; unless there's a large-scale refactoring in the future, fixing it now won't help.

airween:
> Hi @airween, thank you for the review. The warning is about legacy code, unrelated to this fix. I'd like to keep the change minimal; unless there's a large-scale refactoring in the future, fixing it now won't help.

Ah, sorry, now I see (I just saw the same commit hash on Sonar's page that you added on GH, but didn't realize you changed the other part in the line).

Sorry, I'll accept those issues.

PR #3639, opened 2026-09-20
fix: drop MDB_WRITEMAP to avoid LMDB freelist assertion crash

fzipi:
## Summary

The `macOS (with lmdb)` CI job fails deterministically on `test/test-cases/regression/action-expirevar.json` and `test/test-cases/regression/issue-1831.json` with:

```
mdb.c:3900: Assertion 'x <= dl[0].mid' failed in mdb_freelist_save()
```

This reproduces with a minimal standalone LMDB program with no ModSecurity code involved: a `get`/`del`/`put`/commit cycle on a key, interleaved with a read-only `get` and a delete-only commit, crashes `mdb_freelist_save()` when the environment is opened with `MDB_WRITEMAP`.

The failing assertion is in the `mt_loose_pgs`/dirty-list reconciliation loop inside `mdb_freelist_save()`, which only executes `if (txn->mt_flags & MDB_TXN_WRITEMAP)`. This guard is still present at the current tip of LMDB's own `mdb.RE/1.0` branch, so it is not fixed by upgrading LMDB.

Apple Silicon macOS uses a 16KB page size (`sysctl hw.pagesize`), versus 4KB on the Linux CI runners. The Linux `with lmdb` jobs use `liblmdb-dev` 0.9.24 from Ubuntu and pass; the macOS job installs whatever Homebrew's `lmdb` formula currently resolves to (1.0.2 at the time of this PR). The same C++ code path is exercised on both platforms; only the macOS job fails, consistently, across multiple runs and multiple unrelated commits on `v3/master`.

## Change

`src/collection/backend/lmdb.cc`: `mdb_env_open()` no longer passes `MDB_WRITEMAP`, only `MDB_NOSUBDIR`.

## Test plan

- [x] Reproduced the crash with a standalone LMDB C program (no ModSecurity involved), confirmed it disappears when `MDB_WRITEMAP` is removed from `mdb_env_open`, using the same environment (LMDB 1.0.2 via Homebrew, arm64 macOS).
- [x] Built `v3/master` plus this one-line change locally with `./configure --with-lmdb --enable-assertions=yes` and `make`.
- [x] `make check`: 5043 total, 5025 pass, 18 skip, 0 fail (previously 3 failures, all in the two files above, both LMDB-collection related).
- [x] Ran `action-expirevar.json` and `issue-1831.json` individually, all sub-tests pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Updated database access behavior to open LMDB files without write-map semantics, improving compatibility with environments where memory-mapped writes are unavailable.


airween (approved):
LGTM, thank you!

airween:
Thank you, @fzipi!

